PR #12, opened 2016-01-12
[F] Add margin controls to reader

naomiyaki:
[FEATURE DELIVERS #111080632] New actions have been added to increment/
decrement margin values in the store, that use a now-abstracted version
of the font-size increment/decrement reducers. In the view, there are
now buttons to increase/decrease the margin size in the reader. It's
important to note that while these buttons do not have disabled states
for minimum and maximum sizes, they do stop changing the store when a
limit is reached.

### DIFF
--- a/client/src/actions/reader/ui/typography.js
+++ b/client/src/actions/reader/ui/typography.js
@@ -3,3 +3,5 @@ import { createAction } from 'redux-actions';
 export const selectFont = createAction('SELECT_FONT', subject => subject);
 export const incrementFontSize = createAction('INCREMENT_FONT_SIZE');
 export const decrementFontSize = createAction('DECREMENT_FONT_SIZE');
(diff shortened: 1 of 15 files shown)


zdavis:
Looks like we need to adjust our Hounds SCSS rules.

Zach

naomiyaki:
A bit! Some of these are OK, and I'd be happy to talk about setting a nesting limit if we want one, but the alphabetical order of CSS properties should probably go.

zdavis:
You may as well add the linting config to this pull request, and also fix errors. That way we can see it pass all CI checks before merging it. Sound OK?

zdavis:
Naomi, can you rebase this onto development and force push so we can see how it does with the new scss linting rules I just merged? Thanks!

zdavis:
Closing in favor of #14

PR #13, opened 2016-01-13
[C] Configured scss linting for Hound CI

naomiyaki:
A configuration yml has been added to trigger a custom configuration of
scss-lint. This initial configuration is based on a combination of
established standards and also Cast Iron Coding house style. Notably
the style for variables, mixins, and functions all follow standard JS
convention (camelcase). We are also trying out recess CSS order.
This initial config is not final, and may be modified after a few
rounds of commits.

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,2 +1,5 @@
 ruby:
   config_file: api/.rubocop.yml
+
+scss:
+  config_file: .scss-lint.yml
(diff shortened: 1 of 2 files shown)


zdavis:
Will merge after my 3:00 PM call

zdavis:
Will merge now, that is

PR #14, opened 2016-01-14
[F] Add margin to reader

zdavis:


zdavis:
Ok, @naomiyaki—couple of things.

First, I'd prefer not to merge this until it passes CI checks. I rebased this on top of your linter changes, thinking that your ruleset would reduce the number of errors, but we're still looking at 70 errors. Could you go ahead and fix these issues in a follow-up commit to this branch?

Also, please do stop using the castiron/manifold repo—you can push your feature branch directly to this repo, which makes it easier for me to commit to it. Notice that I've moved your branch to this repo, and this pull request is against that branch, not the one on the castiron account.

Zach

PR #15, opened 2016-01-14
[F] Improve HTML and CSS validation

zdavis:


zdavis:
Shitsnaps. On my machine, rubocop doesn't show any errors—apparently because it's not linting my specs—and all tests pass (except for one). I'll look into what's going on here and update this branch to address it.

PR #20, opened 2016-02-04
[F] Add scroll aware fade to heading

naomiyaki:
[FEATURE DELIVERS #112837725] A component has been added that can wrap
any component/group of components and add a "not-top" class based on a
scroll threshold that can be set as a component prop (or otherwise has
a default); This is used to conditionally show/hide a fading bottom
border on both the reader and front-end header banner.


zdavis:
Rebased to clean up commit history

PR #22, opened 2016-02-05
Add pagination to text-sections

naomiyaki:
Add pagination to text-sections

[FEATURE DELIVERS #112835767] Feature adds functional "Previous" and
"Next" buttons to text-sections. The buttons link to paths generated
based on the current section and the existence of previous and next
sections in the table of contents.

### DIFF
--- a/client/src/components/reader/SectionPagination.js
+++ b/client/src/components/reader/SectionPagination.js
@@ -2,17 +2,30 @@ import React, { Component, PropTypes } from 'react';
 
 export default class SectionPagination extends Component {
 
-  static propTypes = {};
+  static propTypes = {
+    textId: PropTypes.string,
+    sectionId: PropTypes.string
+  };
+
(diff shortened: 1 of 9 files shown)


client/src/components/reader/SectionPagination.js
```diff
@@ -0,0 +1,70 @@
...
+    return path;
+  }
+
+  getPreviousLink() {
+    let previousLink = '';
+    const previousNode = this.getSiblingSection(Number(this.props.sectionId), - 1);
+    if (previousNode) {
+      const previousPath = this.getSectionPath(previousNode);
+      previousLink = (
+        <a href={previousPath} className="pagination-previous">
```
zdavis: I merged this without catching this—what you're doing here will cause a full page reload. For links, we need to use the <Link> component provided by React Router, as we do in the reader TOC.

I'll fix in a subsequent commit.

PR #28, opened 2016-02-16
[F] Show section title when user scrolls down

naomiyaki:
Initial Commit:
[FEATURE DELIVERS #112837675] In the reader, the header banner now
shows the title of the text-section after the user begins to scroll
down into the reader. The design logic behind this is that the user
almost always sees the section title when the reader is at the top of a
given section, so showing the text title gives the most context. Once
the user is further into the section, we switch to the section title to
provide a persistent reminder of the user's location in the TOC.

From a front-end standpoint, this section title is always in the markup
and the swap happens with a CSS translateX function.

In addition to this markup, a layout bug has also been addressed which
ensures that the titles are truncated if they are longer than their
allotted space.

It should be noted that the section title value is dependent on commit
1e26ef412c78182e6dc3a289f5cb62597de1c21e (submitted as a pull request
in feature/pagination) as it gets the sectionId which is added to the
reader props in that commit. If the pagination branch is rejected for
any reason, this will need to be substituted.

commit b9f9a33a90dc88395f516a88f9d18bf4bf01bf99:
The scroll aware component has been
refactored to pass its state down to the first child component that it
wraps. Using this, the new "TextTitles" component can apply the state
to swap titles. Additionally, a click handler has been added that
un-toggles the section title for a set amount of time.

### DIFF
--- a/client/src/components/reader/Header.js
+++ b/client/src/components/reader/Header.js
@@ -9,6 +9,7 @@ export default class Header extends Component {
 
   static propTypes = {
     text: PropTypes.object,
+    sectionId: PropTypes.string,
     authenticated: PropTypes.bool,
     visibility: PropTypes.object,
     appearance: PropTypes.object,
@@ -25,6 +26,17 @@ export default class Header extends Component {
     setColorScheme: PropTypes.func,
     startLogout: PropTypes.func
   };
+
+  getSectionTitle(id) {
+    let title = '';
+    this.props.text.attributes.toc.forEach((section) => {
+      if (section.id === id) {
+        title = section.label;
+      }
+    });
+
+    return title;
+  }
 
   handleContentsButtonClick = () => {
     this.props.visibilityToggle('tocDrawer');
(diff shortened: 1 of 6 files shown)


client/src/components/reader/Header.js
```diff
@@ -23,9 +24,21 @@ export default class Header extends Component {
     incrementMargins: PropTypes.func,
     decrementMargins: PropTypes.func,
     setColorScheme: PropTypes.func,
+    scrollAware: PropTypes.object,
     startLogout: PropTypes.func
   };
 
+  getSectionTitle(id) {
```
zdavis: To me, the way this is being done has a bad smell. Traversing the TOC to get the title of a section for display in the header is wrong, since the section model has its title stored on it. In looking at the code, I can see why you're doing this—the reader container hasn't fetched the section from the section from the store, so you don't have access to the model. Rather than working around it as you've done here, the better solution would be to do some light refactoring. Instead of finding an indirect way to get the section in the header, I suspect we should move the section fetching to the reader container.

I'll make a story to do this refactoring, and put it in a separate commit.

client/src/components/reader/TextTitles.js
```diff
@@ -0,0 +1,56 @@
...
+export default class TextTitles extends Component {
+  static propTypes = {
+    textTitle: PropTypes.string,
+    sectionTitle: PropTypes.string,
+    showSection: PropTypes.bool
+  };
+
+  constructor(props) {
+    super(props);
+    console.log(this.props);
```
zdavis: Probably shouldn't have this console.log statement in the commit.

zdavis:
This looks good, and works as expected. Remove that console.log statement and I'll merge it.

zdavis:
I went ahead and removed the log call.

PR #43, opened 2016-05-09
Refactor API entity storage in client; add sign-up and basic profile editing

zdavis:


api/db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160228232735) do
+ActiveRecord::Schema.define(version: 20160511233642) do
```
zdavis: This can be safely ignored, as schema is automatically generated.

PR #44, opened 2016-05-17
Add functional sign-in and sign-up interfaces, and various other improvements

zdavis:


zdavis:
Going forward, hound should ignore the API specs. These hound complaints should not prevent this from being merged, IMO.

PR #62, opened 2016-10-21
Feature/headers icons

naomiyaki:
This trues up icons with headers so that they match the comps in size
and alignment. Also adjusts alignment/size for button nav and ensures
proper styling when there are two buttons.


zdavis:
Please update the Jest snapshot and force-push. I can help you with this if you're not sure how to do it.

zdavis:
Also, and this is a nitpick—the verb tense is wrong on your second commit message. "Removes timeout on button nav" should be "Remove timeout on button nav". Could you amend that commit and include it in your force push? Thanks.

PR #116, opened 2017-01-18
[C] Refactor date-fns into utility component

SMaxOwok:
[Delivers #137295167]

### DIFF
--- a/api/app/serializers/collection_serializer.rb
+++ b/api/app/serializers/collection_serializer.rb
@@ -1,18 +1,10 @@
 # Provides a serialization of a collection model.
 class CollectionSerializer < ActiveModel::Serializer
   cache key: "collection", expires_in: 3.hours
-  attributes :id, :title, :thumbnail_url, :created_month, :created_year, :description,
+  attributes :id, :title, :thumbnail_url, :created_at, :description,
              :resource_kinds
 
   has_many :resources
   belongs_to :project
-
-  def created_month
-    object.created_at.strftime("%-m")
-  end
-
-  def created_year
-    object.created_at.strftime("%-Y")
-  end
 
 end
--- a/api/app/serializers/resource_serializer.rb
+++ b/api/app/serializers/resource_serializer.rb
@@ -3,7 +3,7 @@ class ResourceSerializer < ActiveModel::Serializer
   cache key: "resource", expires_in: 3.hours
   attributes :title, :kind, :attachment_url, :attachment_file_name,
              :attachment_content_type, :attachment_file_size, :attachment_updated_at,
-             :created_month, :created_day, :created_year, :updated_at, :project_id,
+             :created_at, :updated_at, :project_id,
              :caption, :description, :fingerprint, :keywords, :alt_text,
              :copyright_status, :copyright_holder, :credit, :external_url, :external_id,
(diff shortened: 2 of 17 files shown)


zdavis:
There are a number of places where the API is returning createdMonth and createdYear. Could you refactor so that the API returns the model createdAt property (if it doesn't already) and the client uses that instead? This way, we can simplify the client-side code and standardize things a bit.

SMaxOwok:
All set!

client/src/components/global/Date.js
```diff
@@ -0,0 +1,49 @@
...
+      } else {
+        date = (
+          format(parse(this.props.date), this.props.format)
+        );
+      }
+    }
+
+    return date;
+  }
+
```
zdavis: No need to declare variables if you're just going to return them. Again, use guard clauses. Also, template strings are better than string concatenation in ES6. You can get this down to a two line method.

```
formatString(date) {
  if (!this.props.prefix || this.props.prefix.length === 0) return date;
  return `${this.props.prefix} ${date}`;
}
```

client/src/components/global/Date.js
```diff
@@ -0,0 +1,49 @@
...
+      output = date;
+    }
+
+    return output;
+  }
+
+  render() {
+    const date = this.formatDate();
+    return (
+      <span>{ this.formatString(date) }</span>
```
zdavis: We typically don't put spaces around the method call here. I might also wrap this up in a small method.
```
value() {
  return this.formatString(this.formatDate(this.props.date));
}

render() {
  return <span>{this.value()}</span>
}
```

zdavis: Note that the above snippet implies one small change to your code, which is to pass the date prop value to this.formatDate rather than to access it from within the formatDate method. This is a subtle but significant change. My approach makes the code less stateful and more functional. We pull the part where the object state is accessed (`this.props.date`) into its own method (`value()`). This means that we could execute formatDate and formatString with a different date value, rather than the this.props.date value. This is important in react, because it's not uncommon to want to see what happens to a prop before it becomes `this.props` on the object (in lifecycle methods, for example). By limiting where you access component state and using a functional style, your code will require less refactoring when you run into those situations. When I say "functional style," I mean writing methods that do not access this.props directly, but rather have it passed to them as arguments—a functional method is one that will always return the same results with the same set of arguments. A stateful method is one that returns different results based on the state of the object. Does that kind of make sense?

client/src/components/global/__tests__/Date-test.js
```diff
@@ -0,0 +1,27 @@
...
+
+describe('components', () => {
+  describe('Date', () => {
+    it('renders the date with correct prefix', () => {
+      const { output } = setup();
+      expect(output.type).toBe('span');
+      expect(output.props.children).toEqual('test January, 2017');
+    });
+  });
+});
```
zdavis: I'd like to see more tests here. What happens when the component is not passed a `date` prop? Does it work? Does it render nothing? What happens when it is not passed a format? Or when it is not passed a prefix? More tests would better illustrate the component's ability to work with different sets of props.

zdavis:
Looks really good. Thanks, Max. The only thing I see that needs to change at this point is the name of the test file. The file is still `Date-test.js`, but should be changed to `FormattedDate-test.js` in keeping with the component name change. Otherwise, great job.

PR #117, opened 2017-01-23
[F] Add backend stats

SMaxOwok:
[Delivers #136390675]


SMaxOwok:
Is not currently using the google analytics profile id user setting.

zdavis:
Let me know if you need any help fixing the linting issues. They must all be fixed before I can merge this PR.

api/app/controllers/api/v1/statistics_controller.rb
```diff
@@ -0,0 +1,16 @@
+module Api
+  module V1
+    # Statistics controller
+    class StatisticsController < ApplicationController
+
+      resourceful! Statistics, authorize_options: { except: [:index] } do
```
zdavis: We actually do need to authorize the index route. These statistics should only be accessible to admin users, not the general public.

zdavis: And there should be a test for the index route showing that a non-admin user can't access it, and that an admin user can.

api/app/controllers/api/v1/statistics_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+        Statistics.new
+      end
+
+      def index
+        render_single_resource(Statistics.new, location: "")
```
zdavis: Can location be `[:api, :v1, :statistics]`?

api/app/controllers/concerns/validation.rb
```diff
@@ -63,7 +63,8 @@ def text_params
 
   def settings_params
     params.require(:data)
-    attributes = [{ general: [:default_publisher, :default_place_of_publication] }]
+    attributes = [{ general: [:default_publisher, :default_place_of_publication,
+                              :ga_id] }]
```
zdavis: This style feels more natural to me:

```
attributes = [{ 
  general: [:default_publisher, :default_place_of_publication, :ga_id] 
}]
```

zdavis: And if that's too long, then this:
```
attributes = [{ 
  general: [
    :default_publisher, 
    :default_place_of_publication, 
    :ga_id
  ] 
}]
```

api/app/lib/statistics.rb
```diff
@@ -0,0 +1,104 @@
...
+  extend ActiveModel::Naming
+
+  def id
+    "statistics-set"
+  end
+
+  def initialize(_range = nil)
+    @client = Google::Apis::AnalyticsV3::AnalyticsService.new
+
+    @client.authorization = Signet::OAuth2::Client.new(
```
zdavis: You're using at least one (Signet), perhaps 2 (Google::Apis?) gem that is not declared in our gem file. It works because I have another gem that relies on this gem. If we were to remove that gem, this would break. Please make gem dependencies explicit in the API Gemfile.

api/app/lib/statistics.rb
```diff
@@ -0,0 +1,104 @@
...
+      token_credential_uri:  "https://accounts.google.com/o/oauth2/token",
+      audience:              "https://accounts.google.com/o/oauth2/token",
+      scope:                 "https://www.googleapis.com/auth/analytics.readonly",
+      issuer:                 ENV["GOOGLE_SERVICE_CLIENT_EMAIL"],
+      signing_key:            OpenSSL::PKey::RSA.new(ENV["GOOGLE_SERVICE_PRIVATE_KEY"])
+    ).tap(&:fetch_access_token!)
+
```
zdavis: Move the creation of the authorization code to a factory model, similar to what I do for drive sessions. Setting up the session isn't the concern of the statistics model, per se, and we may have other cases outside of statistics where we need to get this token, so you may as well abstract it to its own class.

api/app/lib/statistics.rb
```diff
@@ -0,0 +1,104 @@
...
+class Statistics
+  include ActiveModel::Validations
+  include ActiveModel::Conversion
+  extend ActiveModel::Naming
+
+  def id
+    "statistics-set"
+  end
+
+  def initialize(_range = nil)
```
zdavis: Do you do anything with _range? It'd be better if I could pass a range to this object when I initialize it.

api/app/lib/statistics.rb
```diff
@@ -0,0 +1,104 @@
...
+    @client = Google::Apis::AnalyticsV3::AnalyticsService.new
+
+    @client.authorization = Signet::OAuth2::Client.new(
+      token_credential_uri:  "https://accounts.google.com/o/oauth2/token",
+      audience:              "https://accounts.google.com/o/oauth2/token",
+      scope:                 "https://www.googleapis.com/auth/analytics.readonly",
+      issuer:                 ENV["GOOGLE_SERVICE_CLIENT_EMAIL"],
+      signing_key:            OpenSSL::PKey::RSA.new(ENV["GOOGLE_SERVICE_PRIVATE_KEY"])
```
zdavis: Confirm that these env vars exist in .env.sample. If they don't, they should be added.

Also, what happens if these variables aren't set, or if the authentication fails? Does this class handle it gracefully?

api/app/lib/statistics.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  def read_attribute_for_serialization(n)
+    send(n)
+  end
+
+  protected
+
+  def get_reader_count(start_date, end_date)
+    @client.get_ga_data(
+      ENV["GOOGLE_ANALYTICS_PROFILE_ID"], # analytics profile ID
```
zdavis: Isn't this the setting that is set in the admin view and stored in the DB? If so, you should be using that value, not a duplicate env var.

api/config/routes.rb
```diff
@@ -11,6 +11,7 @@
       resources :makers
       resources :resources
       resources :texts
+      resources :statistics, only: [:index]
```
zdavis: Seems like this should be a singular resource, not plural, and that your index code should really be in a #show method.

client/src/App.js
```diff
@@ -52,6 +54,15 @@ export default class App extends Component {
     this.store.dispatch({ type: 'CLIENT_LOADED', payload: {} });
     this.store.dispatch(currentUserActions.login);
     this.forceUpdate();
+    ReactGA.initialize('UA-90773269-1'); // Google Analytics Tracking ID
```
zdavis: I'll address the hard coded ID.

client/src/containers/backend/Dashboard.js
```diff
@@ -88,7 +96,11 @@ class DashboardContainer extends PureComponent {
                       {'Activity'} <i className="manicon manicon-pulse-small"></i>
                     </h3>
                   </header>
-                  <DashboardComponents.Activity />
+                  { this.props.statistics ?
```
zdavis: This check should be in the DashboardComponents.Activity component, not outside of it. By placing it outside of the component, the check has to happen every time it's used. Inside of it, it only happens once. In the render method of DashboardComponents.Activity, just return null if there's no this.props.statistics value.

client/src/components/backend/Dashboard/Activity.js
```diff
@@ -5,35 +5,48 @@ export default class Activity extends Component {
...
+
+    if (increase > 0) {
+      out = "+ " + increase.toString();
+    } else if (increase < 0) {
+      out = "- " + Math.abs(increase).toString();
+    } else {
+      out = "0";
+    }
+    return out;
+  }
```
zdavis: Simplify your code. Lines 12 through 23 can be reduced to 4 lines of flat code rather than 11 lines of nested code:

```
const increase = stats.readerIncrease;
if (increase > 0) return "+ " + increase.toString();
if (increase < 0) return "- " + Math.abs(increase).toString();
return 0;
```

client/src/components/backend/Dashboard/Activity.js
```diff
@@ -5,35 +5,48 @@ export default class Activity extends Component {
 
   static displayName = "Dashboard.Activity";
 
-  static propTypes = {};
+  static propTypes = {
+    statistics: PropTypes.object
```
zdavis: Define the shape of the object. For example:

```
  // An object taking on a particular shape
  optionalObjectWithShape: React.PropTypes.shape({
    color: React.PropTypes.string,
    fontSize: React.PropTypes.number
  })
```

client/src/components/backend/Dashboard/Activity.js
```diff
@@ -5,35 +5,48 @@ export default class Activity extends Component {
...
+    } else if (increase < 0) {
+      out = "- " + Math.abs(increase).toString();
+    } else {
+      out = "0";
+    }
+    return out;
+  }
 
   render() {
+    const stats = this.props.statistics.attributes;
```
zdavis: `if (!stats) return null;`

client/src/components/backend/Dashboard/Activity.js
```diff
@@ -5,35 +5,48 @@ export default class Activity extends Component {
...
         <tbody>
           <tr>
             <td>Texts added this week</td>
-            <td>2</td>
+            <td>{stats.newTextsCount}</td>
           </tr>
           <tr>
-            <td>Readers this weeks</td>
-            <td>2349</td>
+            <td>Readers this week</td>
```
zdavis: Good catch on the typo.

client/src/containers/backend/Settings/General.js
```diff
@@ -41,6 +41,12 @@ class SettingsGeneralContainer extends PureComponent {
             name="attributes[general][defaultPlaceOfPublication]"
             placeholder="Enter Default Place of Publication"
           />
+          <Form.TextInput
```
zdavis: Go ahead and add a masked input to this to ensure that the format is right.

zdavis:
Here's where the CI build failed:

https://travis-ci.org/ManifoldScholar/manifold/builds/194946163#L654

api/app/controllers/concerns/validation.rb
SMaxOwok: &#10003;

client/src/components/backend/Dashboard/Activity.js
SMaxOwok: &#10003;

client/src/components/backend/Dashboard/Activity.js
SMaxOwok: &#10003;

client/src/components/backend/Dashboard/Activity.js
SMaxOwok: &#10003;

client/src/containers/backend/Dashboard.js
SMaxOwok: &#10003;

client/src/components/backend/Dashboard/Activity.js
SMaxOwok: ✓

api/app/lib/statistics.rb
SMaxOwok: ✓

SMaxOwok: This is separate from the existing setting we have ('UA-00000-01'), which is the property tracking ID for logging views.  This is a user's profile ID used in the stat request.  That being said, I've moved this to a setting and am using that now.

api/app/lib/statistics.rb
SMaxOwok: ✓

api/app/lib/statistics.rb
SMaxOwok: ✓

api/app/controllers/api/v1/statistics_controller.rb
SMaxOwok: ✓

api/config/routes.rb
SMaxOwok: ✓

api/app/controllers/api/v1/statistics_controller.rb
SMaxOwok: ✓

api/app/lib/statistics.rb
SMaxOwok: ✓

api/app/lib/statistics.rb
zdavis: What does the profile ID look like? I wonder if it should be stored in the environment after all. How does one get one's profile ID from google?

client/src/containers/backend/Settings/General.js
zdavis: And don't hesitate to let me know if you want help. The library we're using for input masks is a little weird.

client/src/components/backend/Form/MaskedTextInput.js
```diff
@@ -13,7 +13,11 @@ export default class FormMaskedTextInput extends Component {
 
   static propTypes = {
     ...sharedPropsValidation,
-    mask: PropTypes.string.isRequired
+    mask: PropTypes.oneOfType(
```
zdavis: This is wrong. OneOfType takes one argument, which is an array of allowed types. You're passing it multiple arguments.

zdavis:
Moving to a new PR

PR #118, opened 2017-01-25
[B] Fix issue with switches only going one way

SMaxOwok:
[Delivers #138298383]


zdavis:
Going with https://github.com/ManifoldScholar/manifold/pull/122 instead of this one.

PR #120, opened 2017-01-26
[F] Allow user record to be deleted

SMaxOwok:
[Delivers #138340443]


SMaxOwok:
Doesn't currently re-render the users list after delete.

PR #133, opened 2017-01-30
[F] Add settings and reindexing into dev load

SMaxOwok:
[Delivers #138663937]
[Delivers #138667707]

### DIFF
--- a/api/app/services/demonstration/data_loader.rb
+++ b/api/app/services/demonstration/data_loader.rb
@@ -5,6 +5,11 @@ module Demonstration
   # Loads demo data into the Manifold installation
   class DataLoader
 
+    ENV_SETTINGS = {
+      general: %w(ga_profile_id ga_tracking_id).freeze,
+      theme: %w(typekit_id).freeze
+    }.freeze
+
     def initialize
       @logger = Logger.new(STDOUT)
       @logger.formatter = proc { |severity, _datetime, _progname, msg|
@@ -19,6 +24,8 @@ module Demonstration
       create_fake_users
       create_pages
       import_projects
+      ensure_settings
+      reindex_records
     end
 
     def import_projects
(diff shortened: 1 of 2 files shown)


api/app/services/demonstration/data_loader.rb
```diff
@@ -5,6 +5,11 @@ module Demonstration
   # Loads demo data into the Manifold installation
   class DataLoader
 
+    ENV_SETTINGS = {
+      general: %w(ga_profile_id ga_tracking_id),
```
zdavis: the arrays should be frozen as well.

api/app/services/demonstration/data_loader.rb
```diff
@@ -70,6 +77,25 @@ def create_pages
       end
     end
 
+    def ensure_settings
+      settings = Settings.instance
+      ENV_SETTINGS.each do |key, values|
+        values.each do |value|
+          env_setting = "MANIFOLD_SETTING_#{key}_#{value}".upcase
+          settings[key][value] = settings[key][value] || ENV[env_setting]
```
zdavis: Use ||= here.

```
irb(main):001:0> a = 1
=> 1
irb(main):002:0> a ||= 2
=> 1
irb(main):003:0> a
=> 1
irb(main):004:0> a = nil
=> nil
irb(main):005:0> a ||= 2
=> 2
irb(main):006:0> a
=> 2
```

zdavis:
Rebased onto development to address failing CI build

api/app/services/demonstration/data_loader.rb
```diff
@@ -70,6 +77,25 @@ def create_pages
       end
     end
 
+    def ensure_settings
+      settings = Settings.instance
+      ENV_SETTINGS.each do |key, values|
+        values.each do |value|
+          env_setting = "MANIFOLD_SETTING_#{key}_#{value}".upcase
+          settings[key][value] = settings[key][value] ||= ENV[env_setting]
```
zdavis: You didn't quite get my meaning here. This line should be:

`settings[key][value] ||= ENV[env_setting]`

api/app/services/demonstration/data_loader.rb
```diff
@@ -70,6 +77,25 @@ def create_pages
...
+          @logger.info("Setting updated: #{key}: #{value}".green)
```
zdavis: This log output occurs whether or not the setting was updated, and it doesn't include any information about what it was set to. Not very useful.

SMaxOwok: I took this out, but I wasn't sure if you wanted it in there and give context based feedback.  I can add it back in that manner if you'd rather.

zdavis:
Could you add a couple simple tests to this that prove:

1. The setting value is overwritten if it's already set
2. The setting value is set from env if it's blank
3. Setting these values works even when the settings object doesn't exist.

PR #134, opened 2017-01-31
[F] Update annotation popup with position smarts

naomiyaki:
Also add new, smaller design, and new stub buttons to add resources and
annotation functions.


zdavis:
Rebased onto development to address failing CI build

PR #138, opened 2017-01-31
[F] Embed resources in texts

zdavis:


client/src/components/reader/Section/ResourceViewer/Wrapper.js
```diff
@@ -0,0 +1,94 @@
...
+      const rect = marker.getBoundingClientRect();
+      if (rect.top > 0 && rect.top < window.innerHeight) {
+        const resourceIds = marker.getAttribute('data-resources').split(',');
+        resourceIds.forEach((id) => {
+          viewableMarkers.push({
+            id,
+            top: rect.top
+          });
+        });
+      }
```
zdavis: You can make 68-77 more concise with a guard clause (always better than an nested condition, IMO) and a map:

```
const top = marker.getBoundingClientRect().top;
if (top > 0 || top < window.innerHeight) return;
const resourceIds = marker.getAttribute('data-resources').split(',');
viewableMarkers.concat(resourceIds.map((id) => { {id, top } }))
```

client/src/components/reader/Section/ResourceViewer/Wrapper.js
```diff
@@ -0,0 +1,94 @@
...
+    });
+
+    this.setState({
+      viewableResourceMarkers: viewableMarkers
```
zdavis: If you were to call this variable `viewableResourceMarkers` instead of `viewableMarkers` then this could be more concise as well: `this.setState({ viewableResourceMarkers });`

client/src/components/reader/Section/ResourceViewer/Wrapper.js
```diff
@@ -0,0 +1,94 @@
...
+  componentDidUpdate(prevProps) {
+    // Update available resource markers if we have new props or state
+    if (prevProps !== this.props) {
+      this.updateAvailableResourceMarkers();
+      this.updateViewableResourceMarkers();
+    }
+  }
+
+  updateAvailableResourceMarkers() {
+    if (this.props.body) {
```
zdavis: Perfect place for a guard clause.
`if (!this.props.body) return;`

client/src/components/reader/Section/ResourceViewer/Wrapper.js
```diff
@@ -0,0 +1,94 @@
...
+    window.addEventListener('scroll', this.throttledUpdateViewableResources);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('scroll', this.throttledUpdateViewableResources);
+  }
+
+  componentDidUpdate(prevProps) {
+    // Update available resource markers if we have new props or state
+    if (prevProps !== this.props) {
```
zdavis: This check shouldn't be necessary. Because it's a pure component, I wouldn't expect it to update if props were the same.

client/src/components/reader/Section/ResourceViewer/Wrapper.js
```diff
@@ -0,0 +1,94 @@
...
+
+  constructor() {
+    super();
+    this.state = {
+      availableResourceMarkers: [],
+      viewableResourceMarkers: []
+    };
+
+    this.updateAvailableResourceMarkers = this.updateAvailableResourceMarkers.bind(this);
+    this.updateViewableResourceMarkers = this.updateViewableResourceMarkers.bind(this);
```
zdavis: These probably don't need to be bound, since they're never called from the context of an event handler. It's OK if they are, but it's not really necessary.

PR #143, opened 2017-02-03
[F] Ingests a word/google doc as HTML

SMaxOwok:
[Delivers #114696225]
[Delivers #138742601]

### DIFF
--- a/api/app/services/ingestor.rb
+++ b/api/app/services/ingestor.rb
@@ -55,7 +55,7 @@ module Ingestor
     # @private
     # @return [Array] Array with [basename, ingestion, strategy]
     def start(path, creator)
-      unless File.exist?(path)
+      unless File.exist?(path) || path.start_with?("https") || path.start_with?("http")
         raise Ingestor::IngestionFailed,
               "Could not find ingestion source"
       end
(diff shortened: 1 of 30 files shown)


api/app/services/ingestor.rb
```diff
@@ -55,7 +55,7 @@ def logger
     # @private
     # @return [Array] Array with [basename, ingestion, strategy]
     def start(path, creator)
-      unless File.exist?(path)
+      unless File.exist?(path) || path.start_with?("https")
```
zdavis: What if it starts with HTTP?

api/app/services/ingestor/strategy/google_doc/strategy.rb
```diff
@@ -0,0 +1,53 @@
...
+            @ingestion.source_path,
+            @ingestion.logger
+          )
+          @logger = @ingestion.logger ||
+                    Naught.build { |config| config.mimic Logger }
+        end
+
+        # Return true if the url is valid
+        def self.can_ingest?(ingestion)
+          inspector = ::Ingestor::Strategy::GoogleDoc::Inspector::GoogleDoc.new(
```
zdavis: DRY up your code into a private method. The 4 lines that return the inspector are repeated here and could be consolidated.

PR #144, opened 2017-02-04
[C] Cache statistics for 3 hours

SMaxOwok:
[Delivers #138865107]


SMaxOwok:
screen shot 2017-02-03 at 4 31 13 pm

zdavis:
Not quite the same as a server side cache, but it'll do for now I think.

PR #146, opened 2017-02-07
[F] Add spine with text section order to text

SMaxOwok:
[Delivers #113292297]


api/app/services/ingestor/strategy/google_doc/inspector/google_doc.rb
```diff
@@ -63,6 +63,13 @@ def ingestion_source
...
+          def spine_ids
+            ingestion_source.map do |item|
+              ::Ingestor::Strategy::Gitbook::Inspector::TextSection.new(item, self)
+                                                                   .source_identifier
+            end
+          end
+
           # rubocop:disable Metrics/LineLength
           def ingestion_source_inspectors
             ingestion_source.map do |source|
```
zdavis: Also not part of the PR.. just seems better to do `[ingestion_source].map` here and have ingestion_source return one thing.

api/app/services/ingestor/strategy/word/inspector/word.rb
```diff
@@ -110,9 +111,21 @@ def stylesheet
...
 
-          def text_section_inspectors
+          def spine_ids
+            text_ingestion_source.map do |item|
+              ::Ingestor::Strategy::Gitbook::Inspector::TextSection.new(item, self)
+                                                                   .source_identifier
+            end
+          end
+
+          def text_ingestion_source
```
zdavis: Another case of a singular method returning an array of things. This is a bad code smell. Refactor this if necessary to make your naming more predictable, unless I'm missing some compelling reason to have it be this way.

api/app/services/ingestor/strategy/word/inspector/word.rb
```diff
@@ -110,9 +111,21 @@ def stylesheet
...
             name = "#{basename}.html"
-            ingestion_source_paths.reject { |p| File.basename(p) != name }.map do |path|
+            ingestion_source_paths.reject { |p| File.basename(p) != name }
+          end
+          memoize :text_ingestion_source
+
+          def text_section_inspectors
+            text_ingestion_source.map do |path|
```
zdavis: ibid.

api/db/schema.rb
```diff
@@ -303,6 +303,7 @@
     t.uuid     "creator_id"
     t.uuid     "start_text_section_id"
     t.integer  "position"
+    t.string   "spine",                              array: true
```
zdavis: Right now this field default to nil, rather than to []. Is that the behavior we want?

api/spec/models/user_spec.rb
```diff
@@ -81,7 +81,7 @@
       User.searchkick_index.refresh
     end
 
-    it "by last name" do
+    it "by first name" do
```
zdavis: good catch.

api/spec/services/ingestor/creator/stylesheets_spec.rb
```diff
@@ -26,7 +26,7 @@ def double_builder(num)
     expect(creator).to respond_to(:warn)
   end
 
-  it "creates an stylesheet for each inspector" do
+  it "creates a stylesheet for each inspector" do
```
zdavis: another good catch.

client/src/components/reader/Section/Pagination.js
```diff
@@ -6,22 +6,22 @@ export default class Pagination extends Component {
...
-    textSections: PropTypes.array
+    spine: PropTypes.array
   };
 
   getSiblingSection(id, shift) {
     let siblingSection = false;
-    this.props.textSections.forEach((section, i) => {
-      if (section.id === id && this.props.textSections[i + shift]) {
-        siblingSection = this.props.textSections[i + shift];
+    this.props.spine.forEach((spineId, i) => {
```
zdavis: What happens when spine is nil?

zdavis: ForEach seems like an inefficient approach here. Seems like you can just use findIndex (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) to get the index of the current section, then add shift to that to get the section you want.

SMaxOwok:
This is ready to get checked out again.

I left spine's default as nil, but added validation on the model.  I think it's beneficial for us to leave it as nil since it should never be empty coming from the ingestor.

PR #148, opened 2017-02-07
[F] Use text's rights attribute for reader footer

SMaxOwok:
[Delivers #139246875]


zdavis:
Could you include a simple snapshot test for the footer component?

zdavis:
This isn't ready to be accepted — it throws console errors as currently implemented due to nesting div tags within P tags. Look at the JSX carefully and see if you can tighten it up.

PR #150, opened 2017-02-08
[F] Display next section title in reader

SMaxOwok:
[Delivers #113292297]

### DIFF
--- a/api/app/serializers/text_partial_serializer.rb
+++ b/api/app/serializers/text_partial_serializer.rb
@@ -3,13 +3,19 @@ class TextPartialSerializer < ActiveModel::Serializer
   # cache key: "text_partial", expires_in: 3.hours
   attributes :id, :title, :creator_names, :unique_identifier, :cover_url, :created_at,
              :start_text_section_id, :published, :annotations_count, :highlights_count,
-             :bookmarks_count, :age, :position, :publication_date, :spine, :rights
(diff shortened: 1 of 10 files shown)


api/app/serializers/text_partial_serializer.rb
```diff
@@ -12,6 +13,10 @@ def start_text_section_id
     object.start_text_section_id ||= object.spine[0]
   end
 
+  def sections_map
+    object.spine.map { |id| Hash[id: id.to_s, name: object.text_sections.find(id).name] }
```
zdavis: This should be more forgiving. If, for example, a text section was deleted from a text, any call to this method would trigger a record not found exception due to `text_sctions.find`. Assuming spine is an array of text section IDs, it might best to get the union of spine and text_sections.pluck(:id) and base the map on that... that way if a text section is removed, it will not longer show up in the spine. You can also use `find_by_id` instead of find, but then you could end up with blanks in the section map.

client/src/theme/Components/reader/_section-next-section.scss
```diff
@@ -0,0 +1,28 @@
+.section-next-section {
```
zdavis: @naomiyaki you cool with this CSS?

SMaxOwok:
Updated the sections_map to be less brittle.

PR #158, opened 2017-02-15
[F] Add ability to associate users with makers

SMaxOwok:
[Delivers #139228287]

### DIFF
--- a/api/app/controllers/api/v1/makers_controller.rb
+++ b/api/app/controllers/api/v1/makers_controller.rb
@@ -3,6 +3,10 @@ module Api
     # makers controller
     class MakersController < ApplicationController
 
+      INCLUDES = %w(
+        users
+      ).freeze
+
       resourceful! Maker, authorize_options: { except: [:index, :show] } do
         Maker
           .filter(with_pagination!(maker_filter_params))
@@ -18,7 +22,7 @@ module Api
 
       def show
         @maker = load_maker
-        render_single_resource(@maker)
+        render_single_resource(@maker, include: INCLUDES)
       end
 
       def create
@@ -29,7 +33,7 @@ module Api
       def update
         @maker = load_and_authorize_maker
         ::Updaters::Maker.new(maker_params).update(@maker)
-        render_single_resource(@maker)
+        render_single_resource(@maker, include: INCLUDES)
       end
 
       def destroy
--- a/api/app/lib/updaters.rb
+++ b/api/app/lib/updaters.rb
@@ -1,54 +1,66 @@
 require "active_support/core_ext/string"
+require "active_support/concern"
+require "active_support/callbacks"
 
 # Updaters are responsible for mapping JSON-API structured params to model updates.
 # This module acts as a base mix-in for model-specific updaters.
 module Updaters
-  attr_accessor :id, :type, :data
+  extend ActiveSupport::Concern
+
+  included do
+    include ActiveSupport::Callbacks
+    attr_accessor :id, :type, :data, :attributes, :relationships
+    define_callbacks :update_attributes, :update_relationships, :update, :save
+  end
 
   def initialize(params)
     @attributes = params[:data][:attributes] || {}
     @relationships = params[:data][:relationships]
   end
 
-  def attributes
-    @attributes
-  end
-
-  def relationships
-    @relationships
-  end
-
   def update_without_save(model)
-    if attributes
-      attr = adjusted_attributes
-      attachmentize_attributes!(attr)
-      model.assign_attributes(attr)
+    @model = model
+    run_callbacks "update" do
+      assign_attributes!(model)
+      update_relationships!(model)
     end
-    update_relationships(model)
     model
   end
 
   def update(model)
-    if attributes
-      attr = adjusted_attributes
-      attachmentize_attributes!(attr)
-      model.assign_attributes(attr)
-    end
-    update_relationships(model)
-    saved = model.save
-    try(:post_update, model) if saved
-    model.reload if model.id && saved
+    @model = model
+    update_without_save(model)
+    save_model(model)
     model
   end
 
   def post_update(model)
+    @model = model
     update_position(model)
   end
 
   protected
 
+  def save_model(model)
+    @saved = false
+    run_callbacks "save" do
+      @saved = model.save
+    end
+    model.reload if model.id && @saved
+    @saved = false
+  end
+
   def attachment_fields
     []
+  end
+
+  def assign_attributes!(model)
+    return unless attributes
+    run_callbacks "update_attributes" do
+      attr = adjusted_attributes
+      attachmentize_attributes!(attr)
+      model.assign_attributes(attr)
+    end
   end
 
   # rubocop:disable Metrics/AbcSize
@@ -70,14 +82,16 @@ module Updaters
   # rubocop:enable Metrics/AbcSize
 
   # rubocop:disable Metrics/CyclomaticComplexity
-  def update_relationships(model)
+  def update_relationships!(model)
     return unless relationships
-    relationships.to_h.each do |name, relationship|
-      models = relationship[:data]
-      if models.respond_to?(:has_key?) || models.nil?
-        update_belongs_to(model, name, models) && next
+    run_callbacks "update_relationships" do
+      relationships.to_h.each do |name, relationship|
+        models = relationship[:data]
+        if models.respond_to?(:has_key?) || models.nil?
+          update_belongs_to(model, name, models) && next
+        end
+        update_has_many(model, name, models) && next if models.respond_to?(:each)
       end
-      update_has_many(model, name, models) && next if models.respond_to?(:each)
     end
   end
   # rubocop:enable Metrics/CyclomaticComplexity
@@ -102,7 +116,6 @@ module Updaters
   def update_has_many(model, name, to_add)
     value = relationship_map(model, name, to_add)
     model.send("#{name}=", value)
-    model.send(name.to_s).try(:association_sort, value)
   end
 
   def relationship_map(model, name, models)
@@ -124,6 +137,18 @@ module Updaters
   def remove_relative_position!(attributes)
     return unless %w(up down top bottom).include?(attributes[:position])
     attributes.delete(:position)
+  end
+
+  # This method is a utility method, used to sort creator and contributor
+  # collaborators after saving the model.
+  def sort_collaborators(type)
+    makers = relationships.to_h.dig "#{type}s", :data
+    return unless makers && !makers.empty?
+    @model.send("#{type}_collaborators").each do |collaborator|
+      index = makers.find_index { |c| c[:id] == collaborator.maker_id }
+      position = index + 1
+      collaborator.set_list_position(position)
+    end
   end
 
   def update_position(model)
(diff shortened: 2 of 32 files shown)


zdavis:
It seems like there's a regression in this pull request. Are you still able to edit makers in the backend in this branch? Looks like there's a bind call in a constructor that needs to be removed, perhaps.

api/app/controllers/api/v1/makers_controller.rb
```diff
@@ -3,8 +3,13 @@ module V1
     # makers controller
     class MakersController < ApplicationController
 
+      INCLUDES = %w(
+        users
+      ).freeze
+
       resourceful! Maker, authorize_options: { except: [:index, :show] } do
         Maker
+          .includes(:users)
```
zdavis: This seems wrong to me, per my comment on the pivotal tracker story. We don't want to expose users on the serialized maker, as there's no reason for the client application to know which users belong to a given maker (it's the other side of the relationship that's more useful)

api/app/lib/updaters.rb
```diff
@@ -102,7 +102,7 @@ def update_belongs_to(model, name, to_add)
   def update_has_many(model, name, to_add)
     value = relationship_map(model, name, to_add)
     model.send("#{name}=", value)
-    model.send(name.to_s).try(:sort, value)
+    model.send(name.to_s).try(:sort_makers, value)
```
zdavis: I don't think this is a prudent change. Sure, it will make it work for this one case, but it's going to break sorting on has_many relationships for every other type of association. This update class provides some automation and abstraction around updating models from params — as such, only relatively generic method calls can exist here.

If you reall need maker-specific code, you should override the update_has_many method in  Updaters::Maker.

zdavis: Or perhaps we should add a post update relationship hook of some kind. Happy to discuss.

api/app/models/concerns/collaborative.rb
```diff
@@ -24,7 +24,7 @@ module Collaborative
     has_many :creators,
              through: :creator_collaborators,
              source: "maker" do
-      def sort(makers)
+      def sort_makers(makers)
         makers.each_with_index do |maker, index|
```
zdavis: Why is this method being renamed?

api/app/serializers/maker_partial_serializer.rb
```diff
@@ -0,0 +1,9 @@
+# Provides a partial serialization of a maker model.
+class MakerPartialSerializer < ActiveModel::Serializer
+  meta(partial: false)
+
+  cache key: "maker", expires_in: 3.hours
```
zdavis: We're going to stop caching these responses for now, so go ahead and remove this. I removed it from all the other serializers. For the reason behind this, see the warning here:

https://github.com/rails-api/active_model_serializers/blob/master/docs/general/caching.md

api/app/serializers/maker_serializer.rb
```diff
@@ -1,8 +1,4 @@
 # Provides a partial serialization of a maker model.
-class MakerSerializer < ActiveModel::Serializer
-  meta(partial: false)
-
-  cache key: "maker", expires_in: 3.hours
-  attributes :id, :first_name, :last_name, :middle_name, :display_name, :avatar_url,
-             :full_name
+class MakerSerializer < MakerPartialSerializer
+  has_many :users, serializer: UserPartialSerializer
```
zdavis: Given that we never want to expose the user on the maker, I think we can go ahead and collapse the MakerSerializer and the MakerPartialSerializer back down to a single serializer. I assume you did this just to break out the include, which shouldn't be necessary.

api/app/serializers/user_partial_serializer.rb
```diff
@@ -0,0 +1,8 @@
+# Serializes a Text model
+class UserPartialSerializer < ActiveModel::Serializer
```
zdavis: Same comment applies here — I don't want to go down the road of a partial and full serializer unless absolutely necessary, and I don't think it is in this case.

client/src/components/backend/Form/HasMany.js
```diff
@@ -10,7 +10,8 @@ export default class FormMakers extends PureComponent {
 
   static propTypes = {
     label: PropTypes.string.isRequired,
-    onNew: PropTypes.func.isRequired,
+    includeCreate: PropTypes.bool,
```
zdavis: Where does this includeCreate prop get used? Is it to disable creation on the maker side of the relationship?

zdavis: Do we really need includeCreate? Does it make sense to act as thought includeCreate == false if no onNew prop is passed? What would happen if includeCreate == true and there's an onNew callback? Seems like we're setting two switches where we only need one. 

We should also keep our naming consistent. Is "new" the same thing as "create"? If we do end up needing two switches here, I think we should call them "onNew" and "allowNew"

client/src/containers/backend/Form/UserClaims.js
```diff
@@ -0,0 +1,125 @@
...
+import { makersAPI, usersAPI } from 'api';
+import { connect } from 'react-redux';
+import { entityStoreActions } from 'actions';
+import { entityUtils } from 'utils';
+import get from 'lodash/get';
+import capitalize from 'lodash/capitalize';
+
+const { request, flush, requests } = entityStoreActions;
+
+class FormUserClaims extends Component {
```
zdavis: In general, we want to avoid making new connected components when possible. I'm a little on the fence about this one. On the one hand, we could hoist these negotiations with the API up to People/Users/Edit, which is the connected component. I can see why you did it this way, in any case, because you want to share the connected logic between the two forms that are using this component. I'm not crazy about this solution, but I also don't have a better approach to offer, so for now let's leave it.

client/src/containers/backend/People/Makers/Edit.js
```diff
@@ -25,6 +26,7 @@ class MakersEditContainer extends PureComponent {
     this.state = {
       confirmation: null
     };
+    this.updateUsers = this.updateUsers.bind(this);
```
zdavis: You're binding a method that doesn't exist, which is triggering an error.

zdavis: Note that this updateUsers function is referenced elsewhere in this component, which seems wrong.

client/src/containers/backend/Form/UserClaims.js
zdavis: Actually, as I look through this, I'm confused — it seems like this container is only used on Users/Edit.js. If it's not also useful on the other side of the equation, in Makers/Edit.js, then why are we abstracting this into its own container? Maybe the maker side isn't finished?

client/src/containers/backend/People/Makers/Edit.js
```diff
@@ -129,6 +132,20 @@ class MakersEditContainer extends PureComponent {
             name="attributes[avatar]"
             remove="attributes[removeAvatar]"
           />
+          <Form.HasMany
```
zdavis: Wondering why you use Form.HasMany on the maker side, and a dedicated container on the user side.

client/src/containers/backend/People/Users/Edit.js
```diff
@@ -3,16 +3,17 @@ import { connect } from 'react-redux';
 import { Drawer, Dialog } from 'components/backend';
 import { entityStoreActions } from 'actions';
 import { entityUtils } from 'utils';
-import usersAPI from 'api/users';
+import { usersAPI, makersAPI } from 'api';
 const { select, meta } = entityUtils;
-const { request } = entityStoreActions;
+const { request, flush } = entityStoreActions;
```
zdavis: Why add the flush action, if you're not using it?

client/src/containers/backend/People/Users/Edit.js
```diff
@@ -3,16 +3,17 @@ import { connect } from 'react-redux';
 import { Drawer, Dialog } from 'components/backend';
 import { entityStoreActions } from 'actions';
 import { entityUtils } from 'utils';
-import usersAPI from 'api/users';
+import { usersAPI, makersAPI } from 'api';
```
zdavis: Why import the makersAPI if you're not using it?

client/src/containers/backend/People/Users/Edit.js
```diff
@@ -3,16 +3,17 @@ import { connect } from 'react-redux';
...
 const { select, meta } = entityUtils;
-const { request } = entityStoreActions;
+const { request, flush } = entityStoreActions;
 import { Form } from 'components/backend';
 import { Form as FormContainer } from 'containers/backend';
 import { browserHistory } from 'react-router';
+import get from 'lodash/get';
```
zdavis: Why import get if you're not using it?

client/src/containers/backend/People/Users/Edit.js
```diff
@@ -127,6 +127,13 @@ class UsersEditContainer extends PureComponent {
             name="attributes[lastName]"
             placeholder="Last Name"
           />
+          <FormContainer.UserClaims
```
zdavis: This needs to be moved outside of the attributes form. Having it above the save button suggests that the changes made require saving, which they do not. It should go below the save button. Look at page 10 of the backend UI comps to see header styles for breaking this out from the main form.

api/app/lib/updaters.rb
SMaxOwok: I think my confusion here came from getting an argument error on `sort`.  Right now, the only time I see `update_has_many` being hit is when authors/collaborators are added to a project.  Those succeed because they have their own sort methods defined that take an argument, but I think any model that hits `update_has_many` will respond to `.try(:sort)`.

It seems that when the associations in this PR are made, they hit the `update_has_many` method and try to call Ruby's sort with an argument, resulting in the error.

`model.send("makers").try(:sort)
=> [#<Maker id: "347c5c38-cde4-4a1...`
 `model.send("makers").try(:sort, model2)
ArgumentError: wrong number of arguments (given 1, expected 0)
`

Does it make sense to change this symbol to something like, `association_sort`?

SMaxOwok:
@zdavis this should be ready to be reviewed again.  I pulled the update methods out of the UserClaims component, which I think is in-line with your comment about keeping it disconnected.  Also please check my comment on `api/app/lib/updaters.rb` above.

zdavis:
Rebased onto development

client/src/components/backend/Form/HasMany.js
```diff
@@ -155,7 +178,7 @@ export default class FormMakers extends PureComponent {
                 return `${option.attributes.firstName} ${option.attributes.lastName}`;
               }
             }
-            onNew={this.onNew}
+            onNew={this.props.onNew}
```
zdavis: This change introduces a regression that breaks the ability to add a new maker to a project on the project collaborators view in the backend. Prior to this change, when the predictiveInput triggers the "onNew" callback, it's triggering the `onNew` method in the HasMany component. That's important, because that method will call this.props.onNew (the onNew method that was passed to HasMany), and then it will call this.props.onChange, because the assumption is that the new record is also now the selected record.

This change just passes the hasMany onNew prop directly to the predictive input, which breaks this behavior.

zdavis:
screen shot 2017-02-17 at 11 13 06 am

When I enter an invalid name in the project hasMany component, I see the error message. In your UserClaims component, it fails silently.

SMaxOwok:
Fixed the missing error messages issue.

PR #159, opened 2017-02-16
[F] Add searching and filtering for project resources

SMaxOwok:
[Delivers #139241941]


client/src/components/frontend/ResourceList/Filters.js
```diff
@@ -1,17 +1,62 @@
...
   static displayName = "ResourceList.Filters";
 
   static propTypes = {
-    kinds: PropTypes.array
+    kinds: PropTypes.array,
+    tags: PropTypes.array,
+    filterChangeHandler: PropTypes.func.isRequired
   };
 
-  constructor() {
```
zdavis: You've changed this component so that it needs `kinds` and `tags` props to function properly, but you haven't gone and made the change to the parent that uses this component to ensure that it still functions. Take a look at `components/frontend/ResourceCollection/Detail`. That component includes `ResourceList.Filters`, but doesn't pass it tags — so when you drill into a collection on a project, you end up with the map in `ResourceList.Filters` render method throwing an exception.

Think about how you can make the filters component more flexible and error-resistant. If it's not passed tags or kinds, shouldn't it just not render those parts of the filters? Or, if it's going to always render them, then at least set default props on it:

```
static defaultProps = {
  kind: [],
  tags: []
};
```

api/app/controllers/api/v1/collections/relationships/resources_controller.rb
```diff
@@ -8,8 +8,9 @@ class ResourcesController < ApplicationController
           before_action :set_collection, only: [:index]
 
           resourceful! Resource, authorize_options: { except: [:index, :show] } do
-            @collection.resources
-                       .query(with_pagination!(resource_filter_params))
+            filter_params =
+              with_pagination!(resource_filter_params).merge(collection: @collection.id)
+            Resource.filter(with_pagination!(filter_params))
```
zdavis: Look at the changes I made to how filtering works. With those changes, I think it should be possible to change this call to look like this:

`Resource.filter(with_pagination!(filter_params), @collection.resources)`

I think this is a cleaner solution, as we can use the association scope rather than fudging the params. Try it and confirm that it works.

api/app/models/concerns/filterable.rb
```diff
@@ -3,14 +3,63 @@ module Filterable
```
zdavis: Look at my extended commit message for my thinking on this. I like where you're going with the filtering—getting the IDs from the active record query and passing them to elastic search if necessary—but having that logic spread out across all our models is going to get kind of crazy. I've refactored the filterable concern so that all filtering (and pagination) logic is in one place. The only thing that stays on the models are the named scopes, which seems correct. This is still a little complex, and we may want to tweak it, but at least it's now DRY.

api/app/models/concerns/paginated.rb
```diff
@@ -1,25 +0,0 @@
-# Model concern that includes pagination scope
```
zdavis: I moved all this into `Filterable`.

api/app/models/project.rb
```diff
@@ -135,6 +117,12 @@ def resource_kinds
       .to_a.pluck(:kind)
   end
 
+  def resource_tags
```
zdavis: Hmmm.... I don't love this approach, as it introduces what could potentially be a hefty query for every single project serialization. It's fine for now, but I wonder if we shouldn't instead do a separate request for tags by project when we need that.

api/app/models/resource.rb
```diff
@@ -63,6 +63,22 @@ class Resource < ApplicationRecord
...
+  }
+  scope :by_kind, lambda { |kind|
+    return all unless kind.present?
+    where(kind: kind)
+  }
+  scope :by_collection, lambda { |collection|
+    return all unless collection.present?
+    where(collection: collection)
+  }
+  scope :with_order, lambda { |by|
```
zdavis: We may as well handle order as a named scope, just like we do with the where clauses.

api/spec/models/resource_spec.rb
```diff
@@ -25,4 +25,24 @@
...
+      @resource_c = FactoryGirl.create(:resource, title: "resource_c", project: @project_b)
+    end
+
+    it "to only include those belonging to a project" do
+      results = Resource.filter({project: @project_a})
+      expect(results.length).to be 2
+      results = Resource.filter({project: @project_b})
+      expect(results.length).to be 1
+    end
+
```
zdavis: I just added a super basic test... would be nice to have a few more.

client/src/components/frontend/Project/Resources.js
```diff
@@ -3,12 +3,13 @@ import { ResourceList } from 'components/frontend';
...
 
-  static displayName = "Project.Resources"
+  static displayName = "Project.Resources";
 
   static propTypes = {
     project: PropTypes.object,
     resources: PropTypes.array,
-    pagination: PropTypes.object
+    pagination: PropTypes.object,
+    filterChange: PropTypes.func.isRequired
```
zdavis: Double check to be sure that everything currently using this component is passing this new required prop in.

client/src/components/frontend/ResourceList/Filters.js
```diff
@@ -1,17 +1,62 @@
...
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (get(prevState, 'filter') !== get(this.state, 'filter')) {
+      this.props.filterChangeHandler(this.state.filter);
+    }
+  }
+
+  setFilters(event, label) {
+    const key = event.target.value;
```
zdavis: Seems confusing to refer to value as key here.

client/src/components/frontend/ResourceList/Filters.js
```diff
@@ -1,17 +1,62 @@
...
+    const filter = Object.assign({}, this.state.filter);
+    const inputs = Object.assign({}, this.state.inputs);
+    if (key === "") {
```
zdavis: There's something about this method I'm not crazy about... it feels verbose, for one. Also, there's the fact that this will always update state, even if nothing has changed in the inputs or the filters. I'm not saying rewrite it, I just think it could probably be tightened up a bit.

client/src/components/frontend/ResourceList/Filters.js
```diff
@@ -41,36 +93,40 @@ export default class ResourceListFilters extends Component {
...
-                value="default"
-                disabled="disabled"
+                value=""
               >
                 Tag:
               </option>
-              <option>Japan</option>
-              <option>Photography</option>
-              <option>Second Interview</option>
+              {this.props.tags.map((tag, index) => {
```
zdavis: If `tags` does not exist, this will choke.

SMaxOwok:
I added a couple more tests and refactored the setFilters method to be a case switch statement and only update when there is a value/label.  I'm not totally certain about your comment about always updating state, since it only fires when a filter is changed.  Would you mind clarifying?

client/src/components/frontend/Resource/Card.js
```diff
@@ -74,12 +75,21 @@ export default class ResourceCard extends Component {
...
+    return out.id;
+  }
+
   detailUrl() {
     const context = this.props.context;
     if (context.type === "collections") {
-      const crid = this.props.resource.id;
       const pid = context.relationships.project.id;
       const cid = context.id;
+      const crid = this.getCollectionResourceId(this.props.resource);
```
SMaxOwok: I had to include the CollectionResources for a Resource because of the different IDs.  Is there a better way we could handle this or does this seem appropriate? It seems like either way we get the resources (by CollectionResource or Resources) we need to reference the association.

zdavis:
Pressing return in the search input submits the form and reloads the page. You need to capture the form onSubmit event.
screen shot 2017-02-24 at 1 02 32 pm

zdavis:
In general the search input needs additional testing and review. Give it another thorough pass.

- Changing sort has no impact on the collection view.
- The keyword search shouldn't be typeahead. It needs to require a submit to update results. 
- If you type the letter "a" (or probably any single letter) into the keyword input, you can't delete it.

client/src/components/frontend/Resource/Card.js
zdavis: That seems OK.

SMaxOwok:
Addressed those issues and fixed the missing collection resource default sorting order.

PR #160, opened 2017-02-20
[WIP] Implement password reset

SMaxOwok:


api/app/controllers/api/v1/passwords_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+      include Password
+
+      def new
+        user = User.find_by(email: params[:email])
+        return user_not_found_error unless user
+        user.process_reset_password_request
+      end
+
+      def update
+        user = User.by_reset_token(params[:reset_token]).first
```
zdavis: I think I'd roll up the .first into the `by_reset_token` method, unless there's a compelling reason not to.

api/app/controllers/api/v1/passwords_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+        return user_not_found_error unless user
+        return invalid_token_error unless user.valid_token?
+        user.update_password(params[:password], params[:password_confirmation])
+        if user.save
+          render status: :no_content
+        else
+          render json: { errors: user.errors.full_messages },
```
zdavis: This seems wrong. Error responses need to follow the correct format for JSON API. Look at the render_single_resource method in the resourceful concern. If there is an error on a resource, we return it with the error serializer, that gives the error a specific format. You're following a different convention here, which feels hacky.

api/app/controllers/concerns/api/v1/password.rb
```diff
@@ -0,0 +1,16 @@
...
+  module V1
+    # Includes password resetting methods
+    module Password
+      def invalid_token_error
+        render json: { errors: ["Invalid or expired token"] },
+               status: :unprocessable_entity
+      end
+
+      def user_not_found_error
+        render json: { errors: ["No user with that email found"] },
```
zdavis: I think we need to rethink how this resource works... the modeling here doesn't feel right to me (my fault). 

controllers/api/v1/users/reset_password_tokens_controller#create <-- takes email as an argument, always responds with 204 no content, even when email is not found. Whether or not email was found, user sees a message like this: "If a matching account was found an email was sent to user@email.com to allow you to reset your password."

controllers/api/v1/users/password/#update <-- takes token, pw, pw_confirm as arguments and updates user's pw. If successful, we need to somehow return the auth token and user object, exactly like what is returned by the login request (TokensController#create). You may need to move some token response logic into a concern. A user should be logged in after a PW change, so we need to return the same response as with login.

There will also need to be some store/reducer logic (see the auth reducer) to handle this. If a reset password action succeeds, it needs to be reduced similar to how a login access is reduced. This might get tricky, so ask me if you get confused.

client/src/components/global/SignInUp/PasswordForgot.js
```diff
@@ -1,28 +1,102 @@
...
+    this.submissionError = this.submissionError.bind(this);
+  }
+
+  componentWillUnmount() {
+    this.props.dispatch(flush(['request-reset-password']));
+  }
+
+  handleSubmit(event) {
+    event.preventDefault(event.target);
+    this.props.dispatch(request(passwordsAPI.new(this.state.email), 'request-reset-password')).promise.then(() => {
```
zdavis: Resetting pw is special and can't be handled as a generic API request using the request helper. We need to make an authorization action and reducer method to handle this, because after the request is resolved, login should occur. Look at how login works, and see if you can figure out how pw reset might work.

api/app/controllers/api/v1/passwords_controller.rb
SMaxOwok: This is a weird one... Moving the `.first` to the `.by_reset_token` method returns what I want in the console, but returns many records in the controller.
`user = User.by_reset_token(params[:reset_token])
  puts user.count => 42`

When it's in the controller, I get the correct user.

SMaxOwok: Potential alternative:
`  class_methods do
    def by_reset_token(token)
      return nil unless token.present?
      where(reset_password_token: token).first
    end
  end`

SMaxOwok:
@zdavis, refactored to include the changes you suggested.

client/src/containers/frontend/PasswordReset.js
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  componentWillUnmount() {
+    this.props.dispatch(flush(['reset-password']));
+  }
+
+  handleSubmit(event) {
+    event.preventDefault(event.target);
+    this.props.dispatch(request(passwordsAPI.update(this.state.password, this.state.passwordConfirmation, this.props.params.resetToken), 'reset-password')).promise.then((response) => {
+      this.postUpdate(response.data);
```
SMaxOwok: Is this where the reducers come into play? I didn't make/update anything in the reducers, so I'm guessing this could be further refined to not directly rely on the response.

zdavis: No... the idea I'm getting at is that this isn't really a request that can be handled within the component like this. Take this approach....

1. Make a new action in src/actions/currentuser.js with the type "START_RESET_PASSWORD". Like this:

export const resetPassword = createAction('START_RESET_PASSWORD', resetPasswordHash => resetPasswordHash);

2. In the component, dispatch the action like this:

`dispatch(resetPassword({hash, password, passwordConfirmation});`

3. If you look in the redux panel, you'll see this action get dispatched with the payload needed to reset the PW. Because this is an asynchronous action, we can go ahead and deal with it first in middleware. Look at src/store/middleware/currentUserMiddleware.js. Add something like this to the exported function:

```
    if (action.type === "START_RESET_PASSWORD") {
      startResetPassword(payload, dispatch);
    }
```

Write a function above called startResetPassword. It will will a lot like the `authenticateWithPassword` method, only it will be using a different API (`passwordsAPI`). When that API call promise is resolved, you'll get a new token back from the server (just like logging in). In the middleware, set the cookie, like we do for a regular login. Then, dispatch some more auth related methods:

```
    dispatch(actions.setCurrentUser(response));
    dispatch(actions.setAuthToken(authToken));
    dispatch(actions.loginComplete());
```

The authentication reducer will see these and login the user.

We'll need to think a little bit about what happens when the reset password fails. We probably handle it a lot like a failed login, and store the error in resetPwError in the authentication reducer state. 

Take a stab at this and see if you can make progress on it.

SMaxOwok:
Updated to return user unauthenticated user on password change and process login separately.  @zdavis, this is ready if you want to look at the input validation stuff.

SMaxOwok:
@zdavis, ready for review again.

PR #162, opened 2017-02-21
As a press user, I can create and ingest a new text [#139242257]

zdavis:


zdavis:
There's useful work in here, but it's not ready for release yet, and needs rebasing with dev.

PR #163, opened 2017-02-24
[F] Implement text annotation counts on text thumb

SMaxOwok:
[Delivers #140524019]


SMaxOwok:
Returning total count for text unless there is a user, then we return their counts.  Let me know if we'd rather always return total counts.

Also hid bookmark count since it's not implemented.

PR #166, opened 2017-02-27
Bug hunt!

SMaxOwok:
[Delivers #140512873]


SMaxOwok:
@zdavis let me know if we don't want to flush all the responses I added in https://github.com/ManifoldScholar/manifold/pull/166/commits/01044bbdb5ddd2582ce50f689785904d1ba1d5af

client/src/containers/frontend/ProjectResources.js
```diff
@@ -18,10 +18,14 @@ class ProjectResourcesContainer extends Component {
     const pageParam = params.page ? params.page : page;
     const projectRequest =
         request(projectsAPI.show(params.id), requests.feProject);
+    // This can be made more robust with other types if need be.
```
zdavis: I'd write this in a more legible way. Eg:

```
    const tag = get(this.props, 'location.query.tag');
    const initialFilter = tag ? { tag } : null;
```

client/src/containers/frontend/ProjectResources.js
```diff
@@ -73,6 +77,9 @@ class ProjectResourcesContainer extends Component {
 
   render() {
     const project = this.props.project;
+    const initialFilter = this.props.location.query.tag ? {
```
zdavis: I'd write this in a more legible way. Eg:
```
const tag = get(this.props, 'location.query.tag');
const initialFilter = tag ? { tag } : null;
```

PR #168, opened 2017-03-02
[F] Add delete icon on resource viewer singles

zdavis:
[#139241305]

### DIFF
--- a/client/src/components/reader/Resource/Viewer/Single.js
+++ b/client/src/components/reader/Resource/Viewer/Single.js
@@ -1,14 +1,21 @@
 import React, { PureComponent, PropTypes } from 'react';
 import get from 'lodash/get';
+import { Link } from 'react-router';
 import throttle from 'lodash/throttle';
 import classNames from 'classnames';
 import { Resource } from 'components/frontend';
+import { Dialog } from 'components/backend';
 import { connect } from 'react-redux';
-import { uiReaderActions } from 'actions';
+import { uiReaderActions, entityStoreActions } from 'actions';
+import { annotationsAPI, requests } from 'api';
+import HigherOrder from 'containers/global/HigherOrder';
+
+const { request, flush } = entityStoreActions;
 
 class ResourceViewerSingle extends PureComponent {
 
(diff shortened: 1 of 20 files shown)


client/src/components/reader/Resource/Viewer/Single.js
```diff
@@ -65,38 +69,62 @@ class ResourceViewerSingle extends PureComponent {
...
-        <Resource.Thumbnail
-          key={resource.id}
-          resource={resource}
-          noCrop={hasImage}
-          showTitle
-          showKind={false}
-          variant={variant}
-          additionalClasses="minimal right"
-        />
+        {this.props.authentication && this.props.authentication.authenticated ?
```
zdavis: @naomiyaki This isn't a sufficient check for whether the button should appear or not, since it would allow any authenticated user to see the button. In the future, use the RequireRole component.

client/src/components/reader/Resource/Viewer/Thumbnail.js
```diff
@@ -0,0 +1,15 @@
...
+
+
+export default class ResourceViewerThumbnail extends PureComponent {
+
+  static displayName = "ResourceViewer.Thumbnail";
+
+  render() {
+    return (
+      <div>
+        {'hi'}
```
zdavis: Should this component be removed?

client/src/components/reader/Resource/Viewer/Wrapper.js
```diff
@@ -26,6 +27,7 @@ export default class ResourceViewerWrapper extends PureComponent {
     if (this.props.annotations !== nextProps.annotations) return true;
     if (this.props.activeAnnotation !== nextProps.activeAnnotation) return true;
     if (this.props.resources !== nextProps.resources) return true;
+    if (this.props.authentication !== nextProps.authentication) return true;
```
zdavis: See my comment above about the RequireRole component... generally, just flowing more props down the tree may not be the right solution.

zdavis:
Ready to merge once tests pass.

PR #171, opened 2017-03-03
Prevent scroll propagation if the cursor is over the drawer. [#139240881]

naomiyaki:
[#139240881]
Initial POC converted to use the TOC; See TocDrawer.js
Note that the HOC (ScrollLock.js) is manipulating the direct child with
a ref, but it might be better if the child can pass a specific ref into
the HOC for cases where the scrolling element may be more deeply nested
(like on a resource drawer inside a ReactCSSTransition)

### DIFF
--- a/client/src/components/backend/Drawer/Wrapper.js
+++ b/client/src/components/backend/Drawer/Wrapper.js
@@ -2,6 +2,7 @@ import React, { PureComponent, PropTypes } from 'react';
 import { browserHistory } from 'react-router';
 import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
 import { Link } from 'react-router';
+import Utility from 'components/global/Utility';
 
 export default class DrawerWrapper extends PureComponent {
 
@@ -9,7 +10,12 @@ export default class DrawerWrapper extends PureComponent {
 
   static propTypes = {
     closeUrl: PropTypes.string,
-    closeCallback: PropTypes.func
+    closeCallback: PropTypes.func,
+    lockScroll: PropTypes.bool
+  };
+
+  static defaultProps = {
+    lockScroll: true
   };
 
   constructor(props) {
@@ -57,36 +63,40 @@ export default class DrawerWrapper extends PureComponent {
     }
   }
 
+  renderDrawer() {
+    return (
+      <div key="drawer" className="drawer-primary">
+        <div className="rel">
+          <div onClick={this.handleLeaveEvent} className="close-button-primary">
+            <i className="manicon manicon-x"></i>
+            <span className="screen-reader-text">
+              Close Drawer
+            </span>
+          </div>
+          {this.props.children}
+        </div>
+      </div>
+    );
+  }
+
+  renderDrawerWrapper() {
+    return this.props.lockScroll ? (
+      <Utility.ScrollLock>
+        {this.renderDrawer()}
+      </Utility.ScrollLock>
+    ) : this.renderDrawer();
+  }
+
   render() {
     return (
       <ReactCSSTransitionGroup
         transitionName="drawer"
         // True value required to enable transform
-        /* eslint-disable */
-        transitionAppear={true}
-        /* eslint-enable */
-        transitionEnter={false}
-        transitionAppearTimeout={1}
+        transitionEnterTimeout={5000}
         transitionLeaveTimeout={200}
       >
-        {this.state.leaving ?
-          null
-          :
-          <div key="drawer" className="drawer-primary drawer-appear">
-            <div className="rel">
-              <div onClick={this.handleLeaveEvent} className="close-button-primary">
-                <i className="manicon manicon-x"></i>
-                <span className="screen-reader-text">
-                  Close Drawer
-                </span>
-              </div>
-
-              {this.props.children}
-            </div>
-          </div>
-        }
+        { this.state.leaving ? null : this.renderDrawerWrapper() }
       </ReactCSSTransitionGroup>
     );
   }
-
 }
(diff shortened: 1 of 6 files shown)


zdavis:
No [WIP] Commits should find their way into development or master. Please rebase this down into one or more commits that follow our [F], [B], etc. conventions. If the bugfixes are also introduced in this branch, then you can probably rebase this all into a single commit.

client/src/components/backend/Drawer/Wrapper.js
```diff
@@ -2,14 +2,20 @@ import React, { PureComponent, PropTypes } from 'react';
 import { browserHistory } from 'react-router';
 import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
 import { Link } from 'react-router';
+import ScrollLock from 'components/global/Utility/ScrollLock';
```
zdavis: In general, we've been trying to import at a higher level. For example:
```
import { Utility } from 'components/global';
 
<Utility.ScrollLock>
</Utility.ScrollLock>
```

Could you do that here as well? It makes it so we don't have so many granular imports at the top of every class, which feels cleaner to me. It also makes the imports less coupled to the file system structure, which is why we have all the index.js files.

client/src/components/global/Utility/ScrollLock.js
```diff
@@ -0,0 +1,70 @@
...
+
+export default class ScrollLock extends PureComponent {
+
+  static displayName = "Utility.ScrollLock";
+
+  static propTypes = {
+    children: React.PropTypes.element.isRequired
+  };
+
+  constructor() {
```
zdavis: signature should be:

```
constructor(props) {
  super(props)
  [..]
}
```

client/src/components/global/Utility/ScrollLock.js
```diff
@@ -0,0 +1,70 @@
...
+    this.state = {
+      touchStart: null
+    };
+
+    this.edgeLock = this.edgeLock.bind(this);
+  }
+
+  edgeLock(scroller, delta) {
+    if (delta < 0 && scroller.scrollTop <= 0) {
+      event.preventDefault();
```
zdavis: This seems like it will trigger an error, since `event` does not exist at this point. Am I missing something?

client/src/components/global/Utility/ScrollLock.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    if (delta > 0 && scroller.scrollTop >= scroller.scrollHeight - window.innerHeight) {
+      event.preventDefault();
```
zdavis: This also seems wrong, since `event` hasn't been declared in this scope.

client/src/components/global/Utility/ScrollLock.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  componentDidMount() {
+    this.scroller = typeof this.props.children.type === 'string' ?
```
zdavis: In general, I prefer to use Lodash here for legibility and consistency. Throughout the code, I use

```
import isString from 'lodash/isString';
isString(this.props.children.type) 
```
instead of 
```
typeof this.props.children.type === 'string';
```
The former seems cleaner to me, and that utility function is already widely included.

client/src/components/reader/TocDrawer.js
```diff
@@ -1,5 +1,6 @@
 import React, { Component, PropTypes } from 'react';
 import { Toc } from 'components/reader';
+import ScrollLock from 'components/global/Utility/ScrollLock';
 import classNames from 'classnames';
```
zdavis: You and I have a minor style inconsistency here. I've gotten in the habit of `import classnames from 'classnames';` whereas you tend to camelCase it. We should decide on one way or another to reduce minor cognitive dissonance. I prefer the former, because it lines up with the module name, but could go either way.

client/src/components/reader/TocDrawer.js
```diff
@@ -1,5 +1,6 @@
 import React, { Component, PropTypes } from 'react';
 import { Toc } from 'components/reader';
+import ScrollLock from 'components/global/Utility/ScrollLock';
```
zdavis: Per my comment above, I'd prefer we not reference the full path, and instead import a higher level module.

client/src/components/reader/TocDrawer.js
```diff
@@ -1,5 +1,6 @@
...
 import classNames from 'classnames';
 
 export default class TocDrawer extends Component {
```
zdavis: It looks like this component might be missing a displayName static property. Not your fault, but we may as well fix in this commit.

client/src/components/reader/TocDrawer.js
```diff
@@ -18,13 +19,16 @@ export default class TocDrawer extends Component {
       'drawer-visible': this.props.visible
```
zdavis: Not part of this commit, but something we could fix. Classnames takes as many arguments as you need. So instead of this:
```
classNames({
  'toc-drawer': true,
  'drawer-hidden': !this.props.visible,
  'drawer-visible': this.props.visible
});
```
we could do:
```
classNames("toc-drawer", {
  'drawer-hidden': !this.props.visible,
  'drawer-visible': this.props.visible
});
```

zdavis:
The failing API tests are my fault. I'm looking into that and will fix in development, and we can rebase.

zdavis:
Failing tests in development have been fixed. Rebasing this onto development should correct the Travis CI build.

client/src/components/global/Utility/ScrollLock.js
naomiyaki: Good catch. I had this problem recently on another project where this works in chrome (mysteriously the event still exists, maybe in a global space?) But in Firefox it does not.

client/src/components/reader/TocDrawer.js
naomiyaki: Nice!

client/src/components/global/Utility/ScrollLock.js
zdavis: Appears fixed by adding `event` to the method signature.

PR #172, opened 2017-03-04
Reader header improvements [#140858771, #141055179]

zdavis:


client/src/components/reader/ReturnMenuBody.js
```diff
@@ -0,0 +1,59 @@
+import React, { Component, PropTypes } from 'react';
+import { Link } from 'react-router';
+import HigherOrder from 'containers/global/HigherOrder';
```
zdavis: Please change to this:
```
import { HigherOrder } from 'containers/global';
```
I'm trying to decouple imports from the file system so that if we move things around in a module, we don't have to refactor anything — the last time I moved stuff around, it was a serious pain to fix all the imports. In general, I'd prefer that imports from containers and components only reference the top level module in each directory.

client/src/components/reader/Header.js
```diff
@@ -157,7 +159,21 @@ export default class Header extends Component {
...
-        <nav className="menu-panels">
+        <nav className="menu-panels-left">
+          <UIPanel
+            id="readerReturn"
+            visibility={this.props.visibility.uiPanels}
+            bodyComponent={ReturnMenuBody}
+
+            // Props required by body component
+            projectId={this.props.text.relationships.project.id}
+            projectTitle="The Project Title"
```
zdavis: Should be accessed from `texts.relationships.project.attributes.title` rather than hard coded.

client/src/components/reader/ReturnMenuBody.js
```diff
@@ -0,0 +1,59 @@
...
+
+export default class ReturnMenuBody extends Component {
+
```
zdavis: This component is missing `static displayName`

client/src/components/reader/ReturnMenuBody.js
```diff
@@ -0,0 +1,59 @@
+import React, { Component, PropTypes } from 'react';
+import { Link } from 'react-router';
+import HigherOrder from 'containers/global/HigherOrder';
+
+export default class ReturnMenuBody extends Component {
```
zdavis: I think I mentioned this before — extend PureComponent rather than Component, unless there's a good reason not to.

client/src/components/reader/Header.js
```diff
@@ -11,9 +13,8 @@ import {
   SearchMenuBody,
   UIPanel,
   UserMenuButton,
-  UserMenuBody
+  UserMenuBody,
```
zdavis: Trailing comma should be removed.

client/src/components/reader/ReturnMenuBody.js
```diff
@@ -0,0 +1,59 @@
...
+
+  static propTypes = {
+    projectId: PropTypes.string,
+    projectTitle: PropTypes.string,
+    toggleSignInUpOverlay: PropTypes.func,
+    moreLink: PropTypes.string,
+  };
```
zdavis: Because there's no logic in the render method to handle the absence of these props, it seems to me that they should all be flagged as required for this component.

client/src/components/reader/Header.js
```diff
@@ -157,7 +159,21 @@ export default class Header extends Component {
...
+          <UIPanel
+            id="readerReturn"
+            visibility={this.props.visibility.uiPanels}
+            bodyComponent={ReturnMenuBody}
+
+            // Props required by body component
+            projectId={this.props.text.relationships.project.id}
+            projectTitle="The Project Title"
+            toggleSignInUpOverlay={this.props.commonActions.toggleSignInUpOverlay}
+            moreLink="http://manifold.umn.edu/about/"
```
zdavis: Please add a `TODO` comment so we have a note that this needs to be pulled from settings. I know there's a story, but good to track this sort of TODO in code as well.

client/src/components/reader/ReturnMenuBody.js
```diff
@@ -0,0 +1,59 @@
...
+  render() {
+    return (
+      <nav className="reader-return-menu">
+        <ul>
+          <li>
+            <Link to={`/browse/project/${this.props.projectId}`}>
+              <i className="manicon manicon-arrow-round-left"></i>
+              {'Project Home'}
+              <span>
+                  {/*
```
zdavis: Reminder to remove content when you correct this issue.

client/src/components/reader/ReturnMenuButton.js
```diff
@@ -0,0 +1,29 @@
+import React, { Component, PropTypes } from 'react';
+
+export default class AppearanceMenuButton extends Component {
+
+  static propTypes = {
+    toggleReaderMenu: PropTypes.func,
```
zdavis: Because this component would throw an exception if toggleReaderMenu were not provided, it should be marked as required (or there should be a noop function as a defaultProp, if you prefer)

client/src/components/reader/ReturnMenuButton.js
```diff
@@ -0,0 +1,29 @@
...
+    active: PropTypes.bool
```
zdavis: Why does this prop exist if it doesn't do anything? See above, where it's passed in as well.

client/src/components/reader/index.js
```diff
@@ -6,6 +6,8 @@ export AppearanceMenuButton from './AppearanceMenuButton';
 export Footer from './Footer';
 export FooterMenu from './FooterMenu';
 export Header from './Header';
+export ReturnMenuButton from './ReturnMenuButton';
```
zdavis: My preference for something like this would be this sort of structure:

```
Reader/ReturnMenu/index.js
Reader/ReturnMenu/Button.js
Reader/ReturnMenu/Body.js
```
Reader would export ReturnMenu, and in JSX you would access them with ReturnMenu.Button and ReturnMenu.Body

zdavis:
Failing test is due to rebasing on dev and is related to text metadata (which was refactored in dev). I am fixing!

PR #176, opened 2017-03-07
Users can annotate and comment on annotations [EPIC: Annotate]

zdavis:


client/src/components/backend/Drawer/Wrapper.js
```diff
@@ -11,11 +11,13 @@ export default class DrawerWrapper extends PureComponent {
   static propTypes = {
     closeUrl: PropTypes.string,
     closeCallback: PropTypes.func,
-    lockScroll: PropTypes.bool
+    lockScroll: PropTypes.bool,
+    style: PropTypes.string,
```
zdavis: This should be made required, and you should use the oneOf react proptype to specify that it must be backend or frontend.

client/src/components/frontend/Utility/ShareBar.js
```diff
@@ -53,7 +58,9 @@ class ShareBar extends Component {
 
     return (
       <nav className="share-nav-primary">
-        <span>Share</span>
+        { this.props.label ?
```
zdavis: Because this change is unrelated to the annotation work, it should be broken out into a separate commit.

client/src/components/reader/Annotation/Selection/Truncated.js
```diff
@@ -0,0 +1,63 @@
+import React, { PureComponent, PropTypes } from 'react';
+import classNames from 'classnames';
+
+export default class AnnotationSelectionTruncated extends PureComponent {
+
+  static displayName = "Annotation.Selection.Truncated";
+
+  static propTypes = {
+    selection: PropTypes.string,
```
zdavis: This should be required, as the component will throw errors without a selection.

client/src/components/frontend/Utility/ShareBar.js
zdavis: I broke it out, so this is resolved.

PR #178, opened 2017-03-08
Add react-social to social bar component [#140858491]

naomiyaki:
[Delivers #140858491]


zdavis:
FYI: I rebased this onto development to fix CI tests.

client/src/components/frontend/Resource/Detail.js
```diff
@@ -10,12 +10,14 @@ export default class ResourceDetail extends Component {
   static propTypes = {
     projectId: PropTypes.string,
     projectUrl: PropTypes.string,
+    resourceUrl: PropTypes.string,
```
zdavis: It's OK to not have this resourceUrl be required, but if it's not, then we shouldn't render the ShareBar if it's absent or blank. Or, alternately, we can make it required and always render the ShareBar.

client/src/components/frontend/ResourceCollection/Detail.js
```diff
@@ -13,6 +13,7 @@ export default class ResourceCollectionDetail extends PureComponent {
   static propTypes = {
     collection: PropTypes.object,
     project: PropTypes.object,
+    collectionUrl: PropTypes.string,
```
zdavis: Same comment as above—the collectionUrl should be required, or the share rendering should be conditional, I think.

client/src/components/frontend/Utility/ShareBar.js
```diff
@@ -1,34 +1,78 @@
...
+    });
+  }
+
   render() {
+    if (!this.state.inBrowser) return null;
+
+    const twitterWindowOptions = ["", "", "width=600,height=300"];
+    const twitterAppId = this.props.twitterAppId ? this.props.twitterAppId : null;
+    const facebookAppId = this.props.facebookAppId ? this.props.facebookAppId : null;
+    {/* Todo: The hostname needs to be inserted here */}
```
zdavis: Is this todo still necessary? It's probably fine to get it from window.location, right?

client/src/containers/frontend/CollectionDetail.js
```diff
@@ -107,7 +107,11 @@ class CollectionDetailContainer extends PureComponent {
   render() {
     const project = this.props.project;
     const collection = this.props.collection;
-    if (!project && !collection) return null;
+
+    if (!project || !collection) return null;
```
zdavis: Why the condition change? Isn't it right to return null if we're missing the project _or_ the collection?

zdavis:
This might be fine as is — just look through my comments and let me know if you think they warrant changes.

client/src/components/frontend/Utility/ShareBar.js
naomiyaki: Nope! Done did.

client/src/containers/frontend/CollectionDetail.js
zdavis: image

PR #188, opened 2017-03-16
Annotation popup has secondary page with share links [#141981779]

naomiyaki:


client/src/components/reader/Annotation/Annotatable.js
```diff
@@ -304,9 +300,9 @@ class Annotatable extends Component {
         : null}
 
         { this.props.currentUser ?
-          <Annotation.Popup
+          <Annotation.Popup.Wrapper
             currentUser={this.props.currentUser}
-            share={this.shareSelection}
+            shareUrl={`/read/${this.props.projectId}/section/${this.props.sectionId}`}
```
zdavis: I haven't tried branch yet, but just eyeballing it this looks wrong. I believe the path to a section contains the text segment, not the project segment.

client/src/components/reader/Annotation/Popup/Annotate.js
```diff
@@ -0,0 +1,57 @@
...
+  static displayName = "Annotation.Popup.Annotate";
+
+  static propTypes = {
+    secondary: PropTypes.string,
+    attachResource: PropTypes.func,
+    highlight: PropTypes.func,
+    annotate: PropTypes.func,
+    bookmark: PropTypes.func,
+    showShare: PropTypes.func,
+    pageClass: PropTypes.string,
```
zdavis: Passing in a class from the parent to the child in this case has a hint of bad code smell to me. It might be fine, but to my eyes exposing a class like this to the parent feels like a leaky abstraction. Why would the annotate and share panels ever have a different class than the one they've been given here? If there is a valid case for the parent to set their appearance, shouldn't it be by way of a directive rather than by passing the child a class, which is generally an implementation detail?

client/src/components/reader/Annotation/Popup/Wrapper.js
```diff
@@ -87,6 +107,9 @@ export default class AnnotationPopup extends Component {
     const halfPopupWidth = popupWidth / 2;
     const clientX = get(selectionClickEvent, 'clientX');
     const clientY = get(selectionClickEvent, 'clientY');
+    // const documentHeight = Math.max( document.body.scrollHeight, document.body.offsetHeight,
```
zdavis: Please remove the commented lines. We don't need to keep these around.

zdavis: Also, could you remove the comment at line 115 that I mentioned in slack? Thanks!

client/src/components/reader/Annotation/Popup/Wrapper.js
```diff
@@ -124,16 +148,47 @@ export default class AnnotationPopup extends Component {
     let direction = up;
     if (topCollisionPossible) direction = down;
     if (preferBottom && !bottomCollisionPossible) direction = down;
-    const naturalTopPosition = window.pageYOffset + rect.top - popupHeight;
+    const naturalTopPosition = window.pageYOffset + rect.top - 60;
```
zdavis: Could you leave a comment as to what the hard coded `60` represents so we know when we come back to it later?

client/src/components/reader/Annotation/Popup/Wrapper.js
```diff
@@ -124,16 +148,47 @@ export default class AnnotationPopup extends Component {
...
+      bottom = 'auto';
+    }
 
-    this.setState(Object.assign(this.state, { top, left, direction }));
+    this.setState({ top, bottom, left, direction });
   }
 
-  render() {
+  showSecondary(page) {
+    this.setState({
```
zdavis: Just a style point — let's one line things like this when we can.

PR #197, opened 2017-03-24
[F] Add functionality to about text link in reader TOC

SMaxOwok:


SMaxOwok:
@naomiyaki, I added https://www.pivotaltracker.com/story/show/142430805 re: refactoring classes.  There were a couple places I used inline styles, so feel free to change those too or let me know if I should. Thanks!

naomiyaki:
Thanks Max,

As it turns out, we're going to refactor not just the class names here, but the meta components themselves and also some of the classes that this component uses. It will probably need some style rescoping, so I'll try to handle this soon.

naomiyaki:
This work is being rolled into #579 , so I'm closing this one.

PR #202, opened 2017-03-28
[F] Display updated alert in project grid

SMaxOwok:
[Delivers #114706759]

### DIFF
--- a/api/app/models/project.rb
+++ b/api/app/models/project.rb
@@ -115,6 +115,14 @@ class Project < ApplicationRecord
     title
   end
 
+  def updated?
+    updated_at.strftime("%F") != created_at.strftime("%F")
+  end
+
+  def recently_updated?
+    updated? && updated_at >= Time.current - 1.week
+  end
+
   private
 
   def trigger_creation_event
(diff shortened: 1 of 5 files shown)


api/app/models/project.rb
```diff
@@ -115,6 +115,14 @@ def to_s
     title
   end
 
+  def updated?
+    updated_at.strftime("%F") != created_at.strftime("%F")
+  end
+
+  def recently_updated?
+    updated? && updated_at >= Time.current - 1.week
```
SMaxOwok: I chose a week, but let me know if there's a different window we want to use for recent.

api/app/models/text.rb
```diff
@@ -28,7 +28,7 @@ class Text < ApplicationRecord
   acts_as_list scope: [:project_id, :category_id]
 
   # Associations
-  belongs_to :project, optional: true
+  belongs_to :project, optional: true, touch: true
```
SMaxOwok: Updating a text will update the project.  This seemed like the only relationship that should do this, but let me know if you think there are others that should as well.

PR #203, opened 2017-03-29
[F] Adjust drive importer to handle press resource sheet style

SMaxOwok:


api/app/models/concerns/with_markdown.rb
```diff
@@ -7,11 +7,11 @@ def render_simple_markdown(input)
...
       no_links: true,
-      no_styles: true
+      no_styles: true,
+      hard_wrap: true
     }
     renderer = Redcarpet::Render::HTML.new(options)
     redcarpet = Redcarpet::Markdown.new(renderer)
-    out = redcarpet.render(input)
-    Regexp.new(%r{\A<p>(.*)<\/p>\Z}).match(out)[1]
+    redcarpet.render(input).strip
```
SMaxOwok: It looks like descriptions can be multi-paragraph, so I changed this to return all the HTML.

client/src/components/frontend/Resource/Detail.js
```diff
@@ -14,6 +14,13 @@ export default class ResourceDetail extends Component {
     resource: PropTypes.object
   };
 
+  createDescription(description) {
```
SMaxOwok: We weren't actually inserting the HTML into the descriptions, so none of the formatting was being preserved.  Let me know if there is a different way we should be doing this.

PR #204, opened 2017-03-30
[C] Refactor resources to use sub kinds for better determination

SMaxOwok:


zdavis:
@SMaxOwok I think this change will also require changes to components/frontend/ResourceList/Slide/SlideVideo.js. As I write this, Naomi is working on that. I just added some logic for non-external videos into that component last night, so you'll likely need to rebase this branch onto development. You might also want to grep the client code for external_type and make sure we're not missing anything else that needs to be refactored in response to this.

SMaxOwok:
Only saw one place where we were using externalType in the client (it _was_ in the video slideshow)  Replaced it with the subKind check and all seems well.

The only other thing was moving the external fields to the ResourcePartialSerializer, so the videos would work in the resource index slideshow.  This is actually a current issue on the dev branch.

PR #208, opened 2017-03-31
[F] Add support for OAuth authentication

scryptmouse:
[Delivers #139242561]


zdavis:
I eyeballed this and am going to work on getting it running in my env over the weekend. It looks like exceptionally good work, at first pass. 

Would you mind including some specs—use your own judgement as to what is most worthy of tests—for the API changes? While our testing is regrettably lighter on the client side, the actions and reducers tend to be really easy to test because they're stateless.

Are the errors that show up in the oAuth monitor component meant to be visible to end users? If so, should we style them up like our other errors?

scryptmouse:
Getting started on that today. The user upsertion is wrapped in `ActiveInteraction`s, so getting specs up for the different providers should go swiftly.

zdavis:
Another thought... in `components/global/SignInUp/Login.js` we need to be sure to not show the buttons for the various providers if the requisite settings haven't been put in place on the server-side. Because the client doesn't have access to the OAuth keys, one way to accomplish this might be to include some flags (google_oauth_enabled, etc) in the settings object that is exposed to the client. Those flags could be based on the existence of the requisite keys in the API env.

zdavis:
Hey Alexa. Is this ready for final review in your opinion, or are you still working on it? If it's ready, please add the `ready for review` label. Thanks!

scryptmouse:
I am still wrapping up the specs & the settings for hiding buttons for providers that aren't set up, but it should be done this evening and will do so

zdavis:
These changes are breaking the rspec run on Travis

```
TypeError: Invalid OAuth app id
```

zdavis:
Rebased onto development, resolved conflicts, and force pushed

zdavis:
@scryptmouse I'm passing this back to you to adjust the environment settings. If you could tackle that this morning, I'd appreciate it. I've put a file called `updated_dot_env` in the Manifold project folder on the NAS. If you put that in your development environment and start the server, Manifold should update the settings from the environment when it starts up. With my change, all the settings you need for OAuth will be available within the API at `Settings.instance.integrations` and `Settings.instance.secrets`.

zdavis:
Sometimes I want to punch Travis CI in his cute little face.

PR #209, opened 2017-03-31
[C] Finesse executables

scryptmouse:
* Improve how the puma script finds its Gemfile
* Use locally-installed eslint & jest rather than
  relying on globals


zdavis:
Thanks, Alexa!

PR #216, opened 2017-04-04
[B] Fix long file names overflowing dropzone

SMaxOwok:
[Delivers #142971517]


SMaxOwok:
Noticed https://www.pivotaltracker.com/story/show/143061593 while completing this.  Probably need @naomiyaki's assistance tackling that one.  #143061593 exists in dev and is not a regression from this.

zdavis:
@naomiyaki can you review this when you have a chance and accept if it looks good.

zdavis:
The vertical alignment of this element is no longer correct. 

screen shot 2017-04-06 at 5 58 04 am

The content should be vertically centered in the box.

PR #218, opened 2017-04-04
Adjustments to project event tiles

SMaxOwok:


zdavis:
Please include a spec on the Resource model asserting that the event is created after the resource.

PR #221, opened 2017-04-05
[B] Only show FE mobile nav when logged in

naomiyaki:


zdavis:
Naomi, can you give this the appropriate label so I know if it's ready or not?

PR #222, opened 2017-04-05
[B] Fix issue with resource markers not highlighting

SMaxOwok:
[Delivers #142081649]

### DIFF
--- a/client/src/components/reader/Resource/Marker.js
+++ b/client/src/components/reader/Resource/Marker.js
@@ -51,8 +51,8 @@ class ResourceMarker extends Component {
               onMouseLeave={() => { this.setActiveAnnotation(null); }}
             >
               <i
-                className="manicon manicon-cube-fill"
-              ></i>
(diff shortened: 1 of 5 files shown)


client/src/components/reader/Resource/Marker.js
```diff
@@ -51,8 +51,8 @@ class ResourceMarker extends Component {
               onMouseLeave={() => { this.setActiveAnnotation(null); }}
             >
               <i
-                className="manicon manicon-cube-fill"
-              ></i>
+                className="manicon resource-cube-fill"
+              >&nbsp;&nbsp;&nbsp;&nbsp;</i>
```
zdavis: This seems weird. Why do we have 4 non-breaking spaces here? Is this the best solution?

SMaxOwok: We had to add spaces to pad the element/icon without breaking the highlight.  When we applied padding/margin to it, the selection highlight doesn't include the padding.

PR #223, opened 2017-04-05
[F] Refactor resource tag list into component

SMaxOwok:
[Delivers #143205011]


zdavis:
I'm seeing a couple problems right off the bat...

- No hover state on the tags
- Clicking a tag should filter the list by that tag (the tag could be a link, since filters can be set via URL)
- If the resource has too many tags, they don't display at all.

SMaxOwok:
@zdavis, I removed the hover states and and links on purpose because there are the tag filters on the view.  The comps also don't show them as links.  Just want to confirm that the desired behavior is for them to link as well before I add that back.

SMaxOwok:
@zdavis rebased this on master

zdavis:
Rebased on master!

PR #224, opened 2017-04-06
[C] Use named routes for all URLs in the client.

zdavis:


zdavis:
Hey @SMaxOwok, could you give this branch a quick QA for me? Pull it and do an aggressive click-around and see if you run into errors or broken links. I think we should be in good shape, but better safe than sorry!!!!!!!

PR #231, opened 2017-04-06
[F] Allow admin to change a user's role

SMaxOwok:
[Delivers #143065151]

### DIFF
--- a/client/src/theme/Components/backend/_drawer.scss
+++ b/client/src/theme/Components/backend/_drawer.scss
@@ -120,4 +120,9 @@
   .list-pagination-primary {
     @include constrainedPaginationResponsive;
   }
+
+  // full width selects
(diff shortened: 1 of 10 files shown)


client/src/theme/Components/backend/_drawer.scss
```diff
@@ -120,4 +120,9 @@
   .list-pagination-primary {
     @include constrainedPaginationResponsive;
   }
+
+  // full width selects
+  .form-secondary .form-select {
```
SMaxOwok: Selects are full width in the backend drawers in the comp.  Let me know we were ok with them staying auto instead.

zdavis:
See my comments here: https://manifold-scholarship.slack.com/archives/C0BCDM82H/p1491520100085173

PR #259, opened 2017-04-14
[F] Minor Markdown ingestion improvements

SMaxOwok:


zdavis:
@naomiyaki I'd like to accept this today. Can you please quickly review the CSS?

zdavis:
@naomiyaki Ping — see my noted from yesterday.

client/src/theme/Components/reader/_toc-drawer.scss
```diff
@@ -87,7 +87,7 @@
       > ul > li {
```
naomiyaki: This works great! However, there's an odd appearance at certain widths (between 1024px and about 600px) where the nested list padding is less than the padding of the list above it. This may have always been the case, but we should wrap it up here. You should be able to fix this in the tocBasePadding mixin. Please have a look and let me know if you have any questions.
manifold scholarship 2017-04-19 11-52-01

SMaxOwok:
@zdavis this should be all set now.  Fixed tablet issue with @naomiyaki

PR #265, opened 2017-04-18
[F] Append collection resource filters as url params

### LINKED ISSUE #257: Filtering resource collections doesn't update URL params

SMaxOwok:
[Fixes #257]

### DIFF
--- a/client/src/components/frontend/ResourceCollection/Detail.js
+++ b/client/src/components/frontend/ResourceCollection/Detail.js
@@ -20,7 +20,8 @@ export default class ResourceCollectionDetail extends PureComponent {
     collectionPagination: PropTypes.object,
     collectionPaginationHandler: PropTypes.func,
     dispatch: PropTypes.func,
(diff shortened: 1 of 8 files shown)


zdavis:
This doesn't seem to work. Steps to reproduce:

1. Go to the collection view.
2. Set filter type to "image."
3. Copy the changed URL. Open a new tab.
4. Paste the URL in the new tab. The page loads without the filters applied.

Either the params aren't being passed to the SSR, or perhaps the container isn't applying them to the initial load. This is not an issue on the all resources filtering.

SMaxOwok:
@zdavis I rebased on master and was not able replicate the issue.

SMaxOwok:
It looks like the filter was being passed as a nested object on the first fetch, which was causing them to be ignored.  Should be fixed now.

zdavis:
This looks good. As part of our effort to improve client-side test coverage, could you please include a basic snapshot test for the two components you change in this PR? Once that's in place, this can be accepted.

PR #286, opened 2017-04-21
[F]  Editors can manage project events in the backend

### LINKED ISSUE #285: Editors can manage project events in the backend

SMaxOwok:
[Fixes #285]

### DIFF
--- a/client/src/components/backend/Event/ListItem.js
+++ b/client/src/components/backend/Event/ListItem.js
@@ -1,0 +1,42 @@
+import React, { PureComponent, PropTypes } from 'react';
+import { Event } from 'components/frontend';
+
+export default class EventListItem extends PureComponent {
+
+  static displayName = "Event.ListItem";
+
+  static propTypes = {
+    event: PropTypes.object,
+    destroyHandler: PropTypes.func
+  };
+
+  constructor(props) {
+    super(props);
+    this.triggerDestroy = this.triggerDestroy.bind(this);
+  }
+
+  triggerDestroy(event) {
+    event.preventDefault();
+    this.props.destroyHandler(this.props.entity);
+  }
+
+  render() {
+    const event = this.props.entity;
+    if (!event) return null;
+    const attr = event.attributes;
+    const content = attr.excerpt ? attr.excerpt : attr.subjectTitle;
+    return (
+      <li key={event.id} className="list-item">
+        <Event.Teaser
+          event={event}
+          showLink={false}
+        />
+        <div className="utility" data-id={'destroy'} onClick={this.triggerDestroy}>
+          <i className="manicon manicon-trashcan" />
+        </div>
+      </li>
+    );
+
+  }
+
+}
--- a/api/app/authorizers/event_authorizer.rb
+++ b/api/app/authorizers/event_authorizer.rb
@@ -1,0 +1,7 @@
+class EventAuthorizer < ApplicationAuthorizer
+
+  def self.deletable_by?(user)
+    user.admin?
+  end
+
+end
--- a/api/app/controllers/api/v1/events_controller.rb
+++ b/api/app/controllers/api/v1/events_controller.rb
@@ -1,0 +1,16 @@
+module Api
+  module V1
+    # events controller
+    class EventsController < ApplicationController
+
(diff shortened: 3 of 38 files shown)


zdavis:
There are a number of UX issues that need to be addressed before this is ready for a code review.

1. Entering anything into the search field breaks the component entirely (see below). I don't think we want to index these events in ElasticSearch — simple text matching in PG should be sufficient for searching here.

screen shot 2017-04-24 at 6 06 26 am

2. Vertical alignment in the list is weird. In fact, I don't see why we should use the same project list style here. I think it makes more sense in this case to list these out as event cards, just like how they're listed on the project page. We'll need to change the colors a bit to fit into the backend dark look, but let's use the same styles here.

screen shot 2017-04-24 at 6 06 12 am

3. This won't matter if you use the cards, but with the existing layout, the relative sizing of the icons is inconsistent. The box looks much smaller than the egg.

screen shot 2017-04-24 at 6 05 50 am

4. Tags are showing up in the list. Again, this should be addressed if we switch to the existing card layout.

screen shot 2017-04-24 at 6 06 01 am

SMaxOwok:
@zdavis this is using the existing event tiles in the list now.

zdavis:
Max,

This looks great and is getting really close. We need @naomiyaki to do a quick review on the CSS, and I noticed a few things which should be fixed:

1. We should have human readable strings in the filters dropdown rather than constants. You'll need to store the mapping between the constant and the string somewhere, and a component is likely not the best spot for it. For now it could be stored in src/config.js, which you can import into a component. We could add a new `locale` hash under `app`, and then include the mapping there.

screen shot 2017-04-25 at 7 57 14 am

2. The icon alignment is off in the screenshot below. The icon should be vertically centered with the text.

screen shot 2017-04-25 at 7 57 01 am

3. I noticed what might be a regression in the paginated events list on the frontend (http://manifold.dev/browse/project/:id/events). In my environment, when I got to the all events view for metagaming and try to use the "next" button in pagination, it always takes me back to the 2nd page for some reason. See if you can reproduce it.

naomiyaki:
I love how this looks! Aside from Zach's comments above, I just have two small points of adjustment. 

- I really like the way the trashcan looks/is positioned, because it's absolutely positioned, there's a very very small possibility that the event text can run into it. This can be avoided by putting a bit more right padding on the `event-data` element in these lists only. 
screen shot 2017-04-25 at 9 38 04 am


- The `li` tags have a min height of 170, which makes sense on the front end where the tiles have tall buttons on the bottom, but seems a bit too tall here, maybe 140-150? @zdavis ? 
manifold scholarship 2017-04-25 09-45-46

Thanks again!

zdavis:
Thanks, Naomi. Yeah, I think 140-150 could work for those.

SMaxOwok:
@zdavis, all the comments have been addressed.

- When the Searchable list tries to map the filter options, it will match to a set of labels, if provided.  I added the event types to the config file under "locale".
- Adjusted the vertical alignment of the bugle.
- Event page navigation was broken because I was missing a filter object in the request, so the page params were being passed as filters instead.
- Added a little more padding to the `event-data` in the tiles.
- Adjusted the size of the event tiles.

zdavis:
Hey Max — piece of advice. Anytime you're working with paginated result sets where it's possible to change the set (eg by deleting an event), see what happens when you remove all the results on a given page. In this case, the pagination goes all out of whack. This isn't a bug that's introduced by your code, and I'm fixing it, but this is a common sort of error, and worth keeping in your testing toolbelt.

client/src/components/backend/Event/ListItem.js
```diff
@@ -0,0 +1,42 @@
...
+    this.props.destroyHandler(this.props.entity);
+  }
+
+  render() {
+    const event = this.props.entity;
+    if (!event) return null;
+    const attr = event.attributes;
+    const content = attr.excerpt ? attr.excerpt : attr.subjectTitle;
+    return (
+      <li className="list-item">
```
zdavis: You need a unique key on this <li> (which I've added). Remember, any time you're rendering a collection like this in React, the wrapper element on each item needs a key.

PR #289, opened 2017-04-24
[B] Fix resource pagination losing filters

### LINKED ISSUE #288: Changing pages in resources list breaks filters
Filters are reset when the page is changed project resource lists.

SMaxOwok:
[Fixes #288]

### DIFF
--- a/client/src/components/frontend/Project/Resources.js
+++ b/client/src/components/frontend/Project/Resources.js
@@ -9,17 +9,13 @@ export default class ProjectResources extends Component {
     project: PropTypes.object,
     resources: PropTypes.array,
     pagination: PropTypes.object,
+    paginationClickHandler: PropTypes.func,
     filterChange: PropTypes.func.isRequired,
     initialFilterState: PropTypes.object
   };
 
   constructor() {
     super();
-    this.paginationClickHandler = this.paginationClickHandler.bind(this);
-  }
-
-  paginationClickHandler(page) {
-    return `/browse/project/${this.props.project.id}/resources/${page}`;
   }
 
   render() {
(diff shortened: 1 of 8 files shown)


client/src/components/frontend/Project/Resources.js
```diff
@@ -9,6 +9,7 @@ export default class ProjectResources extends Component {
     project: PropTypes.object,
     resources: PropTypes.array,
     pagination: PropTypes.object,
+    paginationClickHandler: PropTypes.object,
```
zdavis: Ok.... really, this should be an object?

client/src/containers/frontend/ProjectResources.js
```diff
@@ -109,6 +128,7 @@ class ProjectResourcesContainer extends Component {
             project={project}
             resources={this.props.resources}
             pagination={this.props.meta.pagination}
+            paginationClickHandler={this.pageChangeHandlerCreator}
```
zdavis: You're passing a function in a prop that's supposed to be an object, per your proptype validation above.

zdavis:
Looks like there's also a `bind` error in here somewhere.

screen shot 2017-04-25 at 8 25 11 am

zdavis:
@SMaxOwok when you rebase this, please keep 1c2763c as a stand-alone commit. I'm a little bit concerned that changing the behavior of the meta entityUtil method will cause some issues for paginated components — please take a quick pass through the backend and make sure that events, resources, projects, etc. all work with this change.

zdavis:
Max, this branch still does not work for me. When I go to the all resources view, set a type filter, then go to the second page, I see results from the first page (after a brief flash of the page 2 results). This behavior is not present in master, so it appears to be introduced by your branch.

SMaxOwok:
Per our Slack discussion:  The bug described here is not a result of this PR, but rather a result of re-fetching data on location changes.  In this case, the pagination utility appends `#pagination-target` to the URL, which causes a refetch in the component.

See #304

PR #290, opened 2017-04-24
[B] Fix resource count for collections

### LINKED ISSUE #287: Collection resource count shows all resources in project
screen shot 2017-04-24 at 12 59 43 pm

SMaxOwok:
[Fixes #287]

### DIFF
--- a/api/db/migrate/20170425220220_add_counter_cache_to_collection.rb
+++ b/api/db/migrate/20170425220220_add_counter_cache_to_collection.rb
@@ -1,0 +1,13 @@
+class AddCounterCacheToCollection < ActiveRecord::Migration[5.0]
+  def up
+    add_column :collections, :collection_resources_count, :integer, default: 0
+
+    Collection.pluck(:id).each do |id|
+      Collection.reset_counters(id, :collection_resources)
+    end
+  end
+
+  def down
+    remove_column :collections, :collection_resources_count, :integer
+  end
+end
--- a/api/app/models/collection_resource.rb
+++ b/api/app/models/collection_resource.rb
@@ -5,7 +5,7 @@ class CollectionResource < ApplicationRecord
   include Authority::Abilities
 
   # Associations
-  belongs_to :collection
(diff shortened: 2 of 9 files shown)


zdavis:
Good catch, however, I don't think the total count should be derived by counting collection.relationships.resources. That approach assumes that we'll always want to return all related resources from the API, which will probably not always be the case. For example, if a collection has, say, 500 resources in it, we're going to have to need to paginate it. We haven't solved for that yet, but we inevitably will at some point because we'll run into problems when we try to serialize large resource sets.

The total count of resources in a collection should be returned by the API as an attribute on the collection, similar to how we return the resource count on the project. If we don't have it already, we'll need a counter_cache column on the collection, and we just included that in the serialized attributes. The cache is important so we don't end up with N+1 count queries.

api/db/migrate/20170425220220_add_counter_cache_to_collection.rb
```diff
@@ -0,0 +1,12 @@
+class AddCounterCacheToCollection < ActiveRecord::Migration[5.0]
+  def change
+    add_column :collections, :collection_resources_count, :integer, default: 0
+
+    execute <<-SQL.squish
+      UPDATE collections
+      SET collection_resources_count = (SELECT count(1)
```
zdavis: I know that I tell you to avoid using models in migrations, and that's still true. However, there is a rails method for what you're doing here, which is appropriate in this case: 

http://api.rubyonrails.org/classes/ActiveRecord/CounterCache/ClassMethods.html#method-i-reset_counters

`Resets one or more counter caches to their correct value using an SQL count query. This is useful when adding new counter caches, or if the counter has been corrupted or modified directly by SQL.`

We should use the more idiomatic rails method, IMO.

SMaxOwok: Ah, I had seen that method, but was trying to stick to SQL.  Will update!

PR #295, opened 2017-04-25
[F] Add job to fetch external video thumbnails

### LINKED ISSUE #278: Youtube and Vimeo resources import thumbnails from providers

SMaxOwok:
[Closes #278]

### DIFF
--- a/api/app/models/resource.rb
+++ b/api/app/models/resource.rb
@@ -25,6 +25,7 @@ class Resource < ApplicationRecord
 
   # Associations
   belongs_to :project
+  has_one :thumbnail_fetch_attempt, dependent: :destroy
   has_many :collection_resources, dependent: :destroy
   has_many :collections, through: :collection_resources
 
@@ -37,6 +38,9 @@ class Resource < ApplicationRecord
 
   has_formatted_attributes :title, :caption, :credit, include_wrap: false
   has_formatted_attributes :description
+
+  # Paperclip direct image from URL
+  attr_reader :variant_thumbnail_remote_url
 
   # Validation
   validates :title, presence: true
@@ -74,6 +78,7 @@ class Resource < ApplicationRecord
   before_validation :update_kind
   before_update :reset_stale_fields
   before_save :update_tags
+  after_commit :queue_fetch_thumbnail, on: [:create, :update]
   after_create :resource_to_event
 
   # Create a new project event for the new resource
@@ -88,6 +93,18 @@ class Resource < ApplicationRecord
     Rails.logger.info(
       "#resource_to_event created an invalid event: #{event.errors.full_messages}"
     )
+  end
+
+  def fetch_thumbnail?
+    return unless Thumbnail::Fetcher.accepts?(self)
+    !variant_thumbnail.present? || previous_changes.key?(:external_id)
+  end
+
+  def queue_fetch_thumbnail(force = false)
+    unless force
+      return unless fetch_thumbnail?
+    end
+    FetchResourceThumbnail.perform_later(id)
   end
 
   def validate_kind_fields
@@ -167,6 +184,14 @@ class Resource < ApplicationRecord
     kind == "video" && sub_kind == "external_video"
   end
 
+  def youtube?
+    external_video? && external_type == "youtube"
+  end
+
+  def vimeo?
+    external_video? && external_type == "vimeo"
+  end
+
   def iframe?
     kind == "interactive" && sub_kind == "iframe"
   end
@@ -178,6 +203,11 @@ class Resource < ApplicationRecord
   def split_iframe_dimensions
     return nil unless iframe_dimensions
     iframe_dimensions.split("x")
+  end
+
+  def variant_thumbnail_remote_url=(url_value)
+    self.variant_thumbnail = URI.parse(url_value)
+    @variant_thumbnail_remote_url = url_value
   end
 
 end
--- a/api/app/jobs/fetch_resource_thumbnail.rb
+++ b/api/app/jobs/fetch_resource_thumbnail.rb
@@ -1,0 +1,10 @@
+# Job to fetch youtube and vimeo thumbnails in the background.
+class FetchResourceThumbnail < ApplicationJob
+  queue_as :default
+
+  def perform(resource_id)
+    resource = Resource.find(resource_id)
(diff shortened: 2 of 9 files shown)


SMaxOwok:
I added the job to the clock file with a schedule of every week.  I'm not sure how frequently we expect new resources to be added.

api/app/models/resource.rb
```diff
@@ -90,6 +94,12 @@ def resource_to_event
     )
   end
 
+  def queue_fetch_thumbnail
+    return unless external_id_changed? || new_record?
+    return if variant_thumbnail.present? && !external_id_changed?
```
zdavis: Your conditions here are too restrictive. They make it impossible, for example, to trigger a thumbnail fetch for a record that existed before this featured was added. You also don't check whether the resource is an external_video in this method. You're relying on the if condition in the after_save callback, but that's insufficient. This `queue_fetch_thumbnail` is a public method on the model and could be called from anywhere in the application — what happens if its called on something that's not a video?

The logic here needs to be made more granular. I'd replace this with something like this:
```
  def queue_fetch_thumbnail
    return unless external_video?
    if !variant_thumbnail.present? || external_id_changed?
      FetchVideoThumbnails.perform_later(id)
    end
  end
```

This is more to the point of what we really want, which is to fetch thumbnails for videos that have changed or for videos that should have a thumbnail but do not.

This change will introduce a different problem, however. If the thumbnail fetch fails for some reason, we'll be caught in an infinite loop in which the video saves without a thumbnail and a new job is spawned. The only way to work around this is to save a record of the failed attempt. I would track this in a separate model, so as not to pollute the resource with unrelated concerns. Make a model called "thumbnail_fetch_attempts" and store the resource, the attempted URL, success boolean, and attempts. We can use this to determine whether an attempt should be queued.

api/app/jobs/fetch_video_thumbnails.rb
```diff
@@ -0,0 +1,10 @@
+# Job to fetch youtube and vimeo thumbnails in the background.
+class FetchVideoThumbnails < ApplicationJob
```
zdavis: Also, this class is too narrowly named, IMO. It's likely that we'll need to fetch thumbnails for other types of resources in the future. Let's call this FetchResourceThumbnails instead of FetchVideoThumbnails.

api/app/jobs/fetch_video_thumbnails.rb
```diff
@@ -0,0 +1,10 @@
...
+  queue_as :default
+
+  def perform(video_id)
```
zdavis: resource_id, not video_id.

api/app/jobs/fetch_video_thumbnails.rb
```diff
@@ -0,0 +1,10 @@
...
+    video = Resource.find(video_id)
+    return unless video
+    VideoThumbnail::Fetcher.new.fetch(video)
```
zdavis: As with the job, the class is too narrowly focused. Thumbnail::Fetcher instead of VideoThumbnail::Fetcher

api/app/models/resource.rb
zdavis: Another thought... logic is in the wrong place here. The ThumbnailFetcher model is responsible for fetching, so the decision as to whether or not it can fetch a thumbnail should be up to it, not up to the resource. 

Look at this approach below, in which we delegate each part of the decision to (in my view) the correct model. We leave it to the fetcher to determine which kinds of resources it can handle. We leave it to the resource to decide whether it should request a fetch based on its current state. Normally, I wouldn't include the Thumbnail::Fetcher.accepts? call below, and would instead just pass the resource to the fetcher and have it do nothing if it chose not to attempt a fetch — but in this case, there's some overhead involved in creating the redis job, so I think we should avoid it if it's unnecessary. 

In the resource:
```
after_commit :queue_fetch_thumbnail

def fetch_thumbnail?
  return unless Thumbnail::Fetcher.accepts?(self)
  external_id_changed? || !variant_thumbnail.present?
end

def queue_fetch_thumbnail
  return unless :fetch_thumbnail?
  FetchVideoThumbnails.perform_later(id)
end
```

In Thumbnail::Fetcher:
```
def self.accepts?(resource)
  return resource.is_external_video?
end

def fetch(resource)
  return unless self.class.accepts?(resource)
  # Check if an attempt has been made and failed (and that nothing has changed to warrant another attempt)
  return unless already_attempted?(resource) 
  [...]
end
```

api/app/models/resource.rb
```diff
@@ -90,6 +94,12 @@ def resource_to_event
...
+    FetchVideoThumbnails.perform_later(id)
```
zdavis: Signature should be `FetchThumbnail.perform_later(id)`, IMO

api/app/models/resource.rb
```diff
@@ -180,4 +190,12 @@ def split_iframe_dimensions
     iframe_dimensions.split("x")
   end
 
+  def variant_thumbnail_remote_url=(url_value)
+    self.variant_thumbnail = URI.parse(url_value)
+    @variant_thumbnail_remote_url = url_value
+
+  rescue OpenURI::HTTPError
```
zdavis: We need to do more than log this. We need to store a record of the failure to ensure that the fetcher not get caught in an infinite loop trying to refetch things that have failed more than one or two times.

api/app/services/video_thumbnail/fetcher.rb
```diff
@@ -0,0 +1,32 @@
+module VideoThumbnail
```
zdavis: Change to `Thumbnail:Fetcher`

api/app/services/video_thumbnail/fetcher.rb
```diff
@@ -0,0 +1,32 @@
...
+  # This class is responsible for fetching tweets related to a given project
+  class Fetcher
+    def fetch(video)
+      case video.external_type
```
zdavis: You're coupling these models too closely by relying on the value of external_type here. What if we change how we store this data? Instead, let the resource have the final say as to what it is.
```
def fetch(resource)
  fetch_youtube_thumbnail if resource.is_youtube?
  fetch_vimdeo_thumbnail  if resource.is_vimeo?
end
```

PR #325, opened 2017-05-05
[B] Footer text nav has incorrect top margin

### LINKED ISSUE #308: Footer text nav has incorrect top margin
screen shot 2017-05-03 at 1 05 17 pm

naomiyaki:
[Fixes #308]

This is a view level fix that prevents Page records from appearing in the site footer unless they have `show_in_footer` set on the instance. The fix serializes show_in_footer so that it is accessible though JSX. This also displays links that don’t have a `nav_title` set using the `title` attribute as a fallback.

### DIFF
--- a/client/src/components/frontend/Layout/Footer.js
+++ b/client/src/components/frontend/Layout/Footer.js
@@ -42,12 +42,14 @@ export default class LayoutFooter extends Component {
   buildContentPages() {
     let pages = [];
     if (this.props.pages) {
-      pages = this.props.pages.map((page) => {
-        return (
-          <Link to={`/browse/page/${page.attributes.slug}`}>
-            {page.attributes.navTitle}
-          </Link>
-        );
+      this.props.pages.forEach((page) => {
+        if (page.showInFooter) {
+          pages.push((
+            <Link to={`/browse/page/${page.attributes.slug}`}>
+              {page.attributes.navTitle ? page.attributes.navTitle : page.attributes.title}
+            </Link>
+          ));
+        }
       });
     }
     return pages;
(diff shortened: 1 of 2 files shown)


client/src/components/frontend/Layout/Footer.js
```diff
@@ -42,12 +42,14 @@ export default class LayoutFooter extends Component {
...
     let pages = [];
     if (this.props.pages) {
-      pages = this.props.pages.map((page) => {
-        return (
-          <Link to={`/browse/page/${page.attributes.slug}`}>
-            {page.attributes.navTitle}
-          </Link>
-        );
+      this.props.pages.forEach((page) => {
+        if (page.showInFooter) {
```
zdavis: This seems wrong to me. Isn't showInFooter an attribute on the page record, and shouldn't this therefore be "page.attributes.showInFooter"?

Did you test this with an actual page record with show_in_footer disabled/enabled? Did it work?

One way you could improve this PR would be to include a test for this element. We have client-side entity factories for our tests now, and you could easily make a page with showInFooter true and one with it as false to confirm we get the direct behavior. Happy to help with this is you'd like me to!

zdavis:
@naomiyaki, could you please redo this PR against the `v0.1.x` branch (see https://manifold-scholarship.slack.com/archives/C0BCDM82H/p1494267357450620)

naomiyaki:
You got it @zdavis ! You are right about `attributes` I missed that object, and was able to test this as working now. I didn't write a test yet because I wanted to get the fix out sooner than later, but I would very much like to go over how to do that when I'm not rushing on another project.
 
manifold scholarship 2017-05-11 09-22-28

zdavis:
I'd be happy to write the test and add it to this PR. If you could make a little time to review it, then I can answer any questions that come up or walk you through it when you have more time.

naomiyaki:
Closing to open PR on v.0.1.x instead.

PR #330, opened 2017-05-08
[F] Implement reader passage citation generator

zdavis:


SMaxOwok:
@zdavis, this should be ok to take a look at the specific formats for each citation style.

PR #334, opened 2017-05-09
[F] Add project avatar selector to backend project view

SMaxOwok:


zdavis:
Max, where are we at with this one? Is it waiting on me for something? Could you rebase it onto master when you have a chance.

SMaxOwok:
This is just waiting on final styling/better responsiveness from @naomiyaki or @blnkt.

PR #340, opened 2017-05-11
[B] Footer text nav has incorrect top margin

naomiyaki:
Only show Page records in footer if they have the attribute `show_in_footer`

### DIFF
--- a/api/app/serializers/page_partial_serializer.rb
+++ b/api/app/serializers/page_partial_serializer.rb
@@ -2,5 +2,5 @@
 class PagePartialSerializer < ActiveModel::Serializer
   meta(partial: true)
 
-  attributes :id, :slug, :title, :nav_title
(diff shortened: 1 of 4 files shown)


zdavis:
Needs tests; ZD to provide.

zdavis:
Hey @SMaxOwok could you please add a test for this that asserts that only pages with a true show_in_footer attribute appear in the footer? I thought I'd get to it last week, but ran out of time.

zdavis:
Hey Max, sorry I didn't catch this earlier — we need to make one more change here. There's a hard coded ULR in the component, which needs to be refactored to use the new link helpers. Could you update this PR to use the link helpers. Note that in doing so, we'll need the PR to be against master, not v0.1.x, which I think is fine.

SMaxOwok:
Rebased on master and updated to use link helper.

Issue #302, opened 2017-04-28
Add keyboard inputs to all modal dialogs


zdavis:
When there's a modal dialog UI with an OK and cancel button, users should be able to interact via keyboard input. Return should trigger the OK callback, and escape should trigger the cancel callback.

zdavis:
I think this is a simple change in Dialog.Confirm...

PR #348, opened 2017-05-16
[F] Enable keyboard controls for modals

SMaxOwok:
[Completes #302]

### DIFF
--- a/client/src/components/backend/Dialog/Confirm.js
+++ b/client/src/components/backend/Dialog/Confirm.js
@@ -14,22 +14,44 @@ export default class DialogConfirm extends PureComponent {
 
   static defaultProps = {
     heading: "Are you sure?"
-  }
+  };
+
+  static contextTypes = {
+    pauseKeyboardEvents: PropTypes.func,
+    unpauseKeyboardEvents: PropTypes.func
+  };
 
   constructor(props) {
     super(props);
     this.handleResolveClick = this.handleResolveClick.bind(this);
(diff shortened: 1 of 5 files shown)


client/src/components/backend/Dialog/Confirm.js
```diff
@@ -72,3 +91,8 @@ export default class DialogConfirm extends PureComponent {
   }
 
 }
+
+DialogConfirm.contextTypes = {
```
zdavis: Can't these be added as static properties on the class? That would be more in keeping with what we're doing elsewhere with propTypes, etc.

client/src/components/backend/Dialog/ResetPassword.js
```diff
@@ -213,6 +211,11 @@ class ResetPasswordWrapper extends PureComponent {
 
 }
 
+ResetPasswordWrapper.contextTypes = {
```
zdavis: Can't these be added as static properties on the class? That would be more in keeping with what we're doing elsewhere with propTypes, etc.

client/src/components/backend/Drawer/Wrapper.js
```diff
@@ -169,4 +188,9 @@ class DrawerWrapper extends PureComponent {
   }
 }
 
+DrawerWrapper.childContextTypes = {
```
zdavis: Can't these be added as static properties on the class? That would be more in keeping with what we're doing elsewhere with propTypes, etc.

PR #352, opened 2017-05-18
[F] Add backend collections management

SMaxOwok:


zdavis:
Hey Max,

I looked at this, and I don't think it's really working to have the collections list and resources list on the same page. Can we break out collections into a separate tab on the project?

PR #376, opened 2017-06-02
[F] Destroy text_created event when text is removed

SMaxOwok:
[Completes #374]


zdavis:
@SMaxOwok — I added a scope to the has_one. It's possible that in the future we'll have different events attached to texts, and this scope will ensure that the correct creation event is used for this has_one association.

PR #382, opened 2017-06-02
[B] Fix issue with resource marker not updating on delete

### LINKED ISSUE #373: Removing a resource annotation doesn't remove it from margin until page reload.

SMaxOwok:
[Fixes #373]


zdavis:
@SMaxOwok I added an extended commit message here that explains what we discovered on Friday. I want to make sure that if we come back to this, we are able to remember why we made the change that we did.

PR #385, opened 2017-06-02
[B] Fix project thumbnail usage and style

### LINKED ISSUE #378: Project thumbnail is stretched on frontend
If you upload a landscape image as a project thumbnail, it seems to be cropped in the backend (good), but stretched on the frontend (bad).

screen shot 2017-06-02 at 11 30 55 am

SMaxOwok:
This PR address bug #378, but also includes a usage change.  We previously were using the project's avatar image on the home grid, while it seems we should be looking for Cover -> Avatar -> Placeholder.

[Fixes #378]

### DIFF
--- a/api/app/models/concerns/attachments.rb
+++ b/api/app/models/concerns/attachments.rb
@@ -59,7 +59,7 @@ module Attachments
 
   # rubocop:disable Metrics/BlockLength
   class_methods do
-    # Sets up paperclip atttachment configuration for `field`. The `type` argument
+    # Sets up paperclip attachment configuration for `field`. The `type` argument
     # references attachment validation in Manifold config attachments.validations. This
     # bit of metaprogramming provides a number of methods to the model for each attachment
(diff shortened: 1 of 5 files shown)


zdavis:
@SMaxOwok — actually, we want to use avatar -> placeholder on the project thumbnail grids. I need to rename the cover field, but that is only rendered on the project page. Can you adjust accordingly?

SMaxOwok:
@zdavis I updated this to only use the avatars or placeholders.

We needed a little bit of calculation to determine which attachment style to use (square version if square or landscape), so it seemed to make sense to store the dimensions in the db rather than calculate each time the page renders.  I worked with @scryptmouse, originally using the PaperclipMeta gem, but simplified it to just add a method to the attachment concern.

This will add the original file dimensions as an attribute on the project.  Any attachment can have this applied with an additional column `#{field}_meta`.

If you want to approach this in a different way, let me know.

Issue #281, opened 2017-04-21
Backend collection management


zdavis:

zdavis:
From PT:

```
This view is going to change a bit, but Lael and I are still working through the design. In the meantime, I want admins to have the basic ability to create collections and put resources in them.

Add a new "Collections" tab to the backend project edit view. Clicking it takes you to a list of collections on that project, which can follow the same format as the resource list. Build create and edit forms for the collection.

Add a has_many type field to the resource view that allows a user to place it in a project collection.
 ```

zdavis:
To start, let's add a searchable list of resource collections to the resources tab on the backend projectDetail view. The collection list can live above the resource list. Users should be able to add a new collection or click through to a collection. The collection edit view should let users edit the collection properties and thumbnails. 

We need to think about an effective way to allow users to add and remove resources in the collection. I may ask Lael to do some design here, but in the meantime I think we should have a stop-gap approach. Perhaps we could have a searchable list of resources with "in collection" and "not in collection" filters, in addition to the regular filters. Each row in the list could include a switch to toggle membership in the collection. Explore this a bit and see what you can come up with.

Note that this is for v0.3.0. Until v0.2.0 is released, this PR should be for a v0.3.x branch, which we can create. Once v0.2.0 is out, the PR can be against master.

PR #387, opened 2017-06-03
[F] Implement backend collection management

zdavis:
[Completes #281]


zdavis:
The lint errors above are inaccurate. We prefer leading dot position, yet Hound is for some reason not seeing that in our rubocop config file. These linting errors can be ignored.

zdavis:
Nice job on this, @SMaxOwok

PR #392, opened 2017-06-09
[F] Add draft mode to projects

SMaxOwok:


zdavis:
@SMaxOwok please rebase onto master when you have a chance.

SMaxOwok:
@zdavis rebased on master

api/app/controllers/api/v1/projects_controller.rb
```diff
@@ -27,6 +27,7 @@ def index
       def show
         @scope_for_projects = Project.friendly
         @project = load_project
+        authorize_action_for @project if @project.draft
```
zdavis: Can you help me understand what's happening here? This doesn't make sense to me. It seems to me that we should always authorize the project show method and handle any conditional logic in the project authorizer. If the project is set to draft and the user is not an admin, then the user is NOT authorized to show the project.

api/app/models/project.rb
```diff
@@ -82,6 +82,7 @@ class Project < ApplicationRecord
 
   # Callbacks
   after_commit :trigger_creation_event, on: [:create]
+  before_update :set_last_updated_at
```
zdavis: I think I understand what you're doing here. You're trying to avoid updating the update timestamp for a project when the draft mode changes. This approach feels wrong to me. None of this update timestamp logic feels like a proper concern for the Project model.

Rails 5 lets you not update a timestamp when you save the model by passing a touch: false option to the save method. For example: `project.save(touch: false)`

I think you should drop all this timestamp logic and just handle this in the project controller's update method. Currently, you have this:

`::Updaters::Project.new(project_params).update(@project)`

Why not change the `update` method in `Updaters` so it accepts a second options argument. In the controller, if the draft mode is the only thing being changed, you could pass `touch: false` to the update method, which could in turn pass the touch false option to the save method. This seems simpler and more reusable to me.

api/app/serializers/project_partial_serializer.rb
```diff
@@ -7,7 +7,8 @@ class ProjectPartialSerializer < ActiveModel::Serializer
              :purchase_price_currency, :purchase_price, :purchase_version_label,
              :twitter_id, :instagram_id, :facebook_id, :hero_styles, :cover_styles,
              :avatar_styles, :recently_updated, :updated, :description_formatted, :slug,
-             :resource_kinds, :resource_tags, :avatar_color, :avatar_meta
+             :resource_kinds, :resource_tags, :avatar_color, :avatar_meta, :draft,
+             :last_updated_at
```
zdavis: Per above, I think this should be removed.

api/spec/models/project_spec.rb
```diff
@@ -227,5 +237,31 @@
     end
   end
 
+  context "does not change the updated_at attribute" do
```
zdavis: This test should be in the controller, not the model.

client/src/api/resources/favoriteProjects.js
```diff
@@ -1,5 +1,7 @@
 export default {
-  index(filter = {}, page = {}) {
+  index(filterParams = {}, page = {}) {
+    const filter = filterParams;
+    filter.draft = false;
```
zdavis: This seems weird, too... seems like if a project is draft, it shouldn't show up. The filter state should depend on whether the user is an admin or not.

client/src/api/resources/projects.js
```diff
@@ -1,5 +1,7 @@
 export default {
-  index(filter = {}, page = {}) {
+  index(filterParams = {}, page = {}, includeDrafts = false) {
```
zdavis: This shouldn't be a client application concern at all, IMO. The API should decide what to return to the user — if the user is an admin, drafts are included in results. If they are not an admin, they can't see any drafts. If we handle it in the client rather than the server, we're not really hiding anything from the user.

client/src/api/resources/projects.js
```diff
@@ -87,6 +89,7 @@ export default {
 
   featured(limit = 6, filterParams = {}) {
     const filter = filterParams;
+    filter.draft = false;
```
zdavis: This is incorrect as well. If a project is a draft project, it's never visible to a non-admin user, regardless of whether or not it's featured. See why we need to handle this check server side?

zdavis:
This branch breaks backend project search. 
screen shot 2017-07-25 at 10 52 34 am

api/db/migrate/20170607190058_add_draft_to_projects.rb
```diff
@@ -0,0 +1,18 @@
...
+    say_with_time 'Setting draft to FALSE for existing projects' do
+      Project.find_each do |project|
+        project.draft = false
+        project.save
+      end
+    end
+  end
+
+  def down
+    add_column :projects, :draft, :boolean
```
zdavis: Shouldn't this be remove_column?

zdavis:
I'm going to accept this. However, we need a way to indicate to the admin user which projects are in draft mode.

naomiyaki:
Note that the draft label will be added to project detail heroes on the front-end as part of another issue to change the hero layout.

zdavis:
@SMaxOwok or @naomiyaki We just need to remove the existing draft label:

image

If someone could do that, I'll accept it.

naomiyaki:
@zdavis Max removed the label! Please give a second look and merge.

api/app/controllers/api/v1/projects_controller.rb
```diff
@@ -27,6 +29,7 @@ def index
       def show
         @scope_for_projects = Project.friendly
         @project = load_project
+        authorize_action_for @project if @project.draft?
```
zdavis: Remove the condition — as above, the `if @project.draft?` is actually authorization logic that doesn't belong in this controller. Imagine if we has another controller that returned projects. We'd have to repeat this logic for every controller that touched projects. This PR is a good example of why DRY is about more than typing less and saving time — DRY code is more flexible and easier to refactor.

api/app/controllers/api/v1/projects_controller.rb
```diff
@@ -37,7 +40,8 @@ def create
 
       def update
         @project = load_and_authorize_project
-        ::Updaters::Project.new(project_params).update(@project)
+        options = @project.draft? || project_params.dig(:draft) ? { touch: false } : {}
```
zdavis: Why is this line here? Again, this seems like the wrong place for this. Why not put it in the updater itself, if it's necessary?

zdavis: The logic here seems weird. "Any time a request to update a project comes in with a draft param, we don't change the timestamp on the project". What is this trying to accomplish? And why? And if this logic is important, why is it only in this controller?

api/app/services/demonstration/data_loader.rb
```diff
@@ -29,6 +29,11 @@ def import_projects
         next if File.file?(File.join(child, ".skip"))
         Importer::Project.new(child, cli_user, @logger).import(true)
       end
+      # Make all demo projects live
```
zdavis: I'd move this to Importer::Project and just have that service create non-draft projects by default. This block is problematic, too, because it's not just making all demo projects live—its making _all_ projects live. If we remove the part of the demo data loader that clears data before import, we'd have a problem.

api/app/services/importer/project.rb
```diff
@@ -47,7 +47,7 @@ def upsert_project(include_texts)
       project.save if project.new_record?
       project.update(@project_json[:attributes])
       assign_project_attachments(project)
-      excludes = %w(cover_ avatar_ hero_)
+      excludes = %w(cover_ avatar_ hero_ draft)
```
zdavis: Why?

client/src/api/resources/favoriteProjects.js
```diff
@@ -1,5 +1,6 @@
 export default {
-  index(filter = {}, page = {}) {
+  index(filterParams = {}, page = {}) {
```
zdavis: What is the point of this change?

client/src/containers/backend/Dashboard.js
```diff
@@ -19,7 +19,7 @@ const perPage = 5;
 export class DashboardContainer extends PureComponent {
   static fetchData = (getState, dispatch) => {
     const projectsRequest = request(
-      projectsAPI.index({}, { size: perPage }),
+      projectsAPI.index({}, { size: perPage }, true),
```
zdavis: Why this change? As far as I can tell, the method signature for the index method is: `index(filter = {}, page = {})`, yet you're passing a third argument to it.

client/src/containers/backend/Dashboard.js
```diff
@@ -54,7 +54,7 @@ export class DashboardContainer extends PureComponent {
   updateResults(event = null, page = 1) {
     const pagination = { number: page, size: perPage };
     const action = request(
-      projectsAPI.index(this.state.filter, pagination),
+      projectsAPI.index(this.state.filter, pagination, true),
```
zdavis: See previous comment.

client/src/containers/frontend/Featured.js
```diff
@@ -43,6 +43,13 @@ export class FeaturedContainer extends Component {
     subjects: PropTypes.array,
     dispatch: PropTypes.func
   };
+
```
zdavis: Why is this revision part of this branch? How is it related to draft projects?

client/src/containers/frontend/Home.js
```diff
@@ -99,6 +104,22 @@ export class HomeContainer extends Component {
     };
   }
 
+  fetchFilteredProjects = () => {
+    const filteredRequest = request(
+      projectsAPI.index(this.props.projectFilters),
```
zdavis: Shouldn't this be respecting pagination?

client/src/containers/frontend/Featured.js
SMaxOwok: This is necessary if you're on the /featured page and log in.  You won't see the draft projects unless we refetch.

PR #413, opened 2017-06-28
[F] Admins can create new user accounts

SMaxOwok:
Adds a new user creation interface for admins with randomly generated password.
Update user sign in overlay to include email, name and password.
Includes new eye with a slash through icon.


SMaxOwok:
I don't have any tests in place for the frontend components at the moment.  I had one for each new component, but because the password is generated in the client, the snapshot never matches.  I can add them back if there are any ideas on how to approach that.

zdavis:
Just FYI, I'm still working on this PR.

PR #448, opened 2017-07-19
[B] Fix postcss webpack config

### LINKED ISSUE #439: Frontend dropdown styles aren't right in Firefox.
screen shot 2017-07-14 at 2 32 04 pm

SMaxOwok:
[Fixes #439]

### DIFF
--- a/client/webpack/base.config.js
+++ b/client/webpack/base.config.js
@@ -50,7 +50,13 @@ module.exports = (options = {}) => {
     use: [
       { loader: "style-loader" },
       { loader: "css-loader", options: "importLoaders=2" },
-      { loader: "postcss-loader", options: "syntax=postcss-scss" },
+      {
+        loader: "postcss-loader",
+        options: {
+          syntax: "postcss-scss",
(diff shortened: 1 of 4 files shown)


client/webpack/base.config.js
```diff
@@ -50,7 +50,13 @@ module.exports = (options = {}) => {
     use: [
       { loader: "style-loader" },
       { loader: "css-loader", options: "importLoaders=2" },
-      { loader: "postcss-loader", options: "syntax=postcss-scss" },
+      {
+        loader: "postcss-loader",
+        options: {
+          syntax: "postcss-scss",
+          plugins: () => [require("autoprefixer")]
```
zdavis: You're requiring a module that isn't explicitly included in our package.json file. It works, because autoprefixer happens to be a dependency of css-loader. But, while it works, it's not correct. Add auto-prefixer to our package.json file to address this.

zdavis:
@SMaxOwok please rebase into a single commit.

PR #449, opened 2017-07-19
[B] Fix text clipping through container

### LINKED ISSUE #401: Long titles break event layout
screen shot 2017-06-21 at 6 08 08 am

SMaxOwok:
Fixes wrapping in several places:
- Project grid titles
- Event card titles
- Project text titles

[Fixes #401]

### DIFF
--- a/client/src/theme/Components/browse/event/_events.scss
+++ b/client/src/theme/Components/browse/event/_events.scss
@@ -126,6 +126,10 @@
       padding-left: 36px;
     }
 
+    > div {
+      width: 100%;
+    }
+
     .manicon {
       position: absolute;
       top: 0;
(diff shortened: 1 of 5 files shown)


client/src/theme/Components/browse/event/_events.scss
```diff
@@ -126,6 +126,10 @@
       padding-left: 36px;
     }
 
+    > div {
```
naomiyaki: Good call. In general, we don't want to use selectors as unspecific as `> div` unless it's for a `ul` inside a `nav` or something (because that's a very common case) but it works well here, since it's a simple flex item wrapper.

zdavis:
@naomiyaki happy with this?

Issue #404, opened 2017-06-21
Annotation permalinks and scroll to annotation


zdavis:

zdavis:
While the issue subject only refers to annotations, there's a large issue that we need to solve for here, which is the ability to trigger a scroll to a specific spot in a text_section based on the URL. I can think of at least three or four use cases for this:

1. As a user, I can enter a URL with a specific annotation ID, and that annotation will be scrolled into view
2. As a user, I can click on a bookmark I've saved for a specific location in the text and be scrolled to that location
3. As a reader, I can click a link that takes me to a specific text node (every text node in the text section body has a unique ID)

Your solution should be abstract enough to eventually help us solve all of these three problems. I haven't given a lot of thought to how I'd approach this. But, off the top of my head, I think I might start with by wrapping the reader in a higher order component whose job was to scroll to these various locations. That HOC would be wrapped by withRouter, so it could have access to the URL. It might pass a callback to the body component that it wraps, which could be called on componentDidMount — the HOC needs to know when the body has been rendered and mounted into the DOM, or it won't be able to figure out where to scroll to. When the hash segment of the URL changes (URL would be a prop for the HOC — see react-router for more on this), the HOC could trigger a scroll to the correct spot.

Keep in mind that many of the EPUBs we've ingested also have anchor links to link to specific spots. You can see this with footnotes in the texts. I can't remember what, if anything we have in place to support this now. If I remember correctly, when I click a footnote link, we get a scroll to the location, rather than the jump I'd expect to see if it was just being handled by the browser. If that's the case, then there's already some code in the reader handling this.

Start exploring this issue a bit, think about how you could approach it, and consult the react-router docs on scroll restoration so you know a bit about how we're already handling scrolling (https://reacttraining.com/react-router/web/guides/scroll-restoration). Once you have an idea of how to solve this, write a brief overview of your approach (or a proof of concept) and I'll look at it before you get in it too deep.

The devil will likely be in the details on this issue. It wont be hard to come up with a simple solution, but coming up with one that supports the different use cases, works well with the browser forward/back button, etc. may be more difficult.

SMaxOwok:
The footnote scrolling is happening in the Text component as `maybeScrollToAnchor`, which compares location hashes.

A quick solution is just expanding the selectors the function looks at, such as:
```
  maybeScrollToAnchor(previousHash, currentHash) {
    if (!currentHash) return;
    if (previousHash === currentHash) return;
[...]

PR #451, opened 2017-07-21
[F] Scroll to various anchors in reader text

SMaxOwok:
[Completes #404]


zdavis:
Just eyeballing this, it looks to me like you'd run into problems if a text were annotated with the same resource twice. Why is a resource annotation handled differently from any annotation?

SMaxOwok:
@zdavis, I updated this to only use the annotation prefix.  Each element still had a unique ID (in the case of multiple markers with the same resource), but I wasn't picking up on that ID being the annotation record's ID.  The functionality should be the same now, but whenever we put this to use the links can be generated based solely on Annotation ID.

Issue #450, opened 2017-07-20
Users can see all of their annotations and highlights on a given text


zdavis:
I want to begin putting together a proof of concept view that lists all of a user's annotations for a given text. I don't yet have design for this, so for now let's put something together that is simple and clean, and we can go back and fix it up. To start this, make a new route and corresponding container... let's put it at /reader/text/:id/my/annotations. The view can render in the reader, and it'll have a list of all annotated passages (maybe using the green style we see in the drawer) and whatever the user wrote about it below.

We'll probably want additional takes on this in the future. For example, I might want to see all my annotations for a given section, for a given text, and for a given project — so think about how to abstract the props to make this component more reusable.

SMaxOwok:
@zdavis, I've added a PR for this.

The annotations can be seen at `/read/:text_id/user/:user_id/annotations`. Right now, any user can view another user's annotation list, but only the creator can edit or view replies.  I wasn't totally sure if the list itself was something we wanted only the creator to be able to see or not.  It can be easily changed to that, if that's the case.

I kept the component props to a list of annotations and dispatch.  Any container that uses the list component can specify an annotation filter (for current implementation it defaults to ones for that text).  Additionally if there is an overall user profile page with all annotations, adding a filter method will be easy.

Let me know how you think it looks.

zdavis:
Wrote a brief for Lael on this and we discussed it at length today. Design is in progress.

zdavis:
Reworking this story title to capture the actual requirements, and the fact that this likely will be in the v0.3.0 release, and not just a POC.

SMaxOwok:
Closed via #497

PR #455, opened 2017-07-24
[F] List all highlights and annotations for a given text

SMaxOwok:
[Completes #450]

### DIFF
--- a/api/app/controllers/api/v1/me/relationships/annotations_controller.rb
+++ b/api/app/controllers/api/v1/me/relationships/annotations_controller.rb
@@ -1,0 +1,37 @@
+module Api
+  module V1
+    module Me
+      module Relationships
+        # Annotations controller
+        class AnnotationsController < ApplicationController
+          resourceful! Annotation, authorize_options: { except: [:index] } do
+            Annotation.filter(
+              with_pagination!(annotation_filter_params),
+              scope: @current_user.annotations
+            )
+          end
+
+          def index
+            @annotations = load_annotations
+            location = api_v1_me_relationships_annotations_url(@current_user.id)
+            render_multiple_resources(
+              @annotations,
+              each_serializer: AnnotationSerializer,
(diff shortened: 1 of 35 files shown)


zdavis:
Rebased onto master

zdavis:
Good start... some changes:

1. Expose this at /read/:text_id/my/annotations in the client.
2. The API route should be at /api/v1/me/relationships/annotations?filter[text]=:text_uuid

The client doesn't have to pass the users UUID to the API, since all the routes under api/v1/me are understood to operate on the current_user

It would be nice if we could keep this more in line with the reader UI. You have it in a very wide container, and none of the reading controls work. How does it look if the width is constrained like it is when reading a text_section?

I also think we need some kind of header here. And the green is too much — need a paler green or a gray. We can get design from Lael, but something to consider.

SMaxOwok:
@zdavis I changed up the routes and added the reader control functionality (moved into helper).  I added some basic dark scheme styles as well, I'm not sure how full-on you want to go with them.  The annotation drawer that this is styled after has no dark scheme in the reader, right now.  The components respond to all the controls with the exception of font-size on the annotated passage.  

annotations

Additionally, I added a button to the annotations in the reader menu.  I'm not sure if it's best there or somewhere in the actual reader as a button, but would be easily moved elsewhere.

screen shot 2017-07-26 at 2 10 31 pm

So far everything that is in place should be easy to move around or change, so let me know what you think as your thoughts on this develop.

zdavis:
Resolving this conflict requires a little discussion. @SMaxOwok and @naomiyaki let's touch base tomorrow morning. Max has moved a bunch of styles out of the Text component, but I don't think that makes sense now that we're moving this into a modal. Meanwhile, I've made changes recently around how classes are set in relation to the reader font size that conflicts with the direction Max is going in.

zdavis:
I'm closing this PR in favor of a different branch that reuses existing components.

PR #457, opened 2017-07-24
[B] Fix issue with reader return text not wrapping

SMaxOwok:


SMaxOwok:
screen shot 2017-07-21 at 10 04 50 am
screen shot 2017-07-21 at 10 04 53 am

client/src/theme/Components/reader/_return-menu.scss
```diff
@@ -58,6 +58,8 @@
       margin-bottom: -4px;
       font-size: 0.692em;
       color: $neutral50;
+      overflow-wrap: break-word;
```
naomiyaki: Looks good! Max, can you put a comment here saying that we're using both attributes to support older browsers?
You can site this too.

SMaxOwok:
This will be fixed with #401

PR #459, opened 2017-07-24
[B] Generate hashtag on project import

SMaxOwok:
Fixes an issue where projects will be overwritten if there is no hashtag in the `project.json`.


zdavis:
This is actually already fixed in a branch I'm working on... I ran into the same problem with importing the standard ebooks

zdavis:
Closing because it's been fixed in 21ecd1099f61675dca97f1a36a96d5e52ffaaf6a

PR #462, opened 2017-07-25
[B] Fix project hero buttons wrapping

### LINKED ISSUE #453: When typekit is disabled, read button wraps
screen shot 2017-07-24 at 9 46 58 am

SMaxOwok:
[Fixes #453]

### DIFF
--- a/client/src/theme/Components/browse/project/_hero.scss
+++ b/client/src/theme/Components/browse/project/_hero.scss
@@ -246,7 +246,7 @@
     }
 
     .project-entry {
-      max-width: 210px;
+      display: inline-block;
       margin-top: 34px;
 
       &.top {
@@ -262,7 +262,7 @@
         display: none;
 
         @include respond($break75) {
-          display: block;
+          display: inline-block;
         }
       }
 
@@ -272,7 +272,7 @@
     }
 
     .button-secondary {
-      width: 100%;
+      display: block;
 
       + .button-secondary {
         margin-top: 14px;


zdavis:
I'm rejecting this because it feels like an unsustainable way to fix this problem. In all programming, you need to think about separation of concerns (https://en.wikipedia.org/wiki/Separation_of_concerns). Your solution here is coupling the project hero component to the presence of a typekit setting so that it can add a very specific class. How would this scale across other components? Do we have to couple the components to our typekit settings every time there's a display discrepency? That feels really unsustainable.

This should be a simple style change. There are classes on the <HTML> element that are applied if typekit is loaded. I would think that the default styling of this button (or the container) would work with our fallback fonts, and if typekit is loaded then we could use the typekit classes to set more appropriate styles. I don't really see why Javascript should be involved at all.

@naomiyaki might have thoughts here as well.

zdavis:
@naomiyaki Can you look at this? Does it make sense to have this declaration in a separate typekit partial? To my mind, it would be better to have the override present in the _hero partial where it can be seen in context. Do you agree?

naomiyaki:
It's true that this is an issue when typekit isn't loaded, but it's also not to blame.
Here's what we're trying to achieve:
Two buttons, each with one line of text
Both of them need to be the same width, and not wrap.

Naomi's old (clearly not quite right) solution:
Wrap two 100% width buttons in a 210px wrapper. This is a problem because the buttons can still wrap.
We could prevent wrapping with white-space no-wrap, but this could get bad if the buttons are indeed really long.

Instead, the wrapper around the buttons should have no explicit width, and be an inline block.
This way, it will always be the width of its largest child automatically.

That way, both buttons can be as large as they need to be (And they can even have a min-width: 210px if needed) and always stay the same size. Here's what I mean!
https://codepen.io/naomirubin/pen/XamKLz

This should all be done in the same hero partial since it's a unique layout to these buttons. Please let me know if you need further help!

Issue #461, opened 2017-07-25
Add tests for the frontend user create components


zdavis:
You can do this until pull request #413 is in master. When you get to it, you should mock the module that generates the password so that it returns the same password every time, which will allow your snapshots to work.

PR #463, opened 2017-07-25
[C] Add tests for new user components

SMaxOwok:
[Completes #461]

### DIFF
--- a/client/__mocks__/password-generator.js
+++ b/client/__mocks__/password-generator.js
@@ -1,0 +1,3 @@
+export default function generatePassword() {
+  return "testtest123";
+}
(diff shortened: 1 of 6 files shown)


zdavis:
I'd prefer not to add a top-level directory for this. Can't we put the mock in client/src/test/mocks? Or, alternately, if you're importing the `generatePassword` function directly from a node module, then we could wrap it in our own function and mock it in place, similar to what we do with the API client.

SMaxOwok:
I'll take a look at doing that.  I was going off of https://facebook.github.io/jest/docs/manual-mocks.html.

> If the module you are mocking is a node module (eg: fs), the mock should be placed in the __mocks__ directory adjacent to node_modules.

PR #469, opened 2017-07-27
[B] Only open tweet events in new tab

### LINKED ISSUE #468: Start reading event link opens in new tab. It shouldn't.
image

(some event links should open a new tab, such as links to tweets — but this should not)

SMaxOwok:
I added a function to determine this in case there are more event types later.  Let me know if you think it's overkill and would rather a variable be set in render.

Fixes #468


zdavis:
Your commit message seems inaccurate... what does this change have to do with "navigating back to home"? Isn't the issue that the content is being opened in the wrong target?

Issue #419, opened 2017-07-03
Allow activity to be hidden on a project-by-project basis.


zdavis:

zdavis:
Let's do some refactoring as part of this... add a new tab for the project called "Appearance". Move the current cover and hero fields from the general tab to this new tab. Add a switch labeled "hide project activity" and make a corresponding field on the project model (boolean, default false). If the switch is true, we don't show activity on the project landing page. There's some weird spacing/background color logic on the project detail page, as I recall. You may need to do some refactoring there to keep it sane as we make lists configurable.

SMaxOwok:
Per slack:

> it occurred to me that maybe it makes more sense for the "hide activity" switch to be in the activity tab as "hide activity on project page" or something

i guess the hero and cover could still live in an appearance tab, along with anything else that comes along for that

but it feels a little weird to go to a separate tab to hide the activity instead of in the activity tab near the top

maybe as a switch that automatically saves, so there's no "save project" button

zdavis:
Ok.. thanks for your patience on this. Let's do the following:

- I agree that a separate tab makes sense. Let's call it "Project Page" and put it below "general" in the left navigation
- Reorganize fields so that fields related to content that really only shows on the project page is on this new tab, and in more or less the same order: 1) hero, 2) cover, 3) description, 4) purchase fields, 5) social fields, 6) events visibility toggle
- Let's move draft and featured to the top of the general tab — can we split the first row into two columns for those? Might look less weird.

PR #470, opened 2017-07-27
[F] Add setting to hide event activity in project

SMaxOwok:


zdavis:
Is this really in progress? Can't remember where we left it...

SMaxOwok:
@zdavis this is just waiting on an opinion for the comment on #419

zdavis:
OK — I added some feedback that should unblock you on this.

PR #484, opened 2017-08-01
[F] Implement collection annotations

### LINKED ISSUE #405: Admins can add resource collection annotations

SMaxOwok:
Closes #405 

Includes significant refactor of reader resources to be more generic reader notations, which can be of type 'resource' or 'collection'.


zdavis:
We have design on this now, which needs to be implemented.

client/src/theme/Components/_index.scss
```diff
@@ -5,6 +5,7 @@
 @import './loading-bar';
 @import './search-menu';
 @import './share-nav-primary';
+@import './toggle_primary';
```
naomiyaki: I moved this to the backend components folder. Even though it's currently used on the front-end, it's used in the context of "backend" styles.

client/src/theme/Components/_toggle_primary.scss
```diff
@@ -0,0 +1,40 @@
+@mixin togglePrimary($height) {
```
naomiyaki: Since this style/component has its own partial and isn't being used in multiple contexts, we don't need a mixin here.

client/src/theme/Components/_toggle_primary.scss
```diff
@@ -0,0 +1,40 @@
...
+  position: relative;
+  height: $height;
```
naomiyaki: I don't think we need a fixed height here unless one day there are multiple toggles next to eachother that need to be the same height.

client/src/theme/Components/_toggle_primary.scss
```diff
@@ -0,0 +1,40 @@
...
+    display: flex;
+    height: 100%;
+
+    .toggle-option {
+      @include utilityPrimary;
+      display: flex;
+      align-items: center;
+      flex-grow: 1;
+      justify-content: center;
+      height: 100%;
```
naomiyaki: I added the font-size from the comps.

zdavis:
@SMaxOwok another quick schema change went out — can you rebase on master and regenerate the schema again?

naomiyaki:
This needs some responsive attention, which I will give!

zdavis:
Also, change the "resources" label in the annotation popup back to "resources" — there are some issues after rebasing on master; Max is fixing.

naomiyaki:
Want to check on this one! Does it need review/changes from me?

zdavis:
I have significantly refactored this PR. I need to do some final testing, and then it should be ready.

zdavis:
Also probably need to check/fix tests.

zdavis:
Ready to accept as soon as tests pass.

PR #486, opened 2017-08-02
[C] Adjust deployment approach

zdavis:


zdavis:
Ignoring hound here, because we don't apply the normal linting to our cap config and tasks.

PR #497, opened 2017-08-03
[F] List all highlights and annotations for a given text

zdavis:


naomiyaki:
@zdavis All the styles are in and working down to mobile, but I want to refactor the overlay styles, it's the right thing to do!

naomiyaki:
@zdavis I've refactored the resource overlay styles so that they use the same wrapper class/partial. I'm about to head out for the weekend, but either one of us can refactor the resource Overlay component to use your new HOC now. Let me know if you take it on!

naomiyaki:
Max to do the refactor mentioned above with message from Zach incoming!

PR #510, opened 2017-08-07
[B] Prevent sign in overlay from updating on authentication

### LINKED ISSUE #473: Edit profile view is briefly flashed after login, which shouldn't happen
The flow code around login/signup is some early code we wrote. I'm sure it could use some improvements. See if you can figure out why this is happening and let's talk about how to fix it.

SMaxOwok:
Fixes #473 

This bug was caused by the component updating after it's props changed before the `hideSignInUpOverlay` function was called.  There is never a case where the component should transition from `login => update` or vice-versa, so we don't have to try and determine which of those to use after it's created.  Instead, we can initialize the component state with the view we want, rather than deferring to the default case of the `renderChild` function.

Additionally, rather than passing down function props for each view change, this commit refactors the `childProps` function to only pass down a single function to handle changing views.

### DIFF
--- a/client/src/components/global/SignInUp/Create.js
+++ b/client/src/components/global/SignInUp/Create.js
@@ -15,8 +15,7 @@ class CreateContainer extends Component {
     response: PropTypes.object,
     user: PropTypes.object,
     authentication: PropTypes.object,
-    showLogin: PropTypes.func.isRequired,
-    showCreateUpdate: PropTypes.func.isRequired
+    handleViewChange: PropTypes.func.isRequired
   };
 
   static mapStateToProps = state => {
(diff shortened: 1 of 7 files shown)


zdavis:
This PR doesn't get us to the desired behavior, in which the user sees the "welcome" + edit profile form after they signup. That behavior appears to be busted in master currently. Please adjust accordingly.

SMaxOwok:
Sorry, I didn't realize we were missing that behavior.  I've added a callback after the user gets created to set the view to the welcome form and prevent the overlay from closing in that case.

zdavis:
Perfect!

PR #511, opened 2017-08-07
[B] Fix error on user sign up

SMaxOwok:
Fixes 509

### DIFF
--- a/api/app/controllers/api/v1/users_controller.rb
+++ b/api/app/controllers/api/v1/users_controller.rb
@@ -30,7 +30,7 @@ module Api
 
       def create
         @user = ::Updaters::User.new(user_params).update(User.new)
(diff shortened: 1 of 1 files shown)


api/app/controllers/api/v1/users_controller.rb
```diff
@@ -30,7 +30,7 @@ def show
 
       def create
         @user = ::Updaters::User.new(user_params).update(User.new)
-        include_password = user_params[:data][:meta][:created_by_admin] == true
+        include_password = user_params.dig("data", "meta", "created_by_admin") == true
```
zdavis: In the future, you can use the Ruby safe navigation operator instead of `dig`. 

http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/

PR #514, opened 2017-08-07
[B] Don't show page nav buttons at mobile sizes

### LINKED ISSUE #512: Nav button redundancy on mobile layout
screen shot 2017-08-07 at 1 46 47 pm

Big button should be hidden when footer nav is added

SMaxOwok:
Fixes #512

Also makes footer nav active states to be more specific to prevent "Projects" being active on featured page.

### DIFF
--- a/client/src/components/frontend/Layout/ButtonNavigation.js
+++ b/client/src/components/frontend/Layout/ButtonNavigation.js
@@ -19,6 +19,7 @@ export default class LayoutButtonNavigation extends Component {
     grayBg: PropTypes.bool,
     showBrowse: PropTypes.bool,
     showFollowing: PropTypes.bool,
+    hideAtNarrow: PropTypes.bool,
     authenticated: PropTypes.bool
   };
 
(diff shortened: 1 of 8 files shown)


zdavis:
It's hard for me to speak to what you're suggesting, Naomi, because I don't know what the impact would be. I'm not sure what it means to hide all "button navs" — does that mean everything with the .button-nav class? If so, how widely used is it? 

If you want to make this change, I'm not for it until I understand what the impact will be. Can you clarify?

Zach

naomiyaki:
I updated this branch and kept Max's active state fix (for the mobile nav) but changed the way the button nav gets hidden. Instead of being a hardcoded class on the button nav component, It's now a non-default property that's added in template to hide a button nav on mobile. @zdavis @SMaxOwok Can you have a quick look at this and give me your thoughts? If it looks good I can squash and push again.

zdavis:
This looks fine to me. I can squash it and accept.

PR #517, opened 2017-08-09
[F] Add resource detail zoom functionality

### LINKED ISSUE #407: Zoom functionality on image resources

SMaxOwok:
Closes #407

### DIFF
--- a/api/config/manifold.yml
+++ b/api/config/manifold.yml
@@ -263,6 +263,8 @@ common: &1
           - "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
           - "application/vnd.openxmlformats-officedocument.presentationml.presentation"
           - "application/zip"
+          - "application/markdown"
+          - "application/epub+zip"
           - !ruby/regexp '/text\/.*/'
           - !ruby/regexp '/audio\/.*/'
           - !ruby/regexp '/video\/.*/'
@@ -293,7 +295,8 @@ common: &1
           - !ruby/regexp '/ppt?x\Z/'
           - !ruby/regexp '/md\Z/'
           - !ruby/regexp '/dmg\Z/'
-
+          - !ruby/regexp '/txt\Z/'
+          - !ruby/regexp '/epub\Z/'
 production: *1
 development: *1
 test: *1
--- a/client/src/components/frontend/Project/__tests__/__snapshots__/Resources-test.js.snap
+++ b/client/src/components/frontend/Project/__tests__/__snapshots__/Resources-test.js.snap
@@ -127,57 +127,64 @@ exports[`Frontend.Project.Resources component renders correctly 1`] = `
           <li
             className="resource-card"
           >
-            <div
-              className="resource-link"
-              onClick={[Function]}
-            >
-              <div
-                className="resource-thumbnail-primary "
-                style={
-                  Object {
-                    "backgroundImage": null,
-                  }
-                }
+            <div>
+              <span />
+              <span
+                onClick={[Function]}
               >
                 <div
-                  className="wrapper"
+                  className="resource-link"
+                  onClick={[Function]}
                 >
-                  <figure
-                    className="resource-type"
+                  <div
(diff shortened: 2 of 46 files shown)


SMaxOwok:
One thing to note:
The overlay will look weird because of the header until #484 is merged in.  That PR includes a change to the global overlay that will hide the overlay header if there is no title.

zdavis:
#484 has been merged to master, and this branch has been rebased onto Master.

zdavis:
I'm seeing a few small things that need to be corrected here:

- Until we can think through zoom a bit more, I think we should remove the zoom button from resources and collections opened in the reading overlay. Having it there leads to a confusing double overlay with two levels of close buttons.

- If you view a collection from the project page, there's no zoom for the resources within the collection slide show. Can we address this?

- The lack of transition on zoom open/close is jarring. Can you get with @naomiyaki  and add a fade transition, or something to soften and refine it a bit? This could probably be something that's added globally to this overlay.

zdavis:
Looks like there's also a snapshot to update after rebasing this onto Master.

SMaxOwok:
I changed the zoom functionality to be a wrapper component, so that we only need the open/close login in one place.  The way I'm disabling the zoom for the reader doesn't feel great, so I'm open to suggestions on how to better handle that.

zdavis:
Max, when you say " The way I'm disabling the zoom for the reader doesn't feel great, so I'm open to suggestions on how to better handle that" could you be more specific? Point me toward the exact line in you code that you're referring to, so I don't have to hunt for it. Thanks!

zdavis:
I did functional testing on this and like the way it works. Still need to do a code review though.

zdavis:
This PR causes problems for many resource types that are not images. Videos, for example, will open a broken overlay. I'm going to fix it.

zdavis:
@SMaxOwok I've done some extensive refactoring to your work in 20d55b. In a nutshell, I did the following:

*Rename "ZoomableResource" as it was solving the problem too narrowly. What you're doing here is less about being able to zoom on a resource, and more about handling that preview link on the resource cards. Clicking on an image resource should zoom it, yes. But what happens when we click on a video or audio resource? We need players in that case.

*Your image preview was called "overlay" which is not quite the right  name. The management of the overlay mechanism was split between the zoomable resource and the overlay component, which is unnecessary. I've broken this apart so that there is a preview component for each previewable type, and this preview could be used anywhere. I've move the globalOverlay component up out of the image-specific component and into the wrapping component to avoid repeating logic throughout all preview types.

*These changes, and the broadening of scope on this PR to cover other resource types means that we need to abstract some players (video player, for example) out of the slide components so that they can be used in preview overlays and in resource slides without duplicate code.

We need a mechanism to determine if a resource is previewable. Your solution, as you had it, broke videos and other resources. Clicking on a video preview link on a card would download the video (previous behavior) and open an empty overlay. You gotta test your changes more carefully, and try to think through the impacts your changes might have. In your PR, you wrapped ZoomablePreview around everything, but didn't implement a check to see if a resource is actually previewable, or write code to handle what happens if someone tries to preview a resource that can't be previewed.

*I wrote static methods on the preview component to report whether or not a resource is "previewable". I refactored the "handlePreviewClick" method in Card.js. It was a confusing and weird switch statment, and now it's a series of (IMO) more clear conditional statements.

zdavis:
I'm happy with the code and how this works. I need to check tests, which probably are no longer passing, and lint the styles. Then the commits need to be reorganized and recommitted, at which point this should be ready to accept.

zdavis:
Note to self: 

There is a zoomable prop on the Resource.Hero component. The Resource.Hero component appears to be a largely unnecessary wrapper around the slide component, so that may need refactoring. In any case, I think that instead of calling this prop zoomable, it should be called slideProps, and just be an arbitrary set of props that get passed through to the wrapped slide object. It's weird to have a zoomable prop for, say, a video hero, since it can't be zoomed.

Issue #501, opened 2017-08-03
Update hero layout with draft button, streamlined makers list


naomiyaki:
![gpu preview 2017-08-03 13-38-43](https://user-images.githubusercontent.com/1628239/28942688-21be2b88-7851-11e7-9ca2-c9e791afeed9.png)

naomiyaki:
@zdavis I need the draft feature to be merged to complete! Please have a look and merge when ready.

zdavis:
@naomiyaki that feature has been merged to master, so you should be unblocked on this.

naomiyaki:
Squashed and pushed!

PR #542, opened 2017-08-21
[F] Update hero section markup

naomiyaki:
[Delivers #501]
Update project hero with simplified maker listings and add draft label


zdavis:
Is the gray background correct? Seems too light:
image

Spacing between desc. and button seems too tight; buttons seem less tight (taller?) than comps:
image

Buy button text isn't right:
image

Instagram icon doesn't match comps:
image

No italics on contributor name in comps:
image

Too much bottom padding on hero compared to comps:
image

Mobile line height issues on authors:
image

I realize typography is a moving target, but I wonder if the size of the desc. text is a bit too large. Thoughts?

client/src/components/frontend/Project/Hero.js
```diff
@@ -14,17 +14,78 @@ export default class ProjectHero extends Component {
...
-        </section>
+        </div>
       );
     }
+    return (
+      <div className="project-creator-list">
+        <ul>
+          {creators.map(creator => {
+            return (
+              <li key={creator.id} maker={creator}>
```
zdavis: This will throw a warning in the console because you can't have a maker prop on the `li` component. Why is this here?

client/src/components/frontend/Project/Hero.js
```diff
@@ -14,17 +14,78 @@ export default class ProjectHero extends Component {
     project: PropTypes.object
   };
 
-  listMakers() {
+  haveAvatars(creators, variant) {
```
zdavis: I'd rename this method to "hasAvatars"

client/src/components/frontend/Project/Hero.js
```diff
@@ -14,17 +14,78 @@ export default class ProjectHero extends Component {
...
+    });
+
+    // If none of the creators had a null avatar
+    return result;
+  }
+
+  listCreators() {
     const creators = get(this.props.project, "relationships.creators");
-    if (creators && creators.length > 0) {
+    if (!creators || creators.length <= 0) return null;
```
zdavis: Seems like this line of code could be abstracted into a `hasCreators` method.

client/src/components/frontend/Project/Hero.js
```diff
@@ -14,17 +14,78 @@ export default class ProjectHero extends Component {
...
+            );
+          })}
+        </ul>
+      </div>
+    );
+  }
+
+  listContributors() {
+    const contributors = get(this.props.project, "relationships.contributors");
+    if (!contributors || contributors.length <= 0) return null;
```
zdavis: As above, seems like this simple check could be abstracted into a method: `hasPeople(type = "creators")` or something like that.

client/src/components/frontend/Project/Hero.js
```diff
@@ -14,17 +14,78 @@ export default class ProjectHero extends Component {
...
+    return (
+      <div className="project-contributor-list">
+        <span className="label">
+          {"Contributors: "}
+        </span>
+        <ul>
+          {contributors.map(creator => {
+            return (
+              <li key={creator.id} maker={creator}>
```
zdavis: This will throw a warning in the console because maker is an invalid prop for `li`

zdavis:
@naomiyaki Could you please prioritize completing this PR? I'd like this to be wrapped up before you begin any new work (eg, mobile resources).

naomiyaki:
On it!

naomiyaki:
@zdavis I did a force push that I believe addresses all of these issues except the Instagram icon. The icon we're using on the front-end is based off of Instagram branding changes that occurred after Lael made the original comps. I believe we had a conversation with Lael at some point about using this one instead.
I think the icon in the comps is clear, but I'm generally inclined to use official brand guidelines since it fits in just fine with the Manifold design.

zdavis:
A number of my comments above weren't addressed, including:

1) The buy link text is still incorrect
2) Contributors text style is still not correct. Italics has been removed from all of it, not just the names as requested.
3) There is still too much bottom padding, IMO, on the hero image (too much space below the social icons)
4) The spacing between title, subtitle, authors, and contributors is still horsey. Happy to fine tune this with you at your desk, if needed — but if you compare to the comp, you'll see that the comp is generally tighter here. I'm thinking authors needs to move up a bit, and contributors needs to be a but closer to authors:
image


There's also at least one new bug that I didn't see before:
1) The text on the draft label is no longer visible

naomiyaki:
@zdavis I've force pushed another set of changes:
**Draft label colors:** Fixed to match comps

**Space between title and subtitle:** There's no example of Title + Subtitle in the UI Updates file with the new hero. The spacing was already much less than on the original comp (UI Projects), but I lowered it even more. Please let me know if this works for you.

**Space between subtitle and authors, space between title (without subtitle)  and authors:** Exact to the comps within 1 pixel

**Space between author (no avatars) and contributors**: Exact to the comps

**Space between author (avatars) and contributors:** No design provided so use same spacing as above

**Space between contributors and divider line:** Exact to comps

**Space between authors (with/without avatars) and divider line:** Exact to comps

**Space between authors (with avatars) and contributors:** No example given, so used same spacing as without avatars plus a pixel or two.

**Space between social and hashtag:** Exact to comps

**Bottom padding on hero:** This was exact to comps before, so I lowered it slightly. Please let me know if you want further adjustments.

zdavis:
@naomiyaki 

Just a reminder to avail yourself of Lodash methods to make your code easier to understand. Instead of this:

```
  hasAvatars(creators, variant) {
    let result = true;
    creators.forEach(creator => {
      const hasAvatarVariant = get(
        creator.attributes,
        `avatarStyles.${variant}`
      );
      if (!hasAvatarVariant) {
        result = false;
      }
    });
    return result;
  }
```

Just do this:

```
  import some from "lodash/some";
  import isEmpty from "lodash/isEmpty";

  hasAvatars(creators, variant) {
    return !some(creators, (c) => isEmpty(c.attributes.avatarStyles[variant]));
  }
```

The lodash version has the benefit of being more performant. With a forEach statement, you loop through every item in the collection. With `some`, the loop terminates as soon as it finds an item in the collection for which the predicate returns true. The shorter version is more legible, too, and it's easier to see what's happening.

PR #548, opened 2017-09-05
[B] Update slideshow styles for mobile

### LINKED ISSUE #482: Resource slideshow controls don't match comps for mobile
screen shot 2017-08-01 at 8 25 53 am
screen shot 2017-08-01 at 8 25 58 am

naomiyaki:
[Fixes #482]

Also add swiping functionality to slideshow on mobile; Update slideshow
transition styles to slide left/right instead of a wipe transition;
Remove inline width styling used with old transitions.


naomiyaki:
I forgot to note something on this one! In the comps, the download button has been moved out of the collection utility (where read more is) and into the button group. I wasn't sure how to accomplish this, since the button list is separate from the active slide data. But I wasn't sure if this was a necessary change, either. I believe I removed the download link from the slider for now, but will add it back if we decide not to use the button group.

gpu preview 2017-09-06 09-35-18

zdavis:
RE: your comment about the download link. I think the comps haven't fully thought through the requirements here. The comps show a download button on the collection overlay, but what would that actually download? The entire collection? I think the correct solution here is to leave the slideshow download link in place, as it is outside of the reader. This strikes me as more clear, since the download UI is more closely tied to the individual slides.

zdavis:
Passing this back to Naomi for additional revisions.

1. My changes remove the description DOM element if it's not set. The goal is to not have a big chunk of whitespace between links and slide when there's no description. Doing this breaks things on mobile:
image

2. The bold state on the links is inconsistent. On desktop, read more and details are bold.
image
But on mobile, only download is bold:
image
Per the comps, _only_ read more should have this style applied.

3. The width of the title sets the width of the buttons nav on mobile. You won't be able to see the problem until #1 is addressed, but it seems that narrow titles can lead to button icon/text wrapping where it shouldn't

4. There is a regression (likely unrelated to this work) in relation to mobile resource cards. At least two are rendering incorrectly on my instance:
image
image

zdavis:
I'm happy with this — it can be accepted once CI passes.

PR #561, opened 2017-09-11
[F] Add mobile notation resource viewer

naomiyaki:
[Delivers #406]

Add a mobile notation resource viewer. The viewer appears responsively when the window is too narrow to show the desktop viewer, which varies depending on the user's margin preferences.

The preview shows one notation at a time: the one closest to the bottom of the user's screen. On desktop, the user can hover over another notation marker to see it's preview. On touch devices, the user can click a notation marker to set it as "active" and show it in the footer preview until it is scrolled out of view*. 

The user shouldn't see anything if there are no notation markers in view.
Ready for stress testing and QA.

*This may present a problem on desktop/laptop devices with touch functionality. One solution would be to conditionally enable this based on a media query, but that would involve having the same media query in CSS and JS, which is something I don't like! We could also refactor this so any media query dealing with marginalia is handled by JS, but that might require adding more resize event handlers, which can also be annoying. Please let me know your thoughts!


zdavis:
Rebased onto master and force pushed.

zdavis:
There are some issues around when this feature gets activated (it only seems to work if I reload the page), and I want to make some changes to how the active state of the notation markers is expressed. I'll make these fixes myself, as it will give me an opportunity to better acquaint myself with your changes.

However, there are a couple things I'd like you to do:

1. When there is no thumbnail, we should show the correct generic thumbnail. Right now you're showing only the icon. See page 8 of the reading comps:
image

2. The margin above and below the resource thumbnail should match, per the comps.

3. The resource title should be top aligned with the top of the thumbnail. It seems a few px off currently.

4. The comps show a green arrow to the right to indicate that you can click through to the resource modal. I'm not seeing that here.

5. It's unclear to me what the logic is for hiding and showing the text controls nav bar at the bottom. It seems almost random — am I missing something?

naomiyaki:
Zach,

I've done a push to address these visual changes. A few notes:
This is no longer up to date with master. It'll need a rebase once we're done. I didn't do this because I know you wanted to add some work, but please let me know if you'd like me to.

The show/hide of the mobile button menu is supposed to be similar to the way mobile Safari shows and hides the browser chrome. I've done a few refinements here as part of these commits, but as a rule:

At the top of the page, the button menu should always be visible.
Once the user scrolls down, it will hide.
If the user scrolls up (say 2~3 lines of text) it will reappear.
I've added a threshold so that the button menu doesn't show or hide if the user only scrolls a bit in either direction. Getting this right may be the key to making sure it feels right.

Also, I'm unable to reproduce the missing arrow icon (see screenshot). The right carat is a new icon, so you may need to restart webpack in order to see it.

screen shot 2017-09-13 at 9 41 08 am

zdavis:
Not quite working yet. See https://www.useloom.com/share/23975a4cf494447695bd9ea75913389e

zdavis:
Here are the things that jump out at me (that I'm surprised you didn't catch during your review of your changes):

1. There is no animation on the mobile resource bar. It appears and it disappears, but we seem to have lost the transition.

2. Icons in resource thumbnails and collection thumbnails are no longer sized correctly. The icons are considerably larger than they should be.

zdavis:
I've made some improvements to this branch. Namely:

1. Restored the transition that you removed when you rebased this onto master in d2873bb.
2. Renamed some methods for clarity
3. Corrected the issue where the preview only worked if the markers existed in the server-side DOM
4. Simplified `updateVisibleMarkers` and renamed it to `updatePreviewEntry`. 
5. Removed the dom markers from the component state because it was unreliable and would have meant adding a time out. The list component mounts before all of its children (which includes the markers) have mounted. Therefore, querying the dom for those markers in List's componentDidMount won't work. The performance gain you get by effectively caching your dom query for markers in component state doesn't outweigh the complexity we'd take on by dealing with timeouts and callbacks to solve this. Therefore, we just query the dom for markers when we need them, rather than caching it.
5. Improved, I think, the algorithm for figuring out which marker is visible. Previously, you were looping through a collection of all markers, calculating the position of each, filtering it, and getting the last one. With my change, we query the dom for the collection, reverse it, and then find the first marker that meets the criteria. Once we hit that marker, we stop calculating positions. This works because querySelectorAll can be relied upon to return dom elements in order.
6. Removed the code for passing the "active" annotation down to the preview. It didn't seem like this was working, and it's not really the right way to handle it. We already have actions in place for "activating" a marker, which is what's used in the marginal notation viewer on hover. We are now passing these same bound actions to the preview. When the preview is mounted, it dispatches an action to activate the corresponding marker. When the content of the preview changes (in response to scroll), we dispatch another action. This way, it's easy to see what marker the preview relates to.

zdavis:
Oh, and I've also reworded all the commits (and fixed the deliver vs fixed language in the extended commit messages, as well as removing references to this PR which is unnecessary — it's best if commit messages only reference issues, since they can only close issues, not PRs. I've made it clear that all of these commits are to be fixed up into a single commit when this is ready.

The last outstanding issue in this branch, that I can see, is that we need to fix the sizing of the icons in the thumbnails.

naomiyaki:
I've gone through and fixed/refined the icon sizes in the viewer, and made adjustments for tall v.s. wide icons.

I also found another bug (perhaps unrelated or maybe another refactoring issue) while testing this, where notation group captions did display raw HTML set in the backend the way that resource titles do. (See screenshot)
manifold scholarship metagaming one 2017-09-18 15-33-22

I made a commit fixing this as well, but only based on my best judgement as I haven't worked with parsed HTML on Manifold this way yet. Please have a look.

I have a couple of questions about the changes to the "active" functionality with respect to the mobile viewer. I'm not sure if this is the intended behavior or not, but it seems that now on desktop the bottommost resource is always being highlighted on scroll. 

Hovering or clicking the resource markers no longer changes what's shown in the mobile footer/viewer, so it's not possible to open a non-active resource in the overlay on mobile.

manifold scholarship metagaming one 2017-09-18 15-46-20

zdavis:
Waiting on me.

zdavis:
I think I finally have this how it should be. I'm going to push my changes and accept this PR shortly, but wanted to record the desired behavior here for posterity:

1. The preview component should never set which notation is active. It should only render the currently active notation.
2. If a user clicks on a notation marker, that notation becomes active and it remains active until it has scrolled out of view.
3. If a user has not clicked on a marker, we automatically activate the last visible marker on scroll. This activation is flagged as "passive" to distinguish between an non-passive activation that is triggered by a click on the marker.
4. The list component has to have some knowledge as to whether or not the preview is visible, because the automatic activation of markers only occurs when the preview is visible.
5. Touching or hovering on a marker will "activate" the notation. Clicking on it will go to the view.

PR #565, opened 2017-09-11
[B] Use actual reply count for 'view replies' text

### LINKED ISSUE #546: Correct hard coded reply count in view.
See https://github.com/ManifoldScholar/manifold/blob/6a121eeb867770b32d2a941b8ddeec86982aca2b/client/src/components/reader/Annotation/Annotation.js#L191

The count is hard coded at 36.

SMaxOwok:
Fixes #546


zdavis:
Your approach here assumes that all replies are always loaded, which is not a safe assumption. What if a thread has 200 replies? Are you sure we'd load them all at once? I believe the API response is in fact paginated for comments, so you'll want to account for that.

Here are some things to consider:

1. On the annotation comment threads, the initial request for comments on a given annotation would set the perPage pagination to 1. This would give you the first comment and the total count, which you'd get from the meta.pagination property on the request.

2. The next request could return, say, the next 50 comments (so, you could repeat the original request with perPage set to 50 instead of 1).

3. If there are more than 50 comments, the link would change to say "load more comments", and you'd request page 2 of comments with 50 per page.

4. This gets tricky, because the response needs to be added to the previous response in the store, rather than replacing it. There are different ways of coming at this. In the resource slideshow, for example, we store the resources in the components state. When it gets a new set of resources, it adds it to the state. Another approach could be to automate this via our entityStoreMiddleware. At line 120 of that file, you'll see how we add models to existing responses. It currently only works for adding a single model, but we could come up with a similar method for adding a paginated request to a previous request in the store.

Think about this a bit more, and discuss with me if necessary.

zdavis:
@SMaxOwok Adjust the text per our discussion in slack, and make sure tests pass, and this is good to accept

PR #568, opened 2017-09-11
Fix null collection image in reader; refactor thumbnail styles

SMaxOwok:
- Fixes an issue with collection thumbnails not being removable
- Refactors resource/collection base thumbnail styles into a mixin.
- Fixes an issue with reader collection notations not displaying when there is no thumbnail

### DIFF
--- a/api/app/controllers/concerns/validation.rb
+++ b/api/app/controllers/concerns/validation.rb
@@ -34,7 +34,7 @@ module Validation
 
   def collection_params
     params.require(:data)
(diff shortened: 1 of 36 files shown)


SMaxOwok:
@naomiyaki I included a decent-sized refactor of the collection/resource thumbnail styles.  Would you mind taking a look and making sure it is ok?  I looked at as many places where those styles and others, such as `resource-type`, were being used and I think it looks ok, but I would appreciate another set of eyes.

zdavis:
@naomiyaki says it looks good.

PR #575, opened 2017-09-12
Text wrap by block on back link and breadcrumb

naomiyaki:
This addresses wrapping on project back links and also ports these updates to the backend breadcrumb:
manifold scholarship 2017-09-12 12-49-05


zdavis:
I don't see any evidence that this fix solves the problem in either place:

image

image

image

Issue #566, opened 2017-09-11
Update TOC design to match latest comps


naomiyaki:

naomiyaki:
This should include styles and functionality for "about this text" button

PR #579, opened 2017-09-13
[F] Add new TOC menu/footer styles

naomiyaki:
Starts #566

This adds new styles for the TOC (and about link).
@zdavis You said you wanted to take a look at this to figure out how to best pass the active text section into the TOC so we could have an active state.
In the meantime I can work on the metadata overlay, but I probably have enough to do first that I can wait for you!


zdavis:
@naomiyaki You should be unblocked on this with my commit e07bdd2. This commit adds an [ACTIVE] string to the end of the currently active TOC item. You should be able to remove that string and replace it with the correct styles.

Note that setting the active state will only happen in the client side render. Because EPUBs often use anchors in TOC links, we can't reliably set the active state server-side, because the server does not have access to the request hash. So, we defer setting the active state until after the component has mounted. Once it mounts, we know that we're on the client, and we can set the active link accordingly.

Issue #556, opened 2017-09-09
Display error in backend select inputs


zdavis:
![image](https://user-images.githubusercontent.com/255909/30235076-ddc013a4-94b8-11e7-847a-7fda8ecb9f66.png)

PR #581, opened 2017-09-13
[B] Update select group input padding

naomiyaki:
Starts #556 

This addresses an issue where a dropdown arrow collides with text in the dropdown. That said, the maximum size of this select is small enough on most sizes that the longest text string (in this case "Project Created") still gets cut off. (See screenshot)

It's inevitable for some longer selects to get cut off, but the width of this select group is being truncated due to style used in the resource/collection filter (screenshot 2).
To mitigate this, I've added an inline class that allows the filters to sit aside the search input in just these cases.

screen shot 2017-09-13 at 1 36 35 pm

screen shot 2017-09-13 at 1 50 09 pm

Please have a quick QA and let me know if I missed any cases where the inline property needs to be applied.

### DIFF
--- a/client/src/components/frontend/Project/__tests__/__snapshots__/Resources-test.js.snap
+++ b/client/src/components/frontend/Project/__tests__/__snapshots__/Resources-test.js.snap
@@ -40,7 +40,7 @@ exports[`Frontend.Project.Resources component renders correctly 1`] = `
         />
       </div>
       <div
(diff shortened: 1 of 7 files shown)


zdavis:
This looks good to me. My issue was mainly about the text rendering behind the arrow, and it looks like this solves it. If I'm ok with it, are you ok with it being merged?

PR #610, opened 2017-09-25
[B] Add initial filter to searchable list for resets

SMaxOwok:


zdavis:
It just so happens that I fixed this today as well in 8b46c14afa69c47151e37154656eed4e14f67967.

PR #615, opened 2017-09-28
[B] Fix ingestion of zips with root dir

SMaxOwok:


zdavis:
I think this would benefit from a test. You could add a sample Markdown ingestion fixture, one with the root dir and one without, and show that it works for both.

PR #619, opened 2017-10-02
[F] Adjust form styles for mobile

naomiyaki:
Closes #589

PR #624, opened 2017-10-03
Update text list button/table styles for mobile

naomiyaki:
Note that this PR is blocked as its dependent on new button styles in #619. Once that PR is accepted, this one can be QA'd (by Naomi), rebased, and then released for review.

### DIFF
--- a/client/src/containers/backend/ProjectDetail/Texts.js
+++ b/client/src/containers/backend/ProjectDetail/Texts.js
@@ -431,19 +431,23 @@ export class ProjectDetailTexts extends PureComponent {
 
         {this.renderRoutes()}
 
-        <div className="buttons-icon-horizontal">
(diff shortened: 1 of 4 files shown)


zdavis:
@naomiyaki #619 has been merged, per your comment above, and I've removed the blocked label. There is a conflict, however, which needs to be resolved. Please address, review, and let me know when this is ready.

naomiyaki:
@zdavis Merge conflict fixed and styles updated a teensy bit based on merged styles from #619

zdavis:
Accepting this for now. I think we need take another pass at this in the future, as the layout still doesn't feel right. That said, it's much better than it was — doesn't feel broken any more. Thanks.

zdavis:
For posterity, this is where we landed:

image

PR #650, opened 2017-10-31
[F] Implement interface to manage/assign subjects

SMaxOwok:
I added this to the General tab of the Project backend.  I had originally made it into its own tab, but then it seemed like it fit well enough in the General group.

Since you can create/assign from the HasMany form, we only needed a way to delete subjects.  I talked with @scryptmouse about it a bit.  I had originally been trying to implement a callback to destroy a subject if it had no projects associated with it, but Alexa suggested running a job with clockwork to destroy any unassociated subjects.

The subject filter on the home page does not update without a refresh when a new subject is created.  I can fix this if desired, but I thought that since the only time this occurs is if a user is currently on Manifold when a subject is added, that it may not be worth re-fetching the subjects in the `Frontend` component.


zdavis:
If I add a subject to one project, and then go to another project and try to add it to that same subject, the subject doesn't appear in the list as I'd expect it to.

I think we need an interface to manage all the subjects rather than automatically deleting empty subjects. Let's add a "subjects" tab in the settings section of the backend. The list should work exactly like the people lists do. We list all the subjects and use the side drawer for editing a subject (which for now might only have a single field).

zdavis:
I wasn't able to add a new subject to a project without going to the subjects page first. The UI for assigning a subject to a project should accept new subjects as well as existing, similar to how the people inputs work.

zdavis:
Also, be sure to add subjects to /api/lib/tasks/search.rake (used to reindex searchable records).

zdavis:
Could you go ahead and rebase your commits before I review? It seems the last CI run failed, so doing this should trigger a new run.

PR #651, opened 2017-10-31
[B] Use installation name in project created event

### LINKED ISSUE #640: Use settings.installationName in the new project event text
image

Replace "manifold" with settings.installationName.

SMaxOwok:
Fixes #640

### DIFF
--- a/api/app/services/factory/event.rb
+++ b/api/app/services/factory/event.rb
@@ -1,6 +1,10 @@
 module Factory
   # Factory class for creating Event models
   class Event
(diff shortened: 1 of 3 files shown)


SMaxOwok:
There could be a little complication with the indefinite article now.  We could use a gem like https://github.com/rossmeissl/indefinite_article, but it seems like even that isn't perfect.  We could also supply the article somewhere as well, e.g. settings.

Let me know if/how you want to try tackling this.

PR #653, opened 2017-11-03
[WIP] Adding user presistent UI settings

SMaxOwok:


client/src/store/reducers/ui/persistent/reader.js
```diff
@@ -1,8 +1,26 @@
-import { combineReducers } from "redux";
+// import { combineReducers } from "redux";
 import colors from "./reader/colors";
 import typography from "./reader/typography";
+import { handleActions } from "redux-actions";
 
-export default combineReducers({
+const initialState = {
```
SMaxOwok: This is where I'm struggling right now.  I want the `initialState` of `reader` to be in the shape of `colors` and `typography` which have their own `initialState`s defined.

zdavis:
See ed3f36f. Isn't this what you want? I think this is what the initial state should look like. The reader module just combines the colors and typography reducers; it doesn't provide any state of its own. With this commit, initial state looks like this:

image

SMaxOwok:
@zdavis this is almost working now (though obviously needs some refactoring).  The only actions that aren't triggering the user to update right now are the `incrementAttribute` and `decrementAttribute` functions.

SMaxOwok:
Closing in favor of a fresh PR

PR #657, opened 2017-11-14
[F] Make reader UI settings persistent for user

### LINKED ISSUE #516: Persist reader settings in user preferences
We need to expose some preferences on the user model. To start with, we need to save changes to the reading experience display and use them when the user returns to the reader. Talk to Alexa and/or I about the best way to approach this.

SMaxOwok:
Closes #516


zdavis:
Note to self — I think there's still an issue here with the SSR being inconsistent with the CSR that I need to address

zdavis:
Moved setPersistentUi call to componentDidMount to address the SSR/CSR mismatch.

PR #669, opened 2017-11-20
[F] Allow text section type to be changed

### LINKED ISSUE #503: Allow admins to choose a name for the sections within a text.
The previous and next buttons in the reader always say "next chapter" or "previous chapter," but not every text is organized around chapters. We should expose this naming on the text model and respect it in the reader.

SMaxOwok:
Closes #503


zdavis:
Ready to accept once CI completes.

Issue #643, opened 2017-10-16
Implement updated reader control design and functionality


zdavis:
Current:
![image](https://user-images.githubusercontent.com/255909/31617425-01f6c9cc-b244-11e7-8ca6-dec1d9719ab8.png)

Comps:
![image](https://user-images.githubusercontent.com/255909/31617434-0931e4d8-b244-11e7-92fe-2089ac20c2ff.png)

To do list is as follows. Page numbers refer to page # in Manifold_UI_Reading_ALL_v1.6.pdf

- [ ] Add "Your Notes" link to reader header (p. 1)
- [ ] Connect "Your Notes" link to see all annotations view (for that text) (p. 7)
- [ ] Finish implementing "see all annotations" components; include highlights (p. 7)
- [ ] Update design of user menu, per comp (p. 15)
- [ ] Adjust sizing in typography control menu to better match comp (p. 15)
- [ ] CSS and HTML for visibility menu, excluding "guided help" (p. 15)
- [ ] Make controls in visibility menu fully functional (p. 15)

zdavis:
Max, I'd like you to work on this, including the styles. You should get guidance from @naomiyaki on the CSS, but see how far you can take it.

PR #675, opened 2017-12-01
[F] Finish implementing my annotations view

SMaxOwok:
Note:  I see this is triggering a SSR mismatch.  @zdavis when you have a chance to review this, if you would also mind looking at the SSR stuff with me, I'd like to get a better grasp on what is causing that for this and future occurrences.


SMaxOwok:
Closing to open a bigger PR that will include all changes for #643

PR #677, opened 2017-12-01
[F] Implement updated reader control design and functionality

SMaxOwok:


zdavis:
This is a good start, but I'd like you to make some revisions.

First, I'm a little uncomfortable with how the `Header` component is getting overloaded with child components. I get why the various UI panels are in there, but having the TOC and the YourNotes component rendered from within the Header doesn't feel right. I wonder if it would be easy to move those out into the reader, since their visibility is already managed by values in the store. This isn't critical to me. It may be that these components belong in `Header` because of the `ScrollAware` component that wraps the header. I just found that as a reviewer, I wasn't expecting to find the `YourNotes` component inside the header.

Second, I'm not crazy about the organization of these components. For example, the detailed list of all components is called `Overlay`. I don't like that because Overlay describes a particular context for that list without describing that list itself. What if I wanted to include that in a view that wasn't an overlay? I think we need to refactor these components. I think we should start with one containers that is reponsible for fetching and managing annotation content:

containers/reader/ReaderNotes

That container would be connected and have a fetchData method. It would include logic for grouping annotations and for re-fetching if filters changed (it would maintain filter state and pass a method down to children that could be used to change the filter state). That's all this component would do. Don't worry about caching the API request or the grouping of annotations. We can come back to it later and improve performance, which won't be hard if all the API request / grouping stuff is isolated to that one container.

Then, you would have dumb components for each type of annotation list

components/reader/Notes/FilteredList (currently "your notes")
components/reader/Notes/DetailedList (currently "overlay"?)

The fact that these are in an overlay or a drawer should not be baked into these components, as that's a separate concern. Some refactoring might be in order here as well. Look at the `Toc` and `TocDrawer` components. Is there a way that we can make that `TocDrawer` into a generic reader drawer that can be used for both the your notes and the TOC drawer? The same should be true for overlays; it's better to have a reusable overlay than reinvent it each time we want to use it. Don't be hesitant to make improvements around this. Keep in mind, it's best if the drawer/overlay doesn't render its children until it starts opening, otherwise they start fetching data before the user opens the drawer.

Wherever you end up including the dumb components, you'd wrap them with the container component. For example:

```
<ReaderNotes>
	<FilteredList />
</ReaderNotes>
```

Reader notes could pass the flat annotations list, grouped annotations, dispatch, a filter update callback, etc to its child. The child gets to stay dumb, and can be used in other ways in the future. For example, if we wanted to use these same views for another user's annotations, we could perhaps wrap it in another component or modify ReaderNotes to take a user ID.

SMaxOwok:
@zdavis pushed some progress.  I have a list of smaller issues in this branch that still need to be worked on as well.

SMaxOwok:
@zdavis ready for review

zdavis:
Vertical alignment on label/checkboxes does not match comp:
image

api/app/serializers/annotation_serializer.rb
```diff
@@ -5,8 +5,9 @@ class AnnotationSerializer < ActiveModel::Serializer
 
   has_one :creator
   attributes :created_at, :end_char, :end_node, :id, :start_char, :start_node,
-             :text_section_id, :updated_at, :format, :resource_id, :creator_id,
-             :body, :private, :subject, :current_user_is_creator, :can_update_object,
+             :text_section_id, :text_section_name, :updated_at, :format,
```
zdavis: This seems problematic. For every annotation that exists in a text, the API will need to do an additional query for the corresponding text section, so you're in a N+1 query situations. I assume that the components on the frontend have access to the text and its sections, and we could determine the title from the text_section ID in the client, rather than making the API do more work.

zdavis: To be clear, "this" in that first sentence refers to the addition of :text_section_name

SMaxOwok:
@zdavis updated to address your comments.  There are a few that I was either stuck on or need clarification for:

5. I'm not replicating the filters not working.  They seem to be behaving appropriately for me.  Was there anything else to that bug?

8. I added navigation to the clicked annotation from the notes drawer, but am still missing opening the annotation drawer after navigation.  There are some issues regarding organization here, I think, as we need to manage the hash id through `Text` -> `Annotatable` -> `Section.Body`.  A couple of the issues are hash comparison (always === on first load), document not finding node (elements not loaded in time). This probably warrants a little more discussion on approach.

10.  We had briefly discussed the comments list in the DetailedList view for annotations and I thought the decision was to leave out comments/replies pending an update to behavior in the Thread component.  Would you mind confirming the behavior you want to see here?

zdavis:
Can't reproduce #5, so we'll call that done for now.

I'm going to look into #8.

Issue #678, opened 2017-12-14
In IE11 selecting text should pop up annotation menu


blnkt:
Working fine in other browsers, but in IE11, selecting text does not yield a pop up regardless of whether you are logged in

![no-pop-up](https://user-images.githubusercontent.com/8166060/34019286-f5bd04c8-e0e2-11e7-9c3a-d5ee79dd9778.gif)

Related to #637?

blnkt:
Updated this issue to reflect the fact that Edge started working apropos of nothing

PR #679, opened 2017-12-15
[B] Fix annotation popup bugs in ie

### LINKED ISSUE #637: Edge: Hover state artifact in annotation popup
virtualbox_msedge - win10_12_10_2017_10_51_21
virtualbox_msedge - win10_12_10_2017_10_51_36
virtualbox_msedge - win10_12_10_2017_10_51_42

blnkt:
RESOLVES #637 

After much investigation and experimentation I came upon this little used css property used to stipulate where the background image or color gets clipped in regard to the element's content-box, border-box, padding-box, etc.  I feel like in the mordern-era this distinction is usually moot but in THIS case it ended up being crucial.  Background-clip it is doing the job of never ever letting the background color creep under (gasp) or past (double gasp gasp) the border.  So now it looks all neat and tidy (as it does in other browsers).  Once #678 is fixed this kind of issue should be checked in ie11 too.

### DIFF
--- a/client/src/containers/reader/Annotation/Annotatable.js
+++ b/client/src/containers/reader/Annotation/Annotatable.js
@@ -74,6 +74,28 @@ class Annotatable extends Component {
     }
   };
 
+  nodeContainNode(parent, child) {
+    if (
+      child.anchorNode instanceof HTMLElement &&
+      child.focusNode instanceof HTMLElement
+    ) {
+      if (parent.contains(child.anchorNode)) return true;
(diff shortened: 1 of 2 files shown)


blnkt:
Tacking on another similar IE bug to this same pull request: #678

zdavis:
@blnkt could you please adjust your extended commit messages to use lower case. Eg, instead of "RESOLVES #123" make it "Resolves #123."

PR #680, opened 2017-12-15
[C] Switch manifold tld in client allowedHosts

SMaxOwok:


zdavis:
Yep. I did this in the branch I'm working on and keep forgetting to commit to master.

zdavis:
But rather than replace the .dev domain, could you add the .lvh domain to the array? That way it will be backwards compatible.

SMaxOwok:
updated

PR #694, opened 2017-12-28
[F] Add backend project social page

SMaxOwok:
This change should be accompanied by executing the runner https://gist.github.com/SMaxOwok/2fa10c8ac3af2f090f5a98dcc8301ed1 to reformat the project tweet fetch configs in the db.


zdavis:
Max, let's start a new rake task for updating existing installations of Manifold to v0.4.0. The task could be executed as `rails manifold:upgrade['v0.4.0']`. My search branch is going to require reindexing models, and it'd be nice to have a single point of execution to handle these kinds of cahnges. I think we'll have similar needs in the future.

Could you consult with @scryptmouse on how to architect this? I expect we'd want to wrap up each upgrade script in some kind of service object. We probably also want these to be idempotent, if at all possible. I envision these as a way to manipulate data to support changes in a new version of Manifold... I think we'll need to be able to interact with active record models, as you do in your gist, which we typically avoid in migrations.

SMaxOwok:
@zdavis I added an upgrade service object with the help of @scryptmouse.  With each version requiring updates we'll add a `SystemUpgrades::ManifoldXXXXXX` class that will contain the necessary changes for that version.  I added the v0.4.0 file in this branch.

api/app/services/system_upgrades/manifold000400.rb
```diff
@@ -0,0 +1,41 @@
+module SystemUpgrades
+  class Manifold000400 < ActiveInteraction::Base
+    object :logger
+
+    def execute
+      logger.info "Updating Project tweet_fetch_config"
+      reformat_tweet_fetch_config
+    end
+
+    def reformat_tweet_fetch_config
```
zdavis: Probably all methods other than `execute` should be private.

SMaxOwok: Damnit, I was going to do that.  Updating.

zdavis: :)

SMaxOwok:
@zdavis I think this is OK to be reviewed now.  The upgrade changes are in a separate branch/PR (#710).  In this branch, I've only added the script to the `v0.4.0` upgrade file.

zdavis:
@SMaxOwok For the three files with conflicts...

client/src/helpers/router/DrawerSwitch.js
client/src/helpers/router/Passthrough.js
client/src/helpers/router/childRoutes.js

I would simply accept what is in master, which I think is behaving as it should. I changed our approach to props passing and moved some of the matching logic into a separate HOC so that if we want another drawer-like component, we won't have to duplicate all that logic.

zdavis:
Some quick UX feedback:

- [ ] If you close without saving, z-index on dialog box is wrong
- [ ] Popularity switch isn't clear enough. I think I'd use a dropdown labeled "Query Type" with the options "most popular" and "most recent"
- [ ] We should include instructions below the query input that link to the documentation on twitter. Remove the "learn more" line form above the list and include it as an instruction on the query input. As it is, nobody will really know what to put in the "query" field
- [ ] Since you have a counter_cache, why not include the total number of events associated with the query somewhere? It could be in the list, or in the form
- [ ] It's odd to have "active" on the left and the query on the right in the list. I'd switch those.
- [ ] Put the query field before the active switch and the popular/recent toggle
- [ ] The list feels weird generally. Even adding some numbering or a small twitter icon to the left of the query might improve it. Thoughts?
- [ ] Deleting a query should delete the events associated with it. We should warn the user before doing this in the modal. If they want to keep those events, they can set the query to inactive.

zdavis:
RE: list style — use the people list and project/authors as guidance:
image
image

Use the query as the label on the left. Where it says "reader" in the people list, you could include active/inactive. We should strive to introduce as little visual inconsistency as possible.

zdavis:
Thank you for all your work on this. I'm accepting this PR! Woot! 🍾 🎉

PR #711, opened 2018-01-03
[C] Rearrange google integration keys

SMaxOwok:


SMaxOwok:
Going to change a few other things

SMaxOwok:
This will be in the google config file upload branch

PR #717, opened 2018-01-06
[F] Set Google service credentials from file

SMaxOwok:
Will load from a .json file either as upload in backend or stored on filesystem with path specified in `.env`


zdavis:
Nice work on this. Looks good to me.

PR #728, opened 2018-01-10
[F] Include managed Elasticsearch service

zdavis:


zdavis:
Hey @SMaxOwok — could you test this out and let me know if it works OK for you?

zdavis:
You should be able to pull this, then run `m start`. Manifold won't find Elasticsearch, and it should give you instructions on what to do. Follow them and let me know how it goes.

PR #729, opened 2018-01-10
[B] Fix project rake task argument references

### LINKED ISSUE #730: Project rake tasks fail with incorrect arg call

SMaxOwok:
Some of these were failing because they were missing the `args` reference.  Adjusted the others so the style matches.

Fixes #730


zdavis:
Could you please create an issue for this and reference it from the commit message? The issue will ensure that it shows up in the changelog.

PR #732, opened 2018-01-10
[B] Fix drive resource importing

### LINKED ISSUE #731: Drive resource import failing
Looks like the importer wasn't updated to support the metadata attributes.

SMaxOwok:
Fixes #731

### DIFF
--- a/api/app/models/resource.rb
+++ b/api/app/models/resource.rb
@@ -30,7 +30,7 @@ class Resource < ApplicationRecord
     series_title container_title isbn issn doi original_publisher
     original_publisher_place original_title publisher publisher_place version
     series_number edition issue volume rights rights_territory restrictions rights_holder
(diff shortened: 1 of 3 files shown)


SMaxOwok:
I only added the rows that are currently in the resource sheet to the `RESOURCE_METADATA_ATTRIBUTES`.  Adding the remaining resource model metadata attributes throws an error (https://github.com/ManifoldScholar/manifold/blob/master/api/app/services/importer/drive_resources.rb#L282). 

If desired, I can add the remaining ones and adjust the sheet or refactor the resource sheet method to not error on a missing column.

zdavis:
Seems to me that the importer shouldn't error if the column is missing.

zdavis:
@SMaxOwok tests failing.

PR #739, opened 2018-01-22
[B] Fail with error if google doc is unavailable

### LINKED ISSUE #715: Throw an error message when someone tries to upload a private google doc
In my test, the text just wasn't added to the project after it was uploaded.

zdavis:
Fixes #715


zdavis:
Can be accepted once CI passes.

PR #742, opened 2018-01-23
[F] Add popup menu for annotations on a link

SMaxOwok:


SMaxOwok:
Still working on this, but am open to suggestions on continuing to further refactor the `Popup.Wrapper`.  It feels like a ton of props are dumped into the wrapper from `Annotatable`.  I was thinking maybe `Annotatable` could specify the child component that `Popup.Wrapper` renders (similar to the drawer).  That way it could be more specific about what props are needed to pass through.

SMaxOwok:
[docs.google.com/document/d/1PKMbOgnoVrmZ8_FZiCqSGeyDCDrqbjDvQ3HCzbgpdmU](Google doc with links in text)

zdavis:
Got some feedback on this...

- [ ] Sizing on the arrow is not correct. The width of all the icons should match. The arrow is narrower than the text balloon, which creates too much space between text and icon for the first row.
- [ ] Instead of "View Link" let's say "Follow Link"
- [ ] If I click on "View Annotation," the popup should disappear. Currently, it persists.

Otherwise, this looks good to me.

One little thing... I get the sense that there's some general refactoring and some new functionality. It would make it easier for me to review if, when possible, you broke these two things into smaller commits. It's hard to easily see what's new and what's refactoring to make the new stuff easier. No big deal, in any case.

PR #758, opened 2018-01-29
[F] Improvements to User/Maker

### LINKED ISSUE #754: Improve management of maker records
Backend users should be able to edit a maker from the lists of authors and collaborators. I think the comps may even show an "edit" link that accomplishes that. It seems to me that we could open the record in a drawer right from the project or text view.

Let's also add a middle name and suffix field to the maker record and expose it in the edit maker form.

SMaxOwok:
This includes a commit that refactors the name parsing into a concern using the Namae gem.  This ends up being somewhat of a lateral move, as the gem does not parse middle names individually.  It does, however, parse suffixes.

I can submit this PR without the parsing refactor, but it may be beneficial if we ever have other user-ish classes.

Closes #754


api/Gemfile
```diff
@@ -33,6 +33,7 @@ gem "kaminari", "~> 1.0"
 gem "memoist", "~> 0.16.0"
 gem "metadown", "~> 1.0"
 gem "money", "~> 6.0"
+gem "namae"
```
zdavis: Please set version to "~> 1.0"

zdavis:
Could you include a spec or two to illustrate what's happening in this PR?

PR #760, opened 2018-01-30
[F] Provide Manifold docker images

zdavis:


zdavis:
Hey @rclarkburns —I've opened a PR for your branch so we have a place to review and discuss while it's in progress.

rclarkburns:
Great, thanks! I'm curious if the named volumes that get created by that little bash wrapper script still end up mounted correctly on both of the containers. The idea that you would create the volumes, build the images, and then the containers is strictly for POC. In reality, we will upload the images to docker hub and docker compose will orchestrate the setup.

zdavis:
I'm still setting up the containers and will let you know how it goes.

Let's move docker-setup to the bin directory instead of putting it in the root.

I think the next thing to tackle is that the domain name. Right now you have it hard coded, for POC purposes. But, that's something that we'll need to be able to set when the containers start up. I'm not sure what the best way to do this is in docker, but this might be a good starting point: https://serverfault.com/questions/577370/how-can-i-use-environment-variables-in-nginx-conf

zdavis:
Should we close this PR? Does the work in ManifoldScholar/manifold-docker replace this?

PR #765, opened 2018-01-30
[F] Citable updates dependent citations after save

### LINKED ISSUE #709: Some texts fail to generate citations properly
- Texts don't have citation formats specified.
- Text isn't being successfully copied to the clipboard.

SMaxOwok:
Closes #709


zdavis:
Looks great.

PR #767, opened 2018-01-30
[B] Fix non-human name parsing

### LINKED ISSUE #766: Maker creation sometimes fails during text ingestion
With the changes introduced in https://github.com/ManifoldScholar/manifold/commit/43ce63e724d3e1c08e705d5f20fbcbd493f4db92, non-human names are breaking maker lookup in https://github.com/ManifoldScholar/manifold/blob/master/api/app/services/ingestor/creator/collaborators.rb#L50.

Currently the attributes are returned as:
`{:last_name=>"Moveable Type", :first_name=>"The League", nil=>"of"}`

Before they were returned as:
`{:first_name=>"The League of Moveable", :last_name=>"Type"}`

SMaxOwok:
This was an issue with non-human name parts causing errors on Maker create.  The Namae gem uses capital letters to delineate name parts and assigns mid-name lowercase words to `:particle`.  This PR maps `:particle` to `:middle_name`.  

The name parsing seems like it may be an ever-evolving issue, though I think we're still in a better starting position here than our previous splitting methods.

screen shot 2018-01-30 at 3 10 15 pm

Fixes #766

### DIFF
--- a/api/app/controllers/concerns/validation.rb
+++ b/api/app/controllers/concerns/validation.rb
@@ -237,7 +237,7 @@ module Validation
   def maker_params
     params.require(:data)
     attributes = [:first_name, :last_name, :middle_name, :suffix, attachment(:avatar),
-                  :remove_avatar]
+                  :remove_avatar, :name]
     relationships = [:users]
     param_config = structure_params(attributes: attributes, relationships: relationships)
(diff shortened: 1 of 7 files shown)


zdavis:
Looks good to me.

PR #782, opened 2018-02-02
[F] Implement roles through Rolify

SMaxOwok:


client/src/store/subscriptions/onUserIsCurrentUserUpdate.js
```diff
@@ -0,0 +1,28 @@
+import { requests, meAPI } from "api";
```
zdavis: I think it would be good to have a simple test for this subscriber. Look at the `redux-mock-store` module for this. You should be able to setup a mock store, then dispatch a fake (synchronous) API_RESPONSE action with a response similar to what we get back from a users API endpoint. Depending on what's in that action, you'd expect to see the subscriber called and the state of the current user in the store to update.

api/app/models/user.rb
```diff
@@ -99,8 +95,27 @@ def send_welcome_email
     AccountMailer.welcome(self, false).deliver
   end
 
+  def role=(new_role)
+    @pending_role = new_role.to_s
+  end
+
+  def role
```
zdavis: This doesn't seem right, but maybe I'm not understanding why it's here. Didn't we decide to treat this global, calculate role as `kind` rather than role? Seems confusing to give this two names.

api/app/serializers/user_serializer.rb
```diff
@@ -6,13 +6,13 @@ class UserSerializer < ActiveModel::Serializer
   attributes :id, :email, :nickname, :first_name, :last_name, :role, :created_at,
              :updated_at, :full_name, :avatar_styles, :is_current_user, :persistent_ui
 
+  has_many :favorites, serializer: FavoriteSerializer
```
zdavis: Why add this back in? Didn't you remove the favorites relation on users in an earlier commit? I don't think we need to serializer all a user's favorites unless that user is the current_user (which comes from the other serializer)

api/db/migrate/20170406200143_add_default_role_to_users.rb
```diff
@@ -1,6 +1,6 @@
 class AddDefaultRoleToUsers < ActiveRecord::Migration[5.0]
   def up
-    change_column :users, :role, :string, default: User::ROLE_READER
+    change_column :users, :role, :string, default: Role::ROLE_READER
     execute "UPDATE users SET role = 'reader' WHERE role IS NULL"
```
zdavis: I'm confused by this as well. With the introduction of Rollify, aren't user roles stored in the user_roles join table? If so, why do we also have a string field on the user model? This seems wrong—but if I'm misunderstanding, please explain.

api/db/migrate/20180202214000_migrate_user_roles_to_rolify.rb
```diff
@@ -0,0 +1,44 @@
...
+        INSERT INTO users_roles (user_id, role_id) (
+          SELECT u.id AS user_id, r.id AS role_id
+          FROM users u
+          INNER JOIN roles r ON r.name = u.role AND u.role IN ('admin', 'reader')
+            WHERE r.resource_id IS NULL AND r.resource_type IS NULL
+        )
+      SQL
+    end
+
+    remove_column :users, :role
```
zdavis: I see, it gets removed here. Should these three migrations be collapsed down into one so it's clearer what's happening?

api/app/serializers/concerns/abilities.rb
```diff
@@ -0,0 +1,68 @@
+# Includes serializer authorization
+module Abilities
```
zdavis: This is going to return results that I know will be false. For example, currently, an unauthenticated user can definitely `read` a comment, but this concern will return false for all abilities if the user isn't logged in (see `CommentAuthorizer.readable_by?`). I don't think returning false for an unauthenticated user is the right way to go... thoughts?

api/spec/serializers/concerns/abilities_spec.rb
```diff
@@ -0,0 +1,69 @@
+require "rails_helper"
```
zdavis: How about including a case for when there is no authenticated user?

client/package.json
```diff
@@ -165,6 +165,7 @@
     "postcss-scss": "^1.0.2",
     "prettier": "^1.5.2",
     "react-hot-loader": "3.0.0-beta.7",
+    "redux-mock-store": "^1.5.1",
```
zdavis: Remove this, since you ended up not using it.

client/src/components/frontend/Event/__tests__/List-test.js
```diff
@@ -11,9 +11,7 @@ const project = build.entity.project("1");
 describe("Frontend.Event.List Component", () => {
```
zdavis: It's not clear why this is part of this commit.

client/src/store/subscriptions/onUserIsCurrentUserUpdate.js
```diff
@@ -0,0 +1,22 @@
...
+  return () => {
+    lastState = currentState;
+    currentState = store.getState();
+    const currentUser = currentState.authentication.currentUser;
+    if (!currentUser) return;
+    const lastStateUser = lastState.entityStore.entities.users[currentUser.id];
+    const currentStateUser =
+      currentState.entityStore.entities.users[currentUser.id];
+    if (!lastStateUser || !currentStateUser) return;
+    if (isEqual(lastStateUser.attributes, currentStateUser.attributes)) return;
```
zdavis: isEqual performs a deep comparison of objects, which means it's a costly execution, and in this case, this deep comparison will happen every single time the store state changes (if the state contains a user model and an authenticated user), which is going to be often. Lodash's isEqual implementation does do an inexpensive shallow check first (see https://github.com/lodash/lodash/blob/4.17.5/lodash.js#L3280), but if it didn't, you'd definitely want that shallow check as a guard in your code.

PR #798, opened 2018-02-12
[B] Swap resource cube inline svgs for icon font

### LINKED ISSUE #636: Edge: Resource cube gets cut off
NOTE: This was seen without typekit fonts.  I confirmed that the behavior does not happen in other browsers without typekit, but I don't know if the IE bug is present with typekit.

virtualbox_msedge - win10_12_10_2017_10_48_40

naomiyaki:
Closes #636 

This replaces the long standing inline `&nbsp` with svg background with an icon-font glyph nested in an inline-flex span element. This seems to work great in all contexts and is a cleaner strategy overall that uses color transitions instead of swapping svgs. The icons look crisper and aren't cut off in Edge or Chrome.

### DIFF
--- a/client/src/components/reader/Notation/Marker.js
+++ b/client/src/components/reader/Notation/Marker.js
@@ -83,8 +83,8 @@ class NotationMarker extends Component {
                 if (!touch) this.setActiveAnnotation(null);
               }}
             >
-              <i className="manicon notation-cube-fill">
-                &nbsp;&nbsp;&nbsp;&nbsp;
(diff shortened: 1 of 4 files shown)


zdavis:
This approach introduces unacceptable whitespace.

This is how the markers render currently in master:
image

And this is how the markers render in nr/636:
image

naomiyaki:
@zdavis I wasn't able to replicate this issue, but I pushed an adjustment that led to better spacing on my end. Please have a look and let me know if this works for you!

Issue #434, opened 2017-07-12
Bulk uploading of resources


zdavis:

tsmyre:
Here's a link to the spreadsheet we had drafted up to begin this conversation: https://docs.google.com/spreadsheets/d/1d9hKa3CMQh6lvPCEqLqQtiRlgKurNBMK0CpCaiRjjzA/edit?usp=sharing

zdavis:
@tsmyre How important is the first row in this spreadsheet? I'd like to write an import mechanism that's not entirely dependent on google sheets. Ideally, I think whatever we write would work with .csv data as well. The sheet you linked to above includes merged cells in that top row, which I don't think are really supported in csv. It also includes an instructional row that the import mechanism will have to know about. I worry that the logic here—skip row one, grab column headers from row two, then skip row three and start importing in row four—is too tailored to this particular spreadsheet and will make it difficult for other presses to utilize this feature unless they have that exact structure to start from. 

I'd like to see us use a simpler structure. The first row could be headers, and the subsequent rows could contain the actual data. If we do this, we could easily allow users to upload from different formats (xml, json, etc.), rather than having the logic bound tightly to this particular spreadsheet implementation.

But I'm also sensitive to what you're trying to accomplish here, which is to make the data entry process easier for authors by providing them with instructions. Could we break that instructional content out into our documentation?

Let me know what you think.

--Zach

tsmyre:
@zdavis The first row was more for cosmetic organization than anything else, so I removed it. As you noted though it would be nice to have the instructions on the sheet (as well as in the documentation) for the sake of immediacy and easy. But they don't have to go below the column heads. I'm not sure if that helps with the logic, but I deleted what was the first row, replacing it with the instructions, so Row 1, Instructions; Row 2, Headers; Row 3, Data. I also removed all merged cells. Does that get at it, or does this require a bit more thinking through?

zdavis:
It's time to get started on this for the v0.4.0 release. First step is to get design for this feature from Lael. My attempt at a design brief for this feature is as follows:

**Summary**
[...]

PR #811, opened 2018-02-17
Add markup and interaction scripting for bulk CSV import

naomiyaki:
Will address #434


zdavis:
This work has been incorporated into zd/434, which is still in progress. Closing this PR.

PR #816, opened 2018-02-21
[F] Add permissions interface to backend

### LINKED ISSUE #791: Project owners and creators can add users to a project and set their roles
1. We have global roles (for now) of Admin, Editor, and Reader
 - Admins can CRUD all objects (CRUD = create, read, update delete).
 - Editors can create projects. They are _owners_ of projects that they create.
  - Readers can’t login to the backend.

2. We have “owner” and “author” roles on projects
  - Owners can fully CRUD projects and all child models on a project (texts, resources, resource collections, etc)
  - Authors do basic resource management on projects if it’s enabled (eventually we’ll add config for what authors can and can’t do, but don’t want to overcomplicate right now)
 - When users with the Author role on a project annotate the project, their annotations show up in the author color.

SMaxOwok:
@zdavis do you want these points in this PR?

> Authors do basic resource management on projects if it’s enabled (eventually we’ll add config for what authors can and can’t do, but don’t want to overcomplicate right now)

>When users with the Author role on a project annotate the project, their annotations show up in the author color.

This branch feels pretty contained around adding permissions interface rather than what the effects of the roles are.  I can address those other requirements in a new PR from this, unless you want theme here.

Let me know what you think.

Closes #791


zdavis:
I agree that we can tackle those points in a separate PR

SMaxOwok:
Moving back to `in-progress` to refactor `Filterable` concern.  We're limiting the backend permission user search to editors only and our search with filter is buggy in that situation.

SMaxOwok:
Should be ready to review again following #828 being merged.  

Or can be reviewed as-is with the knowledge that #828 will resolve the issue with permission user search not being limited to editors.

zdavis:
Rebased on master.

PR #839, opened 2018-03-16
[F] Add bulk resource import mechanism

zdavis:


zdavis:
I wish hound would respect my _actual_ rubocop rules. Shove it, Hound.

PR #841, opened 2018-03-19
[F] Update Webpack to v4

zdavis:


zdavis:
The Node upgrade has already made it into master. The other parts are still applicable.

PR #842, opened 2018-03-19
[F] Add global search to Manifold

zdavis:


naomiyaki:
I've made a push that completes markup and styles for all the search result types. A lot of it is coming from real data, but there are a few things missing that I'm logging below:

1. On project results, the `relationships.makers` are null, so a static list is used
2. On resources, the `relationships.project` is null, so the link logic is provisionally setup to check for a project and return `#` as the link if it doesn't exist.
3. Similarly, there's a project subtitle on resources that isn't currently showing because projects aren't available.
4. Resources can't be found by title.
5. There is a jest error:
1 bash 2018-03-20 16-58-28

SMaxOwok:
- [ ] Fix null relationships
- [ ] Resources can't be found by title
- [ ] Fix tests

PR #861, opened 2018-03-26
[B] Fix maker drawer redirect after destroy

### LINKED ISSUE #858: Deleting a maker in a drawer always redirects back to maker landing page
Steps to reproduce:

1. Go to project makers in the backend
2. Edit a maker, which opens the drawer.
3. Delete the maker

At this point you'll be redirected to the maker landing page, when instead the drawer should close and you should remain on the project page.

SMaxOwok:
Fixes #858


SMaxOwok:
I had to do the redirect in a function on the container, rather than through props because we don't have access to `this.props.history` in `defaultProps`.

zdavis:
Yeah, that makes sense. There's no instance scope in the static defaultProps property.

PR #864, opened 2018-03-26
 [B] Fix incorrect spacing on SearchableList without buttons

### LINKED ISSUE #860: Optional button implementation in searchableList leaves extra wrapper element
This is leading to the space above the first item in the list

image

SMaxOwok:
Fixes #860


zdavis:
Per slack, there's a logic error in the searchable list component. Lemme know when you've fixed it. Looks OK otherwise.

SMaxOwok:
@zdavis updated.

Also added an unrelated prettier fix that was blocking `m check client`

PR #866, opened 2018-03-26
[B] Fix redirect after stylesheet save

### LINKED ISSUE #857: Saving a text stylesheet redirects incorrectly

SMaxOwok:
Fixes #857


SMaxOwok:
I took a guess at what the "correct" redirect was meant to be based on what was there/the bug described.  If it's meant to be something like redirect to sheet if new, refresh without redirect if existing, let me know and I'll update this.

PR #877, opened 2018-03-27
 [C] Show/hide terms and conditions based on link presence

SMaxOwok:


SMaxOwok:
I went with a regular text input and do a check for url/manifold page when rendering.

I had started with a `PredictiveInput` that pulled page records.  To go in this direction instead, we'd probably need a new component that accepted a string or pulled an attribute from an entity in the predictive list.  I can go back and continue with this approach if it's preferable.

client/src/components/global/SignInUp/Create.js
```diff
@@ -84,6 +85,22 @@ class CreateContainer extends Component {
...
+  renderTermsAndConditions(props) {
+    const terms = props.settings.attributes.general.termsUrl;
+    if (!terms) return null;
+    const name = props.settings.attributes.general.installationName;
+    const absoluteUrl = /^https?:\/\//i;
+    const url = absoluteUrl.test(terms) ? terms : `/page/${terms}`;
+
+    return (
+      <p className="login-links">
+        {`By creating this account, you agree to ${name}'s `}
```
zdavis: Please use the possessivize function from utils/string helper.

zdavis:
Ready to accept after this minor revision.

PR #888, opened 2018-03-28
[F] Implement new maker interface

### LINKED ISSUE #882: New Maker Button in People Menu
For ease of use and parallel structuring, it would be nice to have an `Add a New Maker` button to the People/Maker section and be available to all users who have permissions to create a new Maker record in Projects/People.

SMaxOwok:
Closes #882

### DIFF
--- a/client/src/api/notifications.js
+++ b/client/src/api/notifications.js
@@ -41,9 +41,8 @@ export default {
     return {
       level: 0,
       heading: "New maker record created",
-      body: `${possessivize(
-        payload.data.attributes.fullName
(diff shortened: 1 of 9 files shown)


zdavis:
I didn't look closely, but I suspect that the JSX in the new component is the same as JSX that already exists in an edit component. I'd like to see this more DRY, without the duplication. Possible?

SMaxOwok:
Absolutely, will update

zdavis:
Typo (unnecessary possessive) in alert:

image

PR #905, opened 2018-03-29
 [B] Fix resource list filters and URL update

SMaxOwok:


client/src/components/frontend/Resource/TagList.js
```diff
@@ -34,14 +34,14 @@ export default class ResourceTagList extends Component {
...
-    if (!tag || !projectId) return null;
+  createTagLink(tag, projectSlug, index) {
+    if (!tag || !projectSlug) return null;
     return (
       <li key={`${tag}-${index}`}>
         <Link
           onClick={this.stopPropagation}
           className={this.props.disabledLinks ? "disabled" : null}
-          to={`/project/${projectId}/resources?tag=${tag.toLowerCase()}`}
+          to={`/project/${projectSlug}/resources?tag=${tag.toLowerCase()}`}
```
zdavis: This is a no-no — please use a link helper here.

SMaxOwok: Good call, missed that

SMaxOwok:
@zdavis The url updating and filtering are fixed in this branch. See #908 for more on the pagination params in the URL.  The behavior re: pagination in this branch is currently unchanged.  It started feeling like a larger refactor than should be undertaken at the moment.

zdavis:
Thank you, max. I’ll check it out early tomorrow and if I see anything that
needs fixing, I’ll probably just tackle it myself. Have a good evening.

PR #915, opened 2018-04-02
[B] Fix incorrect guard on Pages.New container

SMaxOwok:
Fixes #913

PR #921, opened 2018-04-02
[F] Allow links in formatted fields

### LINKED ISSUE #914: MD link support in metadata fields
It would be good to have the ability to create links in discursive metadata fields, like `Caption`, `Description`, etc. in line with standard MD syntax: `[anchor](target)`.

SMaxOwok:
Closes #914


SMaxOwok:
Note CI is failing on unrelated tests that are fixed in #915

zdavis:
Can you fix up the tests. The snapshots are rendering "undefined" as the text, which is wrong and actually points to a problem—if the property is undefined, we shouldn't be rendering it.

PR #923, opened 2018-04-02
 [B] Fix backend project search not matching makers

SMaxOwok:


SMaxOwok:
Please note 10229a5.  Querying projects (I didn't look at other record types) in the backend was returning large numbers of IDs to apply to the search scope, which was slowing the search to a crawl.  Making sure we only pull distinct IDs cut the number of IDs down from 14k to 10.  I was not able to determine the cause of the large number of IDs.

It's worth noting that the querying on the staging site did not seem to be affected by this (it was fast), but we replicated this behavior locally on several machines.

PR #926, opened 2018-04-03
[B] Fix broken external links in reader

### LINKED ISSUE #916: HTML Ingestion does not handle external link correctly.
I created a new text (HTML ingestion) that contained a link in the title. The anchor appears as a link, but the [target is undefined](https://manifold.umn.edu/read/an-interview-with-professor-mareis-and-dr-allen-critical-media-lab/section/2421c365-8798-49f2-9da9-1dbc72d8bce8).

SMaxOwok:
Fixes #916

### DIFF
--- a/client/src/components/reader/Section/BodyNodes/LinkNode.js
+++ b/client/src/components/reader/Section/BodyNodes/LinkNode.js
@@ -19,15 +19,17 @@ class LinkNode extends Component {
     if (!this.hasUri()) {
       return false;
     }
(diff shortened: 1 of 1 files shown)


SMaxOwok:
This ended up being a frontend issue in how the `LinkNode`s are rendered.

`reference` is not a property on the object returned by `URI.parse()`, so we were always considering links as relative.  

This fix also makes the hash conditional, since not all of our relative links are to a hash anchor.  In those cases we were appending `#undefined` to routes.

PR #974, opened 2018-04-16
[C] Update to React 16.3.1

SMaxOwok:
There are a couple `UNSAFE_` functions left that require a more involved refactor, so I've adjusted the function name to have the prefix as well as tagging them with a `TODO: Zach ...`

I can clarify the approach to any of the changes, if necessary.

Also includes a refactor to use arrow functions to bind component functions.


SMaxOwok:
This branch will fail CI right now as it uses `setState` in `componentDidUpdate` often now.  These should be reviewed and ignored in lint if they should stay.

zdavis:
Max, I've accepted a bunch of PRs. Let's get this into master. I'm going to discus with you how best to do this.

SMaxOwok:
@zdavis I've updated this and I think it's ready now.

There are two `UNSAFE_` lifecycle methods left, which will need to be addressed before React 17.  There are also several places where `setState` is called in `componentDidMount`/`componentDidUpdate`, which throws a linting warning.  I've added a commit that disables these warnings because this seems to be an acceptable pattern when used with caution.

https://reactjs.org/docs/react-component.html#componentdidmount

PR #977, opened 2018-04-17
[F] Implement list ordering

### LINKED ISSUE #941: Sort users feature
## Feature Description
As Manifold installations grow and gain lots of users, it might be nice to be able to sort the list of users. I would like to be able to do that by role, so that, for instance, I could see everyone with an admin account together, but maybe also by name

It will help admins organize their installation and in particular will make it easier to change roles

## User Stories
An admin realizes that people with the role Project Creators should actually be classified as editors. Rather than going through the list and picking out names individually, if the admin could sort by role, it would make it easier to find everyone currently characterized as Project Creator

SMaxOwok:
Closes #941

### DIFF
--- a/api/app/controllers/concerns/validation.rb
+++ b/api/app/controllers/concerns/validation.rb
@@ -284,7 +284,7 @@ module Validation
   end
 
   def collection_filter_params
-    params.permit(filter: [:keyword])[:filter]
+    params.permit(filter: [:keyword, :order])[:filter]
   end
 
   def resource_filter_params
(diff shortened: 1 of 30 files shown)


SMaxOwok:
@naomiyaki please review the style changes.

zdavis:
Please double check this against the design on the issue. It's not clear to me that @laelito intended the sorting to be right-aligned and the filters to be left-aligned. Confirm this with him. Also, the implementation has a different link color on the reset search than we see in the comps. Is this intentional? If so, why.

SMaxOwok:
@zdavis, spoke with Lael and got the ordering/wrapping/spacing of the filters sorted out.  Also it looks like I was using an old comp for the reset button colors; I've adjusted them now.

PR #982, opened 2018-04-17
[B] Fix encoding on .csv imports

### LINKED ISSUE #963: Bulk Load CSV Error
When I save an Excel Spreadsheet as a CSV and try to use it in the bulk load process I get the following error.

I'll DM @zdavis with the spreadsheet and Drive folder ID on Slack.

screen shot 2018-04-12 at 1 11 22 pm

SMaxOwok:
The errors preventing bulk import can be fixed by specifying an encoding to `CSV.read`, but the output doesn't always translate the character correctly.  This is a little bit of a heavier solution, as it detects the input encoding, but the result is that all characters are maintained and converted to UTF-8.

**Specifying Encoding**
screen shot 2018-04-17 at 3 03 24 pm

**This Fix**
screen shot 2018-04-17 at 2 57 27 pm

Fixes #963


zdavis:
Really appreciate the inclusion of the test here — makes it much easier for me to accept a PR like this without manually testing it. Thanks!

PR #992, opened 2018-04-18
 [F] Add library placeholder when no instance projects

SMaxOwok:


zdavis:
If the user isn't logged in as an admin, can we display this language instead:

**Uh-oh. This Manifold Library is empty.**

But it’s easy to create and publish projects with Manifold. If you have an Admin account, _login to the backend_ to get started. For more help, you can learn about creating and publishing Manifold Projects _here_.

(italics indicate links)

PR #1007, opened 2018-04-25
[F] Improvements to bulk resource import

### LINKED ISSUE #944: Add a Skip field to Bulk Load Process
In the CSV bulk resource upload process, it would be great if the system could recognize a column labelled `Skip`, which would be a boolean means to prevent upload of that particular resource while still allowing space for information about it to be stored.

SMaxOwok:
Per @tsmyre's comments about how the "Special instructions" column is being used, individual statements in the column should be delimited with a semi-colon.  The import will then skip any row that is indicated as such in this column.

```
Special Instructions
Author wants this image to be set inline and not at the top or bottom of the page.; skip;
```

Closes #944

### DIFF
--- a/api/app/models/resource_import.rb
+++ b/api/app/models/resource_import.rb
@@ -51,7 +51,7 @@ class ResourceImport < ApplicationRecord
     %w(
       title kind sub_kind created_at caption description fingerprint external_url
       external_id external_type allow_download slug
-      minimum_width minimum_height tag_list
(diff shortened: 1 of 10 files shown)


SMaxOwok:
Gonna fold in the collection stuff here too.

SMaxOwok:
Collection column takes collection titles separated by `,` or `;`.

zdavis:
Appreciate the specs on this.

PR #1013, opened 2018-04-26
[B] Fix updater reordering

### LINKED ISSUE #1012: Category reordering isn't working
See #1009

SMaxOwok:
Fixes #1012


zdavis:
Would it be possible to include a test that shows the problem and the solution? It's kind of difficult for me to tell why these changes are being made without digging in more deeply. Tests really help me understand why a change is happening, and show me that the change works without me having to manually test it.

SMaxOwok:
@zdavis updated

PR #1026, opened 2018-05-09
[B] Allow HEAD <title> to be set by installation

### LINKED ISSUE #1024: Allow installation page title to be configurable
target `v1.0.x` branch

SMaxOwok:
The fix here has a few parts:
- Expose a `headTitle` attribute on settings.
- Pick that title up at the top level (`ManifoldContainer`) or use the default as specified in the config file.
- Set a `titleTemplate` prop for helmet by default.  The `titleTemplate` potentially also uses the title from settings, so it must be defined when helmet is first used (again in `ManifoldContainer`).  
- Expose a prop `nested` on the `HeadContent` component.  When included, the page will use the `titleTemplate`, which is defined at the higher scope.  When `nested` is not included as a prop, we pass `titleTemplate: false` to helmet to override the setting.

Please also review the language in the backend form.  I feel like it's a little unclear, but was struggling with naming/describing it.

Fixes #1024

### DIFF
--- a/api/app/controllers/concerns/validation.rb
+++ b/api/app/controllers/concerns/validation.rb
@@ -198,7 +198,8 @@ module Validation
           :social_share_message,
           :contact_url,
           :copyright,
-          :terms_url
+          :terms_url,
+          :head_title
         ]
       },
       {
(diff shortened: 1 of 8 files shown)


zdavis:
I changed the language, so snapshots need to be regenerated. Please change the `nested` prop to `appendTitle`

PR #1032, opened 2018-05-10
[B] Store emails in case-insensitive column

### LINKED ISSUE #1031: User accounts should not treat emails case sensitively
I'm dealing with a situation where someone has two different Manifold accounts because for one account, they capitalized a letter in the email address, and in the other, they didn't. I think that the convention in email is that they are not treated in a case-sensitive way; for Manifold, I would follow Gmail's conventions, where a punctuation mark like a period could be used to differentiate an account (e.g. username@gmail.com would be different from user.name@gmail.com) but capitalized letters would not result in a different account (e.g. Username@gmail.com would be treated as a duplicate of username@gmail.com)

SMaxOwok:
Fixes #1031

### DIFF
--- a/api/app/models/user.rb
+++ b/api/app/models/user.rb
@@ -35,7 +35,7 @@ class User < ApplicationRecord
   # Validation
   validates :password, length: { minimum: 8 }, allow_nil: true, confirmation: true
   validate :password_not_blank!
-  validates :email, presence: true
+  validates :email, presence: true, case_sensitive: false
   validates :email, uniqueness: true, email_format: { message: "is not valid" }
   validates :pending_role, inclusion: { in: Role::ALLOWED_ROLES }, allow_nil: true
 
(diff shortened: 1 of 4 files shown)


zdavis:
Max,

I'm afraid you need to do a little more work on this. As it stands, running this migration won't actually solve the problem. We have to deal with the invalid data that now exists because of the bug. For example, on the UMP Manifold installation, there are 2 users who have multiple accounts because of this bug. 

I think there are going to be very few instances of this problem in production, and I think deleting data here is unavoidable. Let's adjust the migration so that in the case of duplicates, we destroy all user accounts except for the last one that was created. Normally I don't like executing AR code in the migrations, but in this case I think we can safely rely on the `User` model being present always and forever. The advantage to executing this through ActiveModel is that callbacks will be called, and dependent records will be destroyed, which I think is necessary for data integrity purposes.

Please also open this PR against 1.x, as this is a pretty serious issue that we need to get fixed in prod envs as soon as possible.

SMaxOwok:
@zdavis, I adjusted the migration to remove duplicate records and keep the most recent.  Let me know if it needs any tweaks.

PR #1038, opened 2018-05-11
 [F] Attempt to reassign annotation start/end nodes when section changes

SMaxOwok:
The behavior of this change ends up being relatively small.  Effectively, the only addition coming from this PR is that annotations will detect their new start and end nodes when the UUIDs change.  The cases/outcomes of annotation reassignment is below.  Both of these cases has the same application when the section nodes are reordered.

- **Node content unchanged**: If the annotation subject remains stable, we use the existing start/end nodes.  If there are additional, new nodes between the start and end nodes, we orphan the annotation.

- **Node content changed**: We find the start and end nodes (new uuids because content has changed) and then see if the content spanning from start to end contains the annotation subject _exactly_.  We reassign to those nodes if it does or orphan if it does not.  If we encounter two nodes that match either the start or end nodes, we mark the annotation as orphaned because we cannot declare either the parent with any certainty.

Additionally on the frontend, we only fetch annotations that have not been orphaned in both the reader and the reader notes drawer.

### DIFF
--- a/api/app/controllers/concerns/validation.rb
+++ b/api/app/controllers/concerns/validation.rb
@@ -325,7 +325,9 @@ module Validation
   def annotation_filter_params
     coerce_filter_to_hash(:filter, :ids)
     coerce_filter_to_hash(:filter, :formats)
-    params.permit(filter: [{ ids: [] }, [{ formats: [] }], :text, :text_section])[:filter]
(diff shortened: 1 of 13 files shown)


zdavis:
@SMaxOwok Please rebase on master. Thanks.

PR #1053, opened 2018-05-17
Accessibility/1.1.1

blnkt:
This does not yet have any work exclusively for the reader so perhaps we'll end up drawing a distinction between backend/frontend vs reader?

This is in progress and not yet ready to be fully reviewed.


blnkt:
Closing in favor of breaking these commits out into their own PRs instead.

Issue #1056, opened 2018-05-17
Lint jsx files for Accessibility related issues


blnkt:
This includes the linter and any rules either so general as to be safe to include in front of more serious changes related to making Manifold more Accessible, or those rules specifically related to the implementation of ARIA attributes (as there are none present now and all future strategies for addressing Accessibility will involve ARIA)

zdavis:
Hey Blake... this has a "has pull request" label, but I'm not seeing it. Please reference the issue from the PR so that the relationship is more clear. Thanks.

blnkt:
Sorry.  I referenced the PR # instead of the issue # in the associated commit.  This was resolved with #1057.

zdavis:
Keep in mind you can always reference the issue in the PR body, which is
sometimes easier! I don’t care where the reference is, as long as I can
find it.

PR #1057, opened 2018-05-17
[F] Add jsx linter to address WCAG concerns

blnkt:
Includes the linter as well as currently applicable rules


blnkt:
Addresses #1056

PR #1058, opened 2018-05-17
[F] Toggle button elements work like real buttons in screen readers

### LINKED ISSUE #1045: Toggle switches should work with screen readers
There are a couple toggle switch style checkboxes that are not real form elements (they are divs with some javascript state changes instead of being checkboxes) and as such are unreachable settings for a screen reader

non-checkbox-2
non-checkbox

blnkt:
Resolves #1045

Adds a role and aria atttributes to treat toggle buttons (div elements) as clickable buttons by screen readers


blnkt:
Closing this pull request in favor of addressing this in the general concern of elements having role attributes befitting of their station: #1079

PR #1060, opened 2018-05-17
[F] Add descriptive alt text to all image tags and elements with background images

### LINKED ISSUE #1059: Images and elements with background images should have descriptive screen reader text
This does not include icons or other visual non-text elements.  This only ensures Image tags have valid and descriptive alt attributes and that background images for non-decorative elements have associated descriptive text

blnkt:
Resolves #1059

Includes updated Client Tests

### DIFF
--- a/client/.eslintrc
+++ b/client/.eslintrc
@@ -99,9 +99,9 @@
     "jsx-a11y/tabindex-no-positive": 2,
 
     // Rules for WCAG related to images
-    "jsx-a11y/accessible-emoji": 0,
-    "jsx-a11y/alt-text": 0,
-    "jsx-a11y/img-redundant-alt": 0,
+    "jsx-a11y/accessible-emoji": 2,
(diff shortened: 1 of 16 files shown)


zdavis:
Does this also capture places in jsx where we're using a backround image via CSS?

zdavis:
I did a quick search of img tags in the source. I found at least one place where the alt text is "". Does the linter not catch empty alt attributes? If so, maybe give it a second pass to make sure you've caught every case where there's missing or inaccurate alt-text.

blnkt:
Good catch @zdavis.  Blank alt attributes are allowed by the linter.  It is only checking to make sure img tags have a valid alt attribute with any value (even an empty string).  I've updated the only instance I found (the no follow animation demo).  Is this otherwise ready to pull in?

zdavis:
Yep

PR #1062, opened 2018-05-17
[F] Add alt text to non-text non-image elements

blnkt:
This mostly includes updates to interactive elements containing no text but only an icon or other non-text visual elements


zdavis:
- Typo in extended commit msg "understaidn". 
- Check for a place where you have an href attribute on a span tag.

blnkt:
I fixed that typo and removed that superfluous href attribute.

PR #1078, opened 2018-05-22
 [C] Refactor project authority abilities to be clearer

### LINKED ISSUE #999: Improve permissioned redirects in backend
Not all backend project routes redirect as expected

screen shot 2018-04-20 at 10 29 24 am

SMaxOwok:
Fixes #999


client/src/containers/backend/Content/Features/Detail.js
```diff
@@ -200,38 +201,47 @@ class FeatureDetailContainer extends PureComponent {
...
-                          Foreground, background, and markdown will not be
-                          displayed until the feature is saved.
-                        </span>
-                        <Layout.Splash feature={previewFeature} preview />
+      <HigherOrder.Authorize
+        entity="feature"
+        failureFatalError={{
+          detail: `You are not allowed to ${action} features.`
+        }}
+        ability={action}
```
zdavis: The action here can be either `update` or `create`. If we're updating, we should authorize against the instance, not the class ability. Something like this:

```
const authProps = isNew ? { entity: feature, action: "update" } : { entity: "feature", action: "create" }
return (
  <HigherOrder.Authorize {...authProps} otherProp="blah" />
```

api/app/authorizers/project_authorizer.rb
```diff
@@ -70,9 +68,65 @@ def makers_updatable_by?(user, _options = {})
...
+    socials.detect do |social|
+      __send__("#{social}_manageable_by?", user) ||
+        __send__("#{social}_creatable_by?", user)
+    end.present?
+  end
+
+  def events_manageable_by?(user, _options = {})
+    updatable_by?(user)
+  end
+
```
zdavis: Per slack, I'm refactoring this and will push a commit that DRYs this up.

client/src/containers/backend/Content/Pages/Detail.js
```diff
@@ -257,10 +260,19 @@ class PageDetailContainer extends PureComponent {
...
     const isNew = this.isNew(this.props);
+    const action = isNew ? "create" : "update";
 
-    if (isNew) return this.renderNew();
-    if (page) return this.renderExisting(page);
-    return null;
+    return (
+      <HigherOrder.Authorize
+        entity="feature"
+        failureFatalError={{
```
zdavis: Same issue as above with features. Need to authorize the instance in the case of update.

client/src/containers/backend/Permission/List.js
```diff
@@ -57,7 +57,7 @@ export class PermissionContainer extends PureComponent {
               }}
             />
           ) : null}
-          <HigherOrder.Authorize entity={entity} ability="updatePermissions">
+          <HigherOrder.Authorize entity={entity} ability="createPermissions">
```
zdavis: I wonder if we should wrap the List.SimpleList above with a check on manage permissions for the entity. Thoughts?

client/src/containers/backend/Project/Log.js
```diff
@@ -66,26 +68,35 @@ export class LogContainer extends PureComponent {
...
-        <Utility.Pagination
-          pagination={this.props.versionsMeta.pagination}
-          paginationPadding={3}
-          paginationClickHandler={this.pageChangeHandlerCreator}
-          paginationClass="secondary"
-        />
-      </React.Fragment>
+      <HigherOrder.Authorize
+        entity={project}
+        ability="update"
```
zdavis: Shouldn't this be authorizing against log_readable rather than update project?

client/src/containers/backend/Project/Resources.js
```diff
@@ -14,9 +16,16 @@ export default class ProjectResourcesContainer extends PureComponent {
     if (!project) return null;
 
     return (
-      <section>
-        <Resource.ResourcesList project={project} />
-      </section>
+      <HigherOrder.Authorize
+        entity={project}
+        ability={["manageResources"]}
```
zdavis: This should probably just be a string, no?

zdavis:
I reviewed this and made a few changes. I kind of re-worked c37438928210fe3cd0274cce98aa9ee929ea95cb so that it's more DRY, per our Slack conversation. I also added a commit onto the end of this branch that I think improves the whole top-level backend container redirect approach (see extended commit message). There are still a couple permission checks that seem wrong to me. Please make those changes, make sure CI passes, and we should be good to go.

client/src/containers/backend/Permission/List.js
SMaxOwok: The permissions container has a `managePermissions` ability check on it.  I think this is the right approach because it is validating a user has permission to see anything under the project permissions path.  This includes the list, but also edit, etc.  To me, I think think it makes sense to keep the "add" button check separate, but if not, it could be removed entirely as you cannot see the List without manage permissions.

SMaxOwok: https://github.com/ManifoldScholar/manifold/pull/1078/files#diff-a7c704a3d4c0102626ad9dece07eab24R23

PR #1093, opened 2018-05-29
[B] Fix authorization check on Page/Feature

### LINKED ISSUE #1092: Page/Feature edit authorization check is broken
Regression from #999.  Because the record fetch is asynchronous, the record is null when the authorization check happens.

SMaxOwok:
Fixes #1092


SMaxOwok:
I felt like checking the authProp was nicer than `if (!isNew && !page)`.  Happy to adjust if desired.

zdavis:
@SMaxOwok Can you eyeball the changes for #999 (if you haven't already) and see if there are similar regressions elsewhere in that PR? Otherwise, this looks good.

ZD

PR #1094, opened 2018-05-29
[B] Fix updater reloading model when errors were present

### LINKED ISSUE #1090: Objects are not returning errors on update
When sending invalid data to update a resource, the errors are not returned.  I just get back the unmodified object.

This appears to be a regression at some point.

SMaxOwok:
Fixes #1090

This was a regression from a3a91d3329b5c3549e6884ee6db4e5a11e962ba9.
The check `model.persisted?` was always true because at that point a
record has either been created or was already present. This then meant
the `#post_update` method was always called, which included a `model.reload`.
This fix uses `save_model`'s return to determine whether the `post_update`
method should be called.

### DIFF
--- a/api/app/lib/updaters.rb
+++ b/api/app/lib/updaters.rb
@@ -31,8 +31,7 @@ module Updaters
   def update(model, creator: nil)
     @model = model
     update_without_save(model, creator: creator)
-    save_model(model)
-    post_update(model) if model.persisted?
+    post_update(model) if save_model(model)
     model
   end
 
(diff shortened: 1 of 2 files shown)


zdavis:
Hey @SMaxOwok I added a spec to this—take a look at it. With bug fixes especially, adding a spec that  fails before the fix and passes after it is very helpful. It only takes a few minutes to add it, and doing so helps assure that the regression doesn't pop up again in the future. It also makes my review faster and easier, which I like :)

best,
Zach

PR #1095, opened 2018-05-29
[B] Fix "delete feature" button states

### LINKED ISSUE #1091: Delete feature button states are weird
I would expect the `X` to change state with the text.

feature

blnkt:
Other similar button arrangements don't have any text for the delete button (see screenshot) so I made that screen reader text instead.  I also cleaned up the ternary for toggling the move up and move down buttons' visibility.

Resolves #1091
screen shot 2018-05-29 at 3 39 32 pm

### DIFF
--- a/client/src/containers/backend/Content/Features/List.js
+++ b/client/src/containers/backend/Content/Features/List.js
@@ -1,5 +1,6 @@
 import React, { PureComponent } from "react";
 import PropTypes from "prop-types";
+import classNames from "classnames";
 import { Dialog } from "components/backend";
 import { Link } from "react-router-dom";
 import lh from "helpers/linkHandler";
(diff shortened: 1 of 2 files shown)


zdavis:
This PR seems to not solve the issue, and perhaps even introduce other problems. Here's why I'm rejecting this. Some of these issues may have already been present, but they need to be addressed as part of this.

1) The placement of the delete text in relation to the position arrows lacks proper margins:
image

2) The hover behavior of the X icon and the text are still not linked. These hover states should not operated independently of one another:
image

The correct fix here is to remove the "delete feature" text. There's no reason to label the delete button here when we don't do it anywhere else in the backend, so please remove that label. The layout of these rows should match what you see in the backend list of texts for a project.

zdavis:
I wasn't looking at the right branch; Blake had it right all along. Sorry!

PR #1100, opened 2018-05-31
All Label elements are used with associated form controls

### LINKED ISSUE #1088: Labels should only be used for form controls, and should be associated with a form element
- [x] Labels should only be used for form controls
- [x] Label should be associated with a form element implicitly or explicitly using layout (wrapping the form element in the label) or using the `for/htmlFor`, `aria-label`, or `aria-labelledby` attributes (`for` attr is preferred but likely all strategies will be necessary --depending on the application).

blnkt:
Resolves #1088
Related to #1101

Some Screen Readers have a "Form Mode" where each form element's description is drawn directly from its context.  If that form control does not have a properly associated label then that description might be drawn from non-associated text.  This PR ensures all labels for form controls currently implemented have the correct attributes to associate them with their form controls.  IT DOES NOT ensure that all form controls have labels.  Incidentally, for the sake of consistency, and in an effort to honor html elements' intrinsic uses, all labels not used to describe form controls were reformatted as h4s and restyled to be match their initial aesthetic as in almost all cases they were being used as a section heading.

In one case I could not find the react component in the frontend end so elected to leave its markup and styles as is until its whereabouts are better known: `client/src/components/reader/Annotation/Share/Citation.js`

### DIFF
--- a/client/.eslintrc
+++ b/client/.eslintrc
@@ -112,21 +112,31 @@
     "jsx-a11y/anchor-has-content": 2,
     "jsx-a11y/anchor-is-valid": 2,
 
+    // Rules for WCAG related to form labels
+    "jsx-a11y/label-has-for": [
+      2,
+      {
+        "components": [ "Label" ],
+        "required": {
+          "some": [ "nesting", "id" ]
+        },
+        "allowChildren": false
+      }
+    ],
+
     // Rules for WCAG we should eventually enable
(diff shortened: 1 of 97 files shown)


blnkt:
Related to #1101

zdavis:
@blnkt Please rebase on master to resolve the conflicts. Thanks!

PR #1113, opened 2018-06-06
[F] Redirect focus when UIPanels open/close

### LINKED ISSUE #1052: Manage focus redirection for UIPanels
UIPanels can exist elsewhere in the DOM (in relation to whatever interaction triggered its opening) so we need to programmatically shift focus to the UIPanel when it opens (otherwise a user using a screen reader would not be able to easily/consistently access the UIPanel).  Similarly, when the UIPanel closes, we need to redirect a screen reader user back to whatever element triggered the UIPanel to open (i.e. return them to wherever they started in the DOM tree when they opened the UIPanel).

Related to #1076

blnkt:
Resolves #1052

### DIFF
--- a/client/src/actions/helpers/commonActions.js
+++ b/client/src/actions/helpers/commonActions.js
@@ -28,7 +28,13 @@ function b(action, dispatch) {
 const commonActions = dispatch => {
   return {
     toggleSearchPanel: b(() => panelToggle("search"), dispatch),
+    hideSearchPanel: b(() => panelHide("search"), dispatch),
     toggleUserPanel: b(() => panelToggle("user"), dispatch),
+    hideUserPanel: b(() => panelHide("user"), dispatch),
+    hideAppearancePanel: b(() => panelHide("appearance"), dispatch),
+    hideVisibilityPanel: b(() => panelHide("visibility"), dispatch),
+    hideNotesPanel: b(() => panelHide("notes"), dispatch),
+    hideReaderReturnPanel: b(() => panelHide("readerReturn"), dispatch),
     toggleSignInUpOverlay: b(
       () => visibilityToggle("signInUpOverlay"),
       dispatch
--- a/client/src/actions/helpers/commonActions.js
+++ b/client/src/actions/helpers/commonActions.js
@@ -30,7 +30,6 @@ const commonActions = dispatch => {
     toggleSearchPanel: b(() => panelToggle("search"), dispatch),
     hideSearchPanel: b(() => panelHide("search"), dispatch),
     toggleUserPanel: b(() => panelToggle("user"), dispatch),
-    hideUserPanel: b(() => panelHide("user"), dispatch),
     hideAppearancePanel: b(() => panelHide("appearance"), dispatch),
     hideVisibilityPanel: b(() => panelHide("visibility"), dispatch),
     hideNotesPanel: b(() => panelHide("notes"), dispatch),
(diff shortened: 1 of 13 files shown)


zdavis:
I'm seeing at least one regression here in my first pass. With these changes, if you click the "menu" button in the reader while it's open, instead of closing it closes and immediately reopens. Any thoughts as to what's causing this?

zdavis:
There's a typo in your commit message. When this is ready for review again, please set the label accordingly.

blnkt:
@zdavis Ok.  This is ready for review again.  Sorry I didn't better anticipate the toggle's unresponsiveness.  I must have become so well acclimated to using the Esc key that I stopped toggling the panels using the toggle as intended.  The problem was two pronged:

1) By default, `focus-trap-react` only honors clicks inside its trap when its active.  There's an option I'm passing it which allows clicks outside of the trap to deactivate the trap.  And that works fine.  However...

2) In order to update the state (and thus close the panel) when the focus trap component detects one of these clicks outside of the trap, I used it's built in `onDeactivate` callback.  The problem with that is when it updates the state of the panel it doesn't know/care about something else also updating the state. So, when that click outside of the trap happens to be on a toggle that opens and closes the panel, the state is updated twice and we see the visibility briefly flash as the panel closes and reopens crazy fast.

The best way I could think of was to make our focus trap dumber and not deal with the panel's state at all.  Instead it is now only responsible for its own activation and deactivation (trapping focus or not) which in turn is dictated by the panel's visibility (state).  But, in order for the panel to be closable by someone using a screen reader (who is caught in the focus trap) I added a "screen reader only" close button to the panel.  Unfortunately this means we also lose the "Press Esc key to close" functionality as that was built into the Focus trap and was using the onDeactivation callback we eliminated.  This feels like a decent trade off to me.  If we like the "Press Esc key to close" functionality enough we should just implement it on the UIPanel component level instead of trying to hack around the Focus Trap component.  I tested every Panel I could find to make sure none of them already had close buttons in their bodies.  I also tested every toggle I could find to ensure they are all behaving as we would expect (i.e. opening and closing the panels without any to do).  I still think the Focus Trap is the best/right approach even if I am running into some of this library's limitations.  That said, the source does not look terribly complicated and if you think there is a dramatically better/cleaner approach/solution I think we could very well fork and adapt `focus-trap-react` to our specific wants/needs.

zdavis:
Ok, I'm seeing a different regression now. It seems to me that we're missing the UI panel open/close animation on all panels except for the contents panel. Let me know if you need me to brainstorm a solution with you on this.

blnkt:
@zdavis This PR has been updated to re-enable the existing animations of the panels sliding open/closed.  The problem was I was mounting/unmounting the panel when the visibility state was changing.  Instead I'm now just using the visibility to activate/deactivate the focus trap.  These panels still won't close on Esc press but they still include a screen reader only close button.  I tested the Reader panels that are activated by the items in the right-side of the header and they all work A-Ok.  

Let me know if you encounter any other issue with this feature.

zdavis:
Hey @blnkt please rebase on master and I'll review this. Thanks!

zdavis:
Nevermind, I did it!

zdavis:
Ready to accept this once tests pass.

blnkt:
@zdavis ok.  Tests updated and passing.  Sorry for the delay on this.  Merge whenever you have the brain-time.

PR #1114, opened 2018-06-06
[WIP] Refactor ingestion process

SMaxOwok:


api/app/services/ingestions/strategies/abstract_strategy.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+
+      # @return [Manifest]
+      def perform
+        raise NotImplementedError, "Must implement #{self.class}##{__method__}"
+      end
+
+      def inspector
+        @inspector ||=
+          "Ingestions::Strategy::#{strategy_name}".safe_constantize::Inspector.new context
```
scryptmouse: You want to be using `constantize` here, not `safe_constantize`. If this class does not exist we want a `NameError` to be raised.

Also, when dealing with abstract classes like this, since this is something that may get overridden by users, we want to provide this as a hook.

```ruby
# @api private
# @abstract
# @return [Class]
def inspector_klass
  "Ingestions::Strategy::#{strategy_name}::Inspector".constantize
end

def inspector
  @inspector ||= inspector_klass.new context
end
```

api/app/services/ingestions/context.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    def initialize(ingestion, logger = Rails.logger)
+      @ingestion = ingestion
+      @logger = logger
+      @user = current_user
+
+      initialize_file_operations
+    end
+
+    def initialize_file_operations
```
zdavis: Not loving this method name. It's not really initializing "file operations" so much as staging files in the source dir.

api/app/services/ingestions/context.rb
```diff
@@ -0,0 +1,32 @@
...
+      @source_path = ingestion.ingestion_source
+      @identifier = ingestion.id
```
zdavis: Seems like these should be set in the initialize method. The distinction between this method and initialize doesn't seem natural to me.

api/app/services/ingestions/fetchers/zip.rb
```diff
@@ -0,0 +1,7 @@
+module Ingestions
+  module Fetchers
+    class Zip < Ingestions::IngestionService
```
zdavis: Is the idea that this will replace the zip functions currently in the context?

api/app/services/ingestions/ingestion_service.rb
```diff
@@ -0,0 +1,24 @@
+module Ingestions
+  class IngestionService < ActiveInteraction::Base
```
zdavis: I think this is actually an abstract interaction class and that should be made clear in the class name. Am I right in reading this as a service that doesn't do anything?

api/app/services/ingestions/strategies/epub.rb
```diff
@@ -0,0 +1,100 @@
+module Ingestions
+  module Strategies
+    class Epub < Ingestions::Strategies::AbstractStrategy
+
+      def perform
+        compose Fetchers::URL, extension: "epub" if epub_url?
```
zdavis: This is wrong. Anything involved with fetching the source file should live outside a strategy. The ingestion source directory should be fully in place before the strategy begins it's work. See https://manifold-community.slack.com/files/U8KBJCT8B/FB2BCB08G/image.png

api/app/services/ingestions/strategies/epub.rb
```diff
@@ -0,0 +1,100 @@
...
+        manifest
+      end
+
+      def determine_ingestibility
```
zdavis: With the change that comes from my comment above, this becomes simpler too, since it's already been fetched. You'll be able to lose the epub_url? method.

PR #1116, opened 2018-06-07
Redirect Focus when Drawers open/close

### LINKED ISSUE #1112: Manage focus redirection for Drawers
Drawers can exist elsewhere in the DOM (in relation to whatever interaction triggered its opening) so we need to programmatically shift focus to the Drawer when it opens (otherwise a user using a screen reader would not be able to easily/consistently access the Drawer). Similarly, when the Drawer closes, we need to redirect a screen reader user back to whatever element triggered the Drawer to open (i.e. return them to wherever they started in the DOM tree when they opened the Drawer).

Related to #1076

blnkt:
Resolves #1112


blnkt:
@zdavis The Tests for this PR are broken.  I believe I've isolated what part of my change is causing things to break but I think I need some direction from you or @SMaxOwok in terms of how to go about troubleshooting the tests.

zdavis:

zdavis:
Passing this back to Blake to style focus states. Discussed at https://manifold-community.slack.com/archives/C9TBE7AHF/p1532365902000374

blnkt:
This PR now includes updated focus styles for all the interactive elements I could find.  When design wasn't provided I used the hover state style as the focus state style

PR #1118, opened 2018-06-07
Redirect Focus when Dialog boxes open/close

### LINKED ISSUE #1117: Manage focus redirection for Dialog boxes
Dialog boxes can exist elsewhere in the DOM (in relation to whatever interaction triggered its opening) so we need to programmatically shift focus to the Dialog when it opens (otherwise a user using a screen reader would not be able to easily/consistently access the Dialog). Similarly, when the Dialog closes, we need to redirect a screen reader user back to whatever element triggered the Dialog to open (i.e. return them to wherever they started in the DOM tree when they opened the Dialog).

Related to #1076

blnkt:
Resolves #1117

### DIFF
--- a/client/src/components/backend/Dialog/Wrapper.js
+++ b/client/src/components/backend/Dialog/Wrapper.js
@@ -4,6 +4,7 @@ import PropTypes from "prop-types";
 import { CSSTransitionGroup as ReactCSSTransitionGroup } from "react-transition-group";
 import { withRouter } from "react-router-dom";
 import classnames from "classnames";
+import FocusTrap from "focus-trap-react";
 import isString from "lodash/isString";
 
 class DialogWrapper extends PureComponent {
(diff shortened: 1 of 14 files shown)


zdavis:
This isn't ready to merge because the problem of focus states is not really solved. I've added a route to the frontend in e3fbe94 that's available at /sandbox. This includes a sample of all the buttons I could find, although it's probably not exhaustive. When these buttons attain focus, the visual state should change to indicate that, which it does not currently. A lazy option would be to just apply hover states on focus, and that would be better than what we have now. 

I've asked @laelito to give us some guidance on button focus states, and am putting this back to in progress until we get that. Another thing I'm seeing on my sandbox page is that the hover states for some of these buttons seem incorrect or missing. We may as well tackle that in this branch.

I also see an inconsistency in button classnames that I'd like to address. We have a class called "button-secondary-dull" although in almost every case we use dull as a separate classname (eg, "button-secondary dull". Can we remove this inconsistency?

blnkt:
Buttons styles were updated in this PR instead: https://github.com/ManifoldScholar/manifold/pull/1116

zdavis:
Can be accepted once tests pass.

Issue #1119, opened 2018-06-07
All interactive elements with non-semantic roles need tab index to be focused on


blnkt:
Elements, like those with `role="button"` and `role="link"`, need an explicit tabIndex attribute if they are to be focused upon using only the tab key.

Related to #1076

Issue #1111, opened 2018-06-06
Manage focus redirection for Overlays


blnkt:
Overlays can exist elsewhere in the DOM (in relation to whatever interaction triggered its opening) so we need to programmatically shift focus to the Overlay when it opens (otherwise a user using a screen reader would not be able to easily/consistently access the Overlay). Similarly, when the Overlay closes, we need to redirect a screen reader user back to whatever element triggered the Overlay to open (i.e. return them to wherever they started in the DOM tree when they opened the Overlay).

Related to #1076

PR #1121, opened 2018-06-07
Updates elements to be accessible when tabbing through page content

### LINKED ISSUE #1076: Using only the tab key the user should be able to reach all interactive elements
This includes form elements, links, buttons, and any other elements behaving as such.  This work does not include those sets of components updated in dedicated commits (see related issues below).

Related to #1052, #1111, #1112, #1117, #1119

blnkt:
This does not include updates to components updated in #1052, #1111, #1112, #1117, #1119

Resolves #1076


zdavis:
Ready to accept once conflicts are resolved and CI passes.

PR #1140, opened 2018-06-18
[B] Fix resource slug validation when importing from google sheet

SMaxOwok:
The fix I ended up using here just resets the slug is its updated value is `nil`.  This will then trigger friendly to regenerate the slug.


zdavis:
I'm pretty sure that this still isn't correct. Validating presence is not what we want—a model with a nil slug should in fact be a valid model that we can save, since friendly ID will generate the slug. The validation should only mark the model as invalid if the slug is an empty string. Nil should continue to pass validation.

SMaxOwok:
Oh, of course.  Will fix ASAP.

PR #1163, opened 2018-07-03
[F] Refactor ingestion process

### LINKED ISSUE #1085: Refactor ingestions to be of two types: single file and multi file
This will let us use one index file format for both google doc and markdown ingestion.  

I believe we could use this for an html ingestion as well and might be able to have one structure generation style for EPUBs and one for everything else.

SMaxOwok:
Closes #1085


SMaxOwok:
Note: Commits are displaying in wrong order above.

zdavis:

SMaxOwok:
Recovered lost commits.

PR #1174, opened 2018-07-09
[B] Fix audio player height when not in slideshow

### LINKED ISSUE #1168: Audio resource hero doesn't need to be so tall
The audio player hero seems larger than it should be. Currently, it's getting 52vh from .resource-slide-figure. That extra height is fine if we're looking at the player in a slideshow. However, on the resource detail view (or on the reader resource overlay), it looks much better if the 52vh isn't applied to slides containing the audio player:

image

SMaxOwok:
Fixes #1168


zdavis:
Max, did you test this on the IEs?

SMaxOwok:
Checked in Edge and IE11.  Height looks ok.  Player looks wonky in IE11.  Marked in #1180

PR #1177, opened 2018-07-09
[C] Remove remaining outdated react lifecycle calls

### LINKED ISSUE #1064: Address jsx linter errors introduced after react 16.3 update
screen shot 2018-05-18 at 9 39 57 am

SMaxOwok:
Closes #1064


zdavis:
I'm concerned that this solution is not really adequate and will introduce problems. The decision update state rather than an instance variable wasn't an arbitrary decision. The `Manifold.bootstrap` method call returns a promise, and only after that promise is resolve do we update the application state with `bootstrapped: true`. By updating state, we could be guaranteed that components would re-render once the bootstrap was complete.

I don't see how your adjustment here can make that same guarantee. When the promise is resolved, you update this.bootstrapped, but how does that in ensure that the component (and its children) will re-render?

SMaxOwok:
I agree with that and did note the use of state for bootstrapped.  I can't remember what issue I was running into using state. It might have just been that it means using `setState` in `componentDidMount`, but as I think on it more, I believe React confirms that pattern is OK.

I'll adjust this a little more to use state again.

SMaxOwok:
@zdavis I've updated this to use state for `bootstrapped`.  A side-effect is that because `getDerivedStateFromProps` is static, I don't have access to `this`, so `store` is also stored in state now.

Is the `Html` component expected to only update once? It seems like unless the app is bootstrapped before mounting, SSR will be missing on all subsequent renders.  Let me know if you want to talk this piece through a little more.

This update should now be 1:1 with the existing behavior though.

zdavis:
This still does not look correct to me. The bootstrap promise callbacks return new state, but what happens with that return, given that the `getDerivedStateFromProps` will have already returned null before the promise has been resolved. Perhaps I should take a stab at this? I'm very wary of changes to this bootstrapping process, which took some time to arrive at, introducing unintended consequences.

The SSR is only relevant for the first request, so I'm not sure I understand your question. The initial request returns the SSR (which includes the Html component), and then the client application is mounted into that. Every subsequent request happens client-side, so I don't really follow your claim that the "SSR will be missing on all subsequent renders."

SMaxOwok:
@zdavis, I think your comment about SSR addresses what I was wondering.

I've updated the PR to use the bootstrapping method if the App isn't bootstrapped when it's mounted, then the promise callback will update the state.  I get what you're saying about the state updating before the promise is returned and it's callback executed.  

I think the way it is now address that issue, but please do take a pass over it if you think it's still inadequate.

zdavis:

zdavis:
Planning on accepting after #1222, which should fix the console warnings.

PR #1182, opened 2018-07-10
[B] Implement default publisher/place behavior

### LINKED ISSUE #1161: Default Publisher and Place Behavior
Entering values in the `Default Publisher` and `Default Place of Publication` fields in the Settings/General menu don't have the expected effect of auto-populating the respective fields when a new project is created. It's not clear what, if any, change is made by entering values in these fields, presently.

SMaxOwok:
Fixes #1161

### DIFF
--- a/api/app/controllers/concerns/validation.rb
+++ b/api/app/controllers/concerns/validation.rb
@@ -197,7 +197,7 @@ module Validation
         general: [
           :installation_name,
           :default_publisher,
-          :default_place_of_publication,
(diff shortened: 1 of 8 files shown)


SMaxOwok:
I set things up this way in case we want more defaults in the future.  `Settings` could have a column `defaults` and the project method could easily be made into a concern and applied to other models.

Let me know if it's overkill and I can simplify a bit.

PR #1184, opened 2018-07-10
[F] Add variant download list to resource detail page

### LINKED ISSUE #1067: Variant fields don't save resource
On the demo site I created a Video resource (YouTube) and tried uploading an MP4 as a variant. It seemed to save, but when I nav away, Manifold asks if I want to save. But it never actually does. Each time I mean to leave, I get the same prompt. If I refresh the variant I tried loading disappears.

SMaxOwok:
Closes #1067


zdavis:
Could you include a screenshot on this PR of what this looks like?

SMaxOwok:
This is what came from a brief discussion with @laelito 

screen shot 2018-07-11 at 9 28 54 am
screen shot 2018-07-11 at 9 29 35 am

zdavis:
Looks great. Thanks!

PR #1185, opened 2018-07-11
[B] Overlay de-focus should not trigger close

### LINKED ISSUE #1183: Visit resource page button is not working
This must be a regression.  Clicking on the "View resource page" button in the reader overlay just closes the overlay.

zdavis:
We recently wrapped our overlays in the FocusTrap component as part of
our accessibility efforts. As part of that implementation, we adjusted
the overlay so that it automatically triggers its close handler when
it loses focus. This is a problem because the close handler generally
will trigger a "back" action in history, or will redirect the user to
the URL from which the overlay was opened. The unintended side effect of
this is that if there is a link in the overlay (as in the case of the
resource notation overlay in the reader), clicking that link will
trigger the close behavior rather than going to the intended link
target.

Fixes #1183

### DIFF
--- a/client/src/components/global/Overlay.js
+++ b/client/src/components/global/Overlay.js
@@ -25,10 +25,18 @@ export default class Overlay extends Component {
     };
   }
 
+  componentDidMount() {
+    window.addEventListener("keyup", this.handleEscape);
+  }
+
   componentDidUpdate(prevProps) {
     if (prevProps.triggerScrollToTop !== this.props.triggerScrollToTop) {
-      this.scrollableEl.scrollTo(0, 0);
+      this.scrollableEl.scrollTop = 0;
     }
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener("keyup", this.handleEscape);
   }
 
   handleCloseEvent = event => {
@@ -39,6 +47,12 @@ export default class Overlay extends Component {
       setTimeout(() => {
         this.props.history.push(this.props.closeUrl);
       }, 200);
+    }
+  };
+
+  handleEscape = event => {
+    if (event.keyCode === 27) {
+      this.handleCloseEvent(event);
     }
   };
 
--- a/client/src/components/global/Overlay.js
+++ b/client/src/components/global/Overlay.js
@@ -92,7 +92,11 @@ export default class Overlay extends Component {
             this.scrollableEl = el;
           }}
         >
-          <FocusTrap>
+          <FocusTrap
+            focusTrapOptions={{
+              escapeDeactivates: false
+            }}
+          >
             {this.renderHeader(this.props)}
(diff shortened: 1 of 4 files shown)


zdavis:
@blnkt when you're back at work tomorrow, please weigh in on this change. I don't think triggering overlay close on deactivation is viable. Can you explain why that was necessary in your changes that introduced the focus trap? What does that accomplish?

zdavis:
(force pushed with updated snapshots)

blnkt:
@zdavis I tied the deactivation of the focus trap to the handleCloseEvent because one default feature of the focus trap wrapper is that you can use Esc to deactivate it.  I used this in other situations and liked getting the functionality of Esc key to close "for free."  And, in fact, in other situations this was the best way to allow screen readers to close something when otherwise a mouse was required (i.e. no close button).  When I was working on these overlays my thinking was the same.  And in some cases this triggering of the "back"  action in the history is how overlays close.  Unfortunately, I did not catch this unintended consequence and it surely is bad, bad, bad.  

However, I don't think only removing that deactivation callback will suffice.  As is, the user can deactivate the focus trap using Esc but the overlay doesn't close so the screen reader user is then navigating outside overlay and getting lost and all sorts of blegh.  An example of this can be seen in the search overlay in the reader:

overlay-deactivation

In this screenshot you can see the black rectangle (representing screen reader focus) trapped in the overlay.  Then I hit Esc and you can see it proceeds through the overlay items and then out into the page at large.

So I'd like to amend a commit to the PR that disables that "esc key to deactivate" functionality.  Not being able to exit on Esc shouldn't be a prob since, as far as I have seen, overlays always have close buttons.   Adding that option would look like this:

screen shot 2018-07-11 at 10 20 57 am

Thank you so much for catching this so promptly.  I'm sorry I didn't catch it in my initial testing.  Let me know if I can add a commit to this PR or if you'd like me to add this change in a separate PR or whathaveyou.  

thanks Zach!!

blnkt:
I went ahead and added the above as a commit to this branch because why not.

zdavis:
Right, that makes sense. I think that we _should_ implement escape to close the overlay, I just think that we should implement that in the overlay rather than in the focus trap.

blnkt:
Adding this Esc Handling now

zdavis:
@SMaxOwok or @blnkt could you please rebase this on master for me so I can accept it? Once this is in and test test pass, I think it will fix the other failing CI builds on PRs.

blnkt:
on it

blnkt:
@zdavis this is rebased and ready to merge.  I did notice that your commit ended up in there twice (660b8169519512892276f23453d6a49477597ae7) which is a shame but I guess git thought they weren't quite the same enough for one to overwrite the other?

Issue #972, opened 2018-04-13
Async component load is causing some issues with Slideshow.Caption


SMaxOwok:
SSR mismatch on collection page, messed up HTML nesting

Missing button on reader collection

SMaxOwok:
More context for this bug -- `Slideshow.Caption` does not have a re-render triggered when `Velocity` is loaded.  The functionality is in place and works, but the refs that are children of the `VelocityComponent` are still undefined.  This messes up our check to apply the expandable classes where necessary.

This fixes the issue, but is mostly a hack/workaround, so I think we should look more closely at properly resolving this.

```
const Velocity = Loadable({
  loader: () =>
    import(/* webpackChunkName: "velocity-react" */ "velocity-react").then(
      velocity => velocity.VelocityComponent
    ),
  render(Component, props) {
    if (!props.reference.state.velocityLoaded) {
      props.reference.setState({ velocityLoaded: true });
    }
    return <Component {...props} />
  },
  loading: () => null
});

class ResourceSlideCaption extends Component {
  render() {
    const animation = {
      animation: {
        height: this.state.expanded ? this.state.targetHeight : "3em"
      },
      duration: 250,
      complete: () => {
        if (this.state.expanded) {
          this._description.style.height = "auto";
        }
      }
    };

    return (
      <Velocity {...animation} reference={this} >
        {stuff}
      </Velocity>
    );
  }
}
```

SMaxOwok:
@zdavis thoughts?

zdavis:
I don't really have insight as to how to best fix this without digging into it. If what you've proposed is the best solution you can come up with, and it addresses the issue, I think you should proceed with it. Perhaps include a comment explaining why it's less than ideal should we circle back to it.

PR #1197, opened 2018-07-12
 [C] Replace velocity-react with react-collapse

SMaxOwok:
We're only using `velocity-react` in two places and the same effect can be easily achieved with `react-collapse`.  This PR came as a result of trying to solve #972.  

We rely on the height of the children of `<VelocityComponent>` to determine when a description can be expanded.  Since we accomplish this by using refs on the children, the are undefined until the async component is loaded.  When component loads asynchronously, there is no re-render triggered, so we never get those child refs.

Using `react-collapse` we achieve the same behavior, but with a much smaller package, allowing us to remove the `velocity-react` dependency entirely.  This allow allows us to not have to use code-splitting.

In it's current release, `react-collapse` uses `react-motion` for animation, but this dependency will be removed and be purely CSS based when version 5 is released.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -103,6 +103,7 @@
     "react-ace": "^5.9.0",
     "react-beautiful-dnd": "^4.0.1",
     "react-dom": "^16.4",
+    "react-collapse": "^4.0.3",
     "react-dropzone": "^4.x",
     "react-ga": "^2.x",
     "react-helmet": "^5.1.3",
@@ -110,6 +111,7 @@
     "react-html5video": "^2.x",
     "react-json-tree": "^0.x",
     "react-loadable": "^5.3.1",
+    "react-motion": "^0.5.2",
     "react-redux": "^5.0.5",
     "react-router": "^4.x",
     "react-router-config": "^1.0.0-beta.4",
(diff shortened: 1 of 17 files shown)


zdavis:
I'm OK with replacing velocity with react-collapse. However, the caption isn't behaving quite right. When I open the description, part of the text continues to be faded out:
image

SMaxOwok:
I wonder if this is currently behaving like you expect?  If I try to enter that `!bootstrapped` block currently in `componentWillReceiveProps`, I get an error that `document` isn't defined.

zdavis:
Did you try carefully to reproduce it? I'm pretty sure that it's a legitimate problem, because I also spent some time code splitting out react-collapse and react-motion to see what size those libraries were, and this was happening with that branch and that code. When I have time, I will look again, just want to be sure you gave it a thorough test.

SMaxOwok:
Yeah, I gave it a good test around.  Are there any steps to replicate besides just opening up the collection in the reader?

caption

zdavis:
I added the caption to a resource, put the resource in a collection, and saw the issue on the slideshow on the collection detail page.

SMaxOwok:
Got it now, thanks.

zdavis:
Can be accepted once CI passes.

PR #1230, opened 2018-07-20
[C] Make audio player error message clearer

### LINKED ISSUE #1226: Adjust audio player error state message
When an audio file can't be played by Manifold, the player presently surfaces an error that reads `Error Decoding Audiobuffer`. That may not be particularly revelatory to users. Instead something like `This audio file is not playable in your browser. Click the download button to listen to the file on your device` may be more useful for readers and provides them with an actionable next step if they want to access the file in question.

SMaxOwok:
Fixes #1226


zdavis:
Rebased on master and force pushed. Ready to accept when CI is complete.

PR #1253, opened 2018-07-26
[F] Add preprocessing to ingestion process before compilation

SMaxOwok:


zdavis:
Unless I read this wrong, applying this update would effectively remove styles from all texts ingested in a prior version of Manifold, because those texts would have no section to stylesheet associations. Please improve your migrations to address this. In versions of Manifold prior to this change, every section effectively displays every stylesheet. By associating each stylesheet with each section in the migration, you would maintain the existing behavior. Make sense?

PR #1261, opened 2018-07-30
Address follow interaction not properly setting state

### LINKED ISSUE #1246: Follow interaction not properly setting state
I was terribly afraid I'd introduced something to break this when I was updating the Follow component to be Web Accessible, so I diligently and methodically stepped through a goodly umber of commits and have narrowed down the issue to this commit: https://github.com/ManifoldScholar/manifold/commit/f17fcc7ab632358f52e65bed3bf32a9ab1545de6

It seems fair to surmise that this issue is related to the react 16.4 upgrade, but I don't immediately see how.  What I can see is that if I log the `view` value as it gets fed into the `setView` method it does accurately reflect the interaction I am doing (hover or click).  However the state never changes once it is set to unfollow and the props coming in are always such that the `getDerivedStateFromProps` seems to always steam roll the state to "unfollow" regardless of anything else.  I'm happy to pick this apart but I suspect @SMaxOwok or @zdavis are going to be better equipped to troubleshoot this issue as I think you both have a better understanding of what changed between 16.3 and 16.4 and how those changes might relate to this issue.

zdavis:
Resolves #1246


zdavis:
Can be accepted once tests pass.

PR #1262, opened 2018-07-30
[B] Blur navigation elements on route change

zdavis:


zdavis:
This probably isn't a good solution, actually. While it fixes nav behavior, it breaks drawer close behavior, where we actually do want the focus to be on the element that triggered opening the drawer. Sigh.

PR #1264, opened 2018-07-31
[F] Add support for larger ingestions via tus.io

### LINKED ISSUE #918: Manifold fails to ingest large EPUBS
There are a couple of issues at play here. With a 60MB EPUB on a VM without a lot of memory/resources, ingestion of a large file will fail. This happens because:

1) the Javascript FileReader API is not really meant to deal with large files and often crashes the browser process when you get up to 100/200MB files.

2) Paperclip chokes because it has to store the large base64 encoded file in memory.

Eventually, we'll move toward chunked uploads. As an interim solution, I think our best bet is to post the files with a multipart request.

scryptmouse:
Resolves #918


zdavis:
I'm working on refactoring the upload components; setting this back to in progress.

zdavis:
I believe this is ready to accept once CI is complete.

PR #1267, opened 2018-07-31
[B] Use redux store to track backend project list filter state

### LINKED ISSUE #1266: Keep backend project list state in store instead of instance variable

SMaxOwok:
Fixes #1266


SMaxOwok:
Note: As discussed, the state is now reset when the backend is left, _not_ when navigating away from dashboard/projects.

PR #1269, opened 2018-07-31
[F] Improve button focus and hover state styles

### LINKED ISSUE #1257: Elements do not lose focus after event completion
focus

blnkt:
Contributes to #1257  

Includes updated "subtle" focus styles for buttons.
Includes updated dotted-outline focus styles for navigational links.
Includes updated hover styles for buttons (conditional on background color of parent element).
Updated hover and focus styles for some types of "dull" buttons against dark background colors.

### DIFF
--- a/client/src/theme/Components/backend/_panel.scss
+++ b/client/src/theme/Components/backend/_panel.scss
@@ -58,15 +58,22 @@
         color: $neutral20;
       }
 
-      &.active {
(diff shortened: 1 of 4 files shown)


zdavis:
I'm accepting this PR, but leaving #1257 open, as there are still a number of small focus-related issues that we need to address somehow. Will update that story.

PR #1270, opened 2018-07-31
[B] Self-close void element tags in html validator

SMaxOwok:
The primary issue here was that the source html had unlcosed `<source>` tags.  While this is valid HTML 5, Nokogiri adds closing `</source>` tags, which is invalid.  Additionally, Nokogiri adds these closing tags in inconsistent places, resulting in some invalid nesting.

This fix adds a step in the HTML validator that will take void elements and ensure they're closed while maintaining their properties.  It will also remove any closing void tags.  It runs as the last step in the validator because we're manipulating the resulting HTML string, not the parsed nodes, and sending the string back through Nokogiri will add back the removed closing tags.

### DIFF
--- a/api/app/services/validator/html.rb
+++ b/api/app/services/validator/html.rb
@@ -4,6 +4,10 @@ module Validator
   # frontend. This mainly involves insuring proper nesting, and making sure that the
   # structure will work with ReactDom.
   class Html
+
+    VOID_ELEMENTS = %w(area base br col embed hr img input keygen
+                       link meta param source track wbr).freeze
+
     def validate(html)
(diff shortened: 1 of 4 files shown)


zdavis:
Why not address all void tags in this?

https://developer.mozilla.org/en-US/docs/Glossary/Empty_element

SMaxOwok:
@zdavis updated

scryptmouse:
Rather than using regexes to correct void elements, we might consider using an HTML5 parser extension for nokogiri:

https://github.com/rubys/nokogumbo

zdavis:
I like Alexa's suggestion. Max, can you take a look at that approach?

zdavis:
I'm going to accept this PR in the meantime, however, but would like to explore the HTML5 parser.

PR #1278, opened 2018-08-01
[B] Fix ToC drawer and trigger behavior

### LINKED ISSUE #1271: Regression in contents button behavior
image

There are some issues with the contents button that we need to address:

- If the contents are open, clicking the contents button should close it. This is no longer the case.
- The contents button maintains focus after selecting a different section. This is not desirable behavior, IMO.
- The focus state of this button is likely too close to the hover or active state.

### LINKED ISSUE #1275: Drawer closes at incorrect times
drawer

It seems like the drawer is closing when it is clicked outside of.  In these cases, the drawer should stay open and populate with the new content.

Note: In gif, the last click is a long click down, then release.

blnkt:
Resolves #1271
Resolves #1275

Updates the focus trap behavior to allow the contents button to toggle the ToC when clicked (even if the ToC is open).
Clears focus when user selects a different section.
Updates the focus state styles to be more subtle (ToC button and Menu button in the reader).
Updates the focus trap to allow interactive items outside of it to be clicked.

### DIFF
--- a/client/src/components/global/Drawer/Wrapper.js
+++ b/client/src/components/global/Drawer/Wrapper.js
@@ -32,7 +32,8 @@ export default class DrawerWrapper extends PureComponent {
     entrySide: PropTypes.string,
     style: PropTypes.string,
     history: PropTypes.object,
-    includeDrawerFrontMatter: PropTypes.bool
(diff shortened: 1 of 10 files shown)


zdavis:
Accept once CI completes

PR #1299, opened 2018-08-09
[F] Apply backend UI design updates for v2.0

zdavis:


zdavis:
- [ ] Show 20 projects per page in all projects grid.
- [ ] Alignment of new twitter query button placement seems off. See twitter_new_button_alignment.png. Should buttons align with the group headers, or with the form fields? Perhaps a question for Lael.
- [ ] Layout of project twitter queries list is too narrow. See twitter_query_list_width.png
- [ ] Saving a new project permission throws exception with "backendProjectProjectPermission is not a valid link handler."
- [ ] Pressing escape in the backend or frontend opens the user dropdown.
- [ ] Backend text reingest view (http://manifold.lvh/backend/projects/text/[:id]/ingestions/new) has unnecessary padding on left side, causing too much space between tertiary nav and reingest content. See reingest_padding.png. Maybe the fix is to add a header?
- [ ] Layout of "step 1" on resource bulk upload is broken (http://manifold.lvh/backend/projects/[:id]/resource-import). See resource_import_step_1_broken.png
- [ ] Too much padding on step 3 of bulk resource import (to test import, use api/spec/data/resource_import/resources.csv, header row value of "2" and Drive ID of 0B6UzSqdJo8o8WU40WHFTeWFPUW8), or perhaps it's intentional. See resource_import_step_3_padding.png
- [ ] With font size reductions, the resource import results seem too big. Perhaps bump down a few px? See resource_import_results.png
- [ ] It looks like project thumbnails in all projects may be being stretched. See project_thumbnail_stretched.png — is the picture of me stretched?
- [ ] Need padding between log list and pagination UI on project log view. See log_padding.png
- [ ] In the record header, the comps put the icon before the text. Current implementation puts the icon after the text. See header_icons.png
- [ ] Must we show tertiary navigation when there is only one option? See unnecessary_tertiary_nav.png
- [ ] Instructional text on Page form should align with the inputs, I think. See instructional_text_alignment.png
- [ ] Default feature record now looks hella broken. See broken_feature.png
- [ ] Fields on the feature model (eg, color fields) would benefit from two col layout in places
- [ ] Keep your console open as you click around — I've now seen a missing "totalCount" (pagination related?) exception lead to a white screen at least twice. I'm not sure how exactly to reproduce it yet though.
- [ ] Vertical centering of form headers is not correct. See form_header_vertical.png
- [ ] When not logged into the backend, user sees a box for backend link. See backend_link_box.png
- [ ] Hamburger nav icon much thinner than design shows. See hamburger_icon.png
- [ ] Placeholder text in mobile nav search input should be "Search...", not "Search for..."
- [ ] Login link in mobile nav should include an icon. See login_link_mobile_nav.png
- [ ] Clicking login link in mobile nav does not do anything.
- [ ] Clicking the current page in the mobile nav does not close the nav.
- [ ] Edit / View links in backend dashboard project list should not be visible. See mobile_dashboard_projects.png
- [ ] Mobile footer implementation does not match design. See mobile_footer.png
- [ ] Backend mobile dropdown nav lacks transition. Animating it would be nicer, if possible.
- [ ] Primary backend navigatino dropdowns are only triggered by clicking on the text, not on the arrow
- [ ] Primary backend navigation dropdowns lack transition. Simple animation would add polish.
- [ ] Primary frontend and backend navigation lacks hover states	
- [ ] Backend user lists shouldn't show role in mobile. See backend_user_list_mobile.png


[global_ui_screenshots.zip](https://github.com/ManifoldScholar/manifold/files/2275906/global_ui_screenshots.zip)

zdavis:
Drawers leave a gap beneath the new mobile header:

image

zdavis:
Too much top margin above model header in mobile:

image

zdavis:
Vertical alignment of subject headers in text list is no longer correct:

image

zdavis:
Notes to self: 
- UX around creating category in drawer is not ideal; post save behavior is wrong.
- Creating a new project permission seems to reload page, must be a JS error.

PR #1302, opened 2018-08-09
[B] Ingestion uploader should validate file ext instead of mime type

SMaxOwok:
This is a regression introduced with the change to Shrine.  

Also includes a change to the manifold config file.  Passing regex matchers are deprecated and will be removed in Shrine 3.


zdavis:
Accept once CI completes.

PR #1304, opened 2018-08-10
[B] Remove extraneous API call from Reader Container

### LINKED ISSUE #1283: Changing sections in the reader is causing the text_section API request to trigger twice
Need to look at the fetch data executions here.

zdavis:
The Section container will fetch the TextSection model for the Reader.

Fixes #1283.


zdavis:
Can accept once CI passes.

PR #1315, opened 2018-08-14
[F] Update Backend UI design for 2.0

zdavis:
This extensive set of changes 1) refactors a number of backend routes
into more predictable patterns, 2) addresses primary, secondary, and
tertiary navigation for mobile and desktop, and 3) resolves lingering
problems with RedirectToFirstChild, 4) introduces new form styles, and
5) generally refines backend typography and layout.


zdavis:
Current list of issues that need to be addressed:

- [ ] Saving a new project permission throws exception with "backendProjectProjectPermission is not a valid link handler."  [ZD]
- [ ] Pressing escape in the backend or frontend opens the user dropdown.  [ZD]
- [ ] Default feature record now looks hella broken. See broken_feature.png [DJ]
- [ ] Keep your console open as you click around — I've now seen a missing "totalCount" (pagination related?) exception lead to a white screen at least twice. I'm not sure how exactly to reproduce it yet though.  [ZD]
- [ ] Backend user lists shouldn't show role in mobile. See backend_user_list_mobile.png [DJ]
- [ ] Normalize top-level headers in backend containers [ZD]
- [ ] The height of drawers on mobile does not line up with the new header; there's a gap between the mobile header and the top of the drawer [DJ]
- [ ] The model header (eg, on a backend project page) has too much space between it and the mobile header [DJ]
- [ ] In the backend list of texts on a project, the subject headers no longer have correct vertical alignment within their container [DJ]
- [ ] UX around creating category in drawer is not ideal; post save behavior is wrong. [ZD]

PR #1319, opened 2018-08-15
[B] Fix EPUB v2 TOC tree generation

SMaxOwok:
The EPUB v2 node structure was selecting _all_ matching child nodes.  This meant that for every item with children, the entire branch was run through the structure generator.

The fix is to only select the direct descendants of each branch.


zdavis:
Waiting on tests before I accept.

PR #1324, opened 2018-08-16
[F] Implement notification preferences on Users

SMaxOwok:


zdavis:
Removed misc. snapshot update commit, moved it to master, and rebased this branch on Master.

SMaxOwok:
Fixed a failing test

zdavis:
Rebased on master.

api/app/controllers/api/v1/projects/relationships/events_controller.rb
```diff
@@ -10,7 +10,7 @@ class EventsController < ApplicationController
           resourceful! Event, authorize_options: { except: [:index] } do
             Event.filter(
               with_pagination!(event_filter_params),
-              scope: @project.events
+              scope: @project.events.excluding_type(%w(comment_created text_annotated))
```
zdavis: Ensure that there is a spec for this behavior:
EventsController does not return events of the type comment_created or text_annotated

api/app/models/annotation.rb
```diff
@@ -31,6 +31,10 @@ class Annotation < ApplicationRecord
   scope :by_formats, lambda { |formats|
     where(format: formats) if formats.present?
   }
+  scope :excluding_private, lambda { |creator|
```
zdavis: I added a FIXUP commit that removes this, as it does not appear to be used anywhere, and seems logically incomplete.

api/app/models/annotation.rb
```diff
@@ -58,6 +62,13 @@ class Annotation < ApplicationRecord
   belongs_to :collection, optional: true
   has_many :comments, as: :subject, dependent: :destroy, inverse_of: :subject,
                       counter_cache: :comments_count
+  has_one :text, through: :text_section
+  has_one :project, through: :text
+  has_one :annotation_created_event, -> { where event_type: Event::TEXT_ANNOTATED },
```
zdavis: Ensure that there is a spec for this behavior:
Annotation enqueues a Event::TEXT_ANNOTATED event on creation

api/app/models/annotation.rb
```diff
@@ -146,4 +160,11 @@ def public?
     !private
   end
 
+  private
+
+  def trigger_event_creation
+    return if %w(highlight resource).include? format
```
zdavis: These should not be string literals. You should instead be referencing the already defined constants. I've added a FIXUP commit that addresses this.

api/app/models/collection.rb
```diff
@@ -50,6 +55,9 @@ class Collection < ApplicationRecord
 
   scope :search_import, -> { includes(:project) }
 
+  # Callbacks
+  after_commit :trigger_creation_event, on: [:create]
```
zdavis: Ensure that there is a spec for this behavior:
Collection enqueues a Event:: COLLECTION_ADDED event on creation

api/app/models/comment.rb
```diff
@@ -28,9 +29,63 @@ class Comment < ApplicationRecord
...
+          dependent: :destroy,
+          inverse_of: :subject
+  delegate :project, to: :subject
 
   # Validations
   validates :body, :subject, presence: true
 
+  # Callbacks
+  after_create :notify_parent_creator!, :notify_project_editors!
+  after_commit :trigger_event_creation, on: [:create]
```
zdavis: Ensure that there is a spec for this behavior:
Comment enqueues a Event:: COMMENT_CREATED event on creation

api/app/models/comment.rb
```diff
@@ -28,9 +29,63 @@ class Comment < ApplicationRecord
...
+          as: :subject,
+          dependent: :destroy,
+          inverse_of: :subject
+  delegate :project, to: :subject
 
   # Validations
   validates :body, :subject, presence: true
 
+  # Callbacks
+  after_create :notify_parent_creator!, :notify_project_editors!
```
zdavis: Ensure that there are specs for both of these behaviors. We have assertions around account creation, for example, that assert that the mailer will be called when things happen. You could use that approach here.

api/app/models/comment.rb
```diff
@@ -28,9 +29,63 @@ class Comment < ApplicationRecord
...
+  after_commit :trigger_event_creation, on: [:create]
+
+  # rubocop:disable Metrics/LineLength, Metrics/AbcSize
+  def subject_url
```
zdavis: I don't love this here, but it's OK for now, as I think we do something similar in other models. At some point, we should move this URL generation into a helper class of some sort.

api/app/models/comment.rb
```diff
@@ -28,9 +29,63 @@ class Comment < ApplicationRecord
...
+  end
+  # rubocop:enable Metrics/LineLength, Metrics/AbcSize
+
+  private
+
+  def trigger_event_creation
+    Event.trigger(Event::COMMENT_CREATED, self)
+  end
+
+  def notify_parent_creator!
```
zdavis: If I understand it correctly, this implementation seems a little naive. Isn't it only notifying the immediate parent? Shouldn't all parents be notified, or perhaps even everyone who is involved in a thread.

Moreover, this model is the wrong place to be making decisions around whether the parent should be notified. It would expect this to kick off a job. In that job (or in a different service class), you can make decisions about who should be notified. It is not the concern of the comment model to figure this out. In other words, this `notify_parent_creator!` method should be a one-liner, not a method that includes logic.

api/app/models/comment.rb
```diff
@@ -28,9 +29,63 @@ class Comment < ApplicationRecord
...
+    return unless parent.present?
+    return if parent.creator == creator
+    return unless parent.creator.wants_notifications_for?("replies_to_me")
+
+    NotificationMailer.reply_notification(parent.creator, self).deliver
+  end
+
+  # rubocop:disable Metrics/AbcSize
+  def notify_project_editors!
```
zdavis: The same problem flagged above in `notify_parent_creator!` applies here. This method is doing work that it shouldn't be. Figuring out who gets notified is not a concern of the comment model. Put this logic in a job, or a service class.

api/app/models/comment.rb
zdavis: Wherever we land on deciding who gets notified, there should be a spec that clearly describes the notification logic.

api/app/models/concerns/notification_preferences.rb
```diff
@@ -0,0 +1,66 @@
+module Concerns
+  module NotificationPreferences
+    extend ActiveSupport::Concern
+
+    included do
+      has_many :notification_preferences, dependent: :destroy, autosave: true
+
+      scope :with_digest_for_frequency, lambda { |frequency|
```
zdavis: What these scopes are intended for is not very clear to me. Can the naming be improved here at all?

api/app/models/flag.rb
```diff
@@ -24,4 +24,13 @@ class Flag < ApplicationRecord
   # # Validations
   # validates :flaggable, presence: true
 
+  # Callbacks
+  after_create :notify_admin!
```
zdavis: Ensure that there is a spec for this behavior:
Flag enqueues an EnqueueFlaggedResourceJob job

api/app/models/notification_preference.rb
```diff
@@ -0,0 +1,41 @@
+class NotificationPreference < ApplicationRecord
+  # Constants
+  BOOLEAN_FREQUENCIES = %w(never always).freeze
```
zdavis: These strings should be captured in constants. You reference them elsewhere in code, and I'd prefer to see those replaced with NotificationPreference::NEVER, for example.

api/app/models/notification_preference.rb
```diff
@@ -0,0 +1,41 @@
...
+  DIGEST_FREQUENCIES = %w(never daily weekly).freeze
```
zdavis: Ibid.

api/app/models/notification_preference.rb
```diff
@@ -0,0 +1,41 @@
...
+  # Scopes
+  scope :by_kind, ->(kind) { where(kind: kind) }
+  scope :by_frequency, ->(frequency) { where(frequency: frequency) }
+
+  # Validations
+  validates :kind, :frequency, presence: true
+  validate :frequency_is_valid!
+
+  # Callbacks
+  before_create :default_followed_projects_to_always!
```
zdavis: Is there a spec for this? If not, there should be.

api/app/models/resource.rb
```diff
@@ -124,7 +124,7 @@ class Resource < ApplicationRecord
   before_validation :update_kind, :set_fingerprint!
   before_update :reset_stale_fields
   after_commit :queue_fetch_thumbnail, on: [:create, :update]
-  after_create :resource_to_event
+  after_commit :trigger_event_creation, on: [:create]
```
zdavis: I think I _did_ see a spec for this. Confirm—if not, add one.

api/app/services/notifications/unsubscribe.rb
```diff
@@ -0,0 +1,16 @@
+module Notifications
+  class Unsubscribe < ActiveInteraction::Base
+    string :token
+
+    validates :token, presence: true
+
+    def execute
+      user_id = UnsubscribeToken.verify(token).dig(:user_id)
+      user = User.find user_id
```
zdavis: What if we can't find a user here for some reason? Should we handle the record not found exception as well? Or add an error if there's no user?

api/app/views/notification_mailer/comment_notification.html.erb
```diff
@@ -0,0 +1,15 @@
+<%
+  if @comment.subject.is_a? Resource
+    title = @comment.subject.title.titleize
```
zdavis: I'm seeing a lot of Demeter violations in these mailer views that could lead to problems. I'm also seeing a lot of logic in them that should probably be elsewhere. Maybe touch based with @scryptmouse and see if she has suggestions for moving logic out of the views and addressing these Demeter violations.

api/app/views/notification_mailer/comment_notification.txt.erb
```diff
@@ -0,0 +1,17 @@
+<%
+  if @comment.subject.is_a? Resource
+    url = "#{@client_url}/projects/#{@comment.subject.project.slug}/resource/#{@comment.subject.slug}"
+    title = @comment.subject.title.titleize
```
zdavis: Demeter violation.

api/app/views/notification_mailer/comment_notification.txt.erb
```diff
@@ -0,0 +1,17 @@
+<%
+  if @comment.subject.is_a? Resource
```
zdavis: More logic than I'd like to see in an email template. Why are we building URL string here, and in other places building them in models? We need to be consistent.

api/app/views/notification_mailer/digest/_weekly.html.erb
```diff
@@ -0,0 +1,18 @@
+Here's your weekly <strong>Manifold</strong> activity update:
+
+<div>
+  <ul class="event-list">
+    <% @projects&.each do |project, events| %>
+      <li>
+        <% counted = events.group_by(&:subject_type).map { |type, list| pluralize(list.length, type) } %>
```
zdavis: Another example of more complex logic than I typically want to see in a view.

api/app/views/notification_mailer/digest/daily/_annotation_event.html.erb
```diff
@@ -0,0 +1,6 @@
+<% annotation = event.subject %>
+<% url = "#{@client_url}/read/#{annotation.text.slug}/section/#{annotation.text_section_id}#annotation-#{annotation.id}" %>
```
zdavis: We gotta get the URL building out of these views. If URLs change in the client, it's going to be more difficult than it should be to track these down. Build a service, or somehow centralize the URL generation instead of embedding this in the view layer.

api/app/views/notification_mailer/digest/daily/_collection_event.html.erb
```diff
@@ -0,0 +1,5 @@
+<% collection = event.subject %>
+
+<li>
+  <a href="<%= "#{@client_url}/collection/#{collection.slug}" %>"><%= collection.title %></a>
```
zdavis: See previous comments.

api/app/views/notification_mailer/digest/daily/_comment_event.html.erb
```diff
@@ -0,0 +1,14 @@
+<% comment = event.subject %>
+<% if comment.subject.is_a? Annotation %>
+  <% url = "#{@client_url}/read/#{comment.subject.text.slug}/section/#{comment.subject.text_section_id}#annotation-#{comment.subject.id}" %>
```
zdavis: See previous comments re: URL generation.

api/app/views/notification_mailer/digest/daily/_project_header.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="project-header">
+  <a href='<%= "#{@client_url}/project/#{project.slug}" %>'>
```
zdavis: See previous comments re: URL generation.

api/config/application.rb
```diff
@@ -75,13 +75,13 @@ class Application < Rails::Application
     # Skip views, helpers and assets when generating a new resource.
     config.api_only = true
 
-    config.eager_load_paths += [
+    config.eager_load_paths += %W(
```
zdavis: Why?

api/spec/requests/notification_preferences/relationships/unsubscribe_controller_spec.rb
```diff
@@ -0,0 +1,20 @@
+require "rails_helper"
+
+RSpec.describe "NotificationPreferences Unsubscribe API", type: :request do
+
+  include_context("authenticated request")
+  include_context("param helpers")
+  let(:user) { FactoryBot.create(:user) }
+  let(:token) { UnsubscribeToken.generate user }
+
+  describe "unsubscribes a user from notifications" do
```
zdavis: I'd like to see additional specs here for what happens when a token is invalid or the token is valid but the user not longer exists.

api/app/models/notification_preference.rb
scryptmouse: I would argue this and `DIGEST_FREQUENCIES` should be a classy_enum, not just a constant. It's the direction we want to move to, and these are brand new.

zdavis: Agreed, classy_enum is better

api/app/views/notification_mailer/comment_notification.html.erb
scryptmouse: For things like this, I suggest we add `draper` so we can be using decorators in the mailer views. Most of our frontend is handled by serializers, but I think for the mailer views, the separation between models and their presentational concerns that it offers is 100% better than complex view logic.

This logic could easily end up in a `CommentDecorator` that is encapsulated and easy to test with its individual methods. That could also be where the URL generation winds up, getting it out of the model but still easily accessible (with draper, any model can be decorated simply with `model.decorate`). Talk to me before implementing if we agree we should go this route, as I can offer some pointers.

Also, as a general rule, branching via `is_a?` / `kind_of?` should never end up in a view. That sort of logic should be in the model / decorator

api/app/models/notification_preference.rb
SMaxOwok: The values are, I'm not referencing the enum there.  Is that what you meant or are you saying the actual constants `BOOLEAN_FREQUENCIES` and `DIGEST_FREQUENCIES` should be enums too?

scryptmouse: The logic for what is valid for digest or boolean should be contained in the classy_enum (that's why they are classy!), and the validator should be reworked.

SMaxOwok: Oh, ok, I think I get it.  Thanks.

api/app/enums/notification_frequency.rb
```diff
@@ -0,0 +1,16 @@
+class NotificationFrequency < ClassyEnum::Base
```
scryptmouse: @SMaxOwok e.g. here you could modify this class:

```ruby
class NotificationFrequency < ClassyEnum::Base
  include ActiveSupport::Configurable

  config.nondigest = false
  config.digest = false

  def digest?
    config.digest.present?
  end

  def nondigest?
    config.nondigest.present?
  end
end

class NotificationFrequency::Never < NotificationFrequency
  config.digest = true
  config.nondigest = true
end

class NotificationFrequency::Always < NotificationFrequency
  config.nondigest = true
end

class NotificationFrequency::Daily < NotificationFrequency
  config.digest = true
end

# ... you get the idea

api/app/controllers/api/v1/projects/relationships/events_controller.rb
SMaxOwok: ✔

api/app/views/notification_mailer/comment_notification.html.erb
zdavis: I'm ok with using draper here.

api/app/models/annotation.rb
SMaxOwok: ✔

api/app/models/collection.rb
SMaxOwok: ✔

api/app/models/comment.rb
SMaxOwok: ✔

api/app/models/flag.rb
SMaxOwok: ✔

api/app/models/resource.rb
SMaxOwok: ✔

api/app/models/comment.rb
SMaxOwok: ✔

api/app/models/comment.rb
SMaxOwok: ✔

api/app/models/notification_preference.rb
SMaxOwok: ✔

api/app/models/notification_preference.rb
SMaxOwok: ✔

SMaxOwok:
One piece that @scryptmouse suggested that didn't make it in yet is to use classy_enums for comment subject kinds.  She wrote it out in https://gist.github.com/scryptmouse/b950f6b74b2e9ea68351387c078acff7.

zdavis:
@SMaxOwok When you get back from vacation, please review fb60a98 above carefully. This branch needed a lot of revision, and I want you to look carefully at what I caught, what I fixed, and what I changed. I'm going to leave this PR here for now and move all this work to another branch so I can rollup my commit for acceptance, while leaving it here for you to review when you're back.

PR #1327, opened 2018-08-21
[F] Add text download options to projects

SMaxOwok:


SMaxOwok:
screen shot 2018-08-21 at 8 41 30 am
screen shot 2018-08-21 at 8 41 34 am

zdavis:
This PR introduces a regression that makes it impossible to move a text from an uncategorized or categorized position into the published spot on a project. Please review and fix.

zdavis:
Too much space between cover and link when download is the only option:

image

I would expect the link to appear at the same position as the buy button when it's visible.

zdavis:
Discussed in slack. If the download link is the only thing below the cover, then we'll give it the bordered button treatment. Default text needs to change to "download ebook"

SMaxOwok:
screen shot 2018-08-23 at 12 34 33 pm
screen shot 2018-08-23 at 12 35 12 pm

PR #1338, opened 2018-08-23
[C] Enforce referential integrity for ingestions

scryptmouse:


api/db/migrate/20180725163014_enforce_ingestion_referential_integrity.rb
```diff
@@ -0,0 +1,69 @@
+class EnforceIngestionReferentialIntegrity < ActiveRecord::Migration[5.0]
+  def change
+    destroy_and_delete! "ingestions without a project" do
+      Ingestion.where project_id: nil
```
zdavis: I don't think this is going to fully address the problem. The issue is that some projects were deleted before we were careful to maintain the referential integrity of the ingestion project_id field. You can see this on the manifold.umn.edu instance:
```
[1] pry(main)> Ingestion.where(project_id: nil).count
=> 0
[2] pry(main)> Ingestion.where("project_id NOT IN (select DISTINCT(id) from projects)").count
=> 11
```
There are really 11 ingestions that should be deleted because the project no longer exists, but we won't catch those if we just look for nil project_ids.

scryptmouse:
We had a break, pushed a change and rebased against recent master changes

PR #1386, opened 2018-09-17
[C] Add NumberInput Form component

SMaxOwok:


zdavis:
This duplicates too much code with the textInput component. Extract shared code into a base component and then export a numberInput and textInput component that wrap the base component and adjust props to get the desired output.

PR #1387, opened 2018-09-17
[C] Refactor HasMany component

SMaxOwok:
This refactor breaks the `HasMany` form component into smaller sub-components.  Additionally, it introduces a new style when neither the `orderable` or `editClickHandler` props are passed through.

Will be necessary for project collection styles.

### DIFF
--- a/api/app/models/stylesheet.rb
+++ b/api/app/models/stylesheet.rb
@@ -17,7 +17,7 @@ class Stylesheet < ApplicationRecord
   has_many :text_sections,
            -> { order(position: :asc) },
            through: :text_section_stylesheets,
-           dependent: :nullify,
+           dependent: :destroy,
            inverse_of: :stylesheets
 
   # Validations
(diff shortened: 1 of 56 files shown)


SMaxOwok:
screen shot 2018-09-17 at 12 35 35 pm

SMaxOwok:
subjects
tags

zdavis:
Remove button is positioned a few pixels too high relative to the label:
image

zdavis:
Seeing errors with a resource that already existed prior to pulling this branch (and did not have any tags):

image

zdavis:
Need to change TagList prop validation to:
```
    value: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
```
(oneOfType instead of oneOf)

- Well should be min-height to size of one tag row.
- Improve vertical alignment in text input and in tag labels.

SMaxOwok:
Added new empty states to the bucket style.
screen shot 2018-09-18 at 3 58 02 pm
screen shot 2018-09-18 at 3 58 10 pm

zdavis:
A little concerned about code duplication between the list that the TagList produces and the non-orderable list that the HasMany produces. Asked Max to review and see if there are opportunities to reduce code repetition.

PR #1395, opened 2018-09-18
[B] Fix text section/stylesheet relationship behavior

### LINKED ISSUE #1393: Stylesheet/TextSection dependent behavior is wrong
Currently set to `dependent: :nullify`, but there is a non-null constraint on the relation columns.  This is halting the ingestion process when stylesheets change.

SMaxOwok:
Fixes #1393


zdavis:
Max to add a SS model spec asserting that deleting the SS does not destroy the associated TSs

PR #1408, opened 2018-09-20
[C] Refactor HasMany to set relationships on model

SMaxOwok:
This commit adjusts Form.HasMany to be wrapped with setter.
Along with this, HasMany will now set records to the relationships
object on a model if `name` prop is passed and `changeHandler` is
not passed.  If `changeHandler` is passed, HasMany will run the
callback function on select.

This allows the HasMany component to be used on a new
record as well as existing ones.


SMaxOwok:
Necessary for Project Collection work

PR #1427, opened 2018-09-25
[B] Don't italicize <em> tags in project hero subtitle

### LINKED ISSUE #1426: If project subtitle is italicized, any italics in the subtitle should be rendered as regular text
image

.project-info .subtitle em, in otherwords, should not be italicized.

SMaxOwok:
Fixes #1426


zdavis:
Should have specified—but can this go to 2.0.x?

PR #1431, opened 2018-09-26
[C] Add option to skip store response update

SMaxOwok:


SMaxOwok:
Necessary for Project Collection work

PR #1449, opened 2018-10-04
[B] Refactor serializer naming and usage

### LINKED ISSUE #1443: Page loads blank
I wanted to note this bug: on the Minnesota install -- manifold.umn.edu, I began by searching to find a book (search term: "Jussi"). I then found the book I wanted (the Anthrobescene) and clicked on the title in the search results. Twice, the page loaded blank. I was able to re-load and the page appeared on reload, but I wanted to see whether anyone else could reproduce the problem. 

I am on Chome on a Mac

Here is a link to the book itself -- https://manifold.umn.edu/projects/the-anthrobscene

SMaxOwok:
**Three notes:**

1. This will in fact fix #1443 as we are fully serializing a project in search results, but not including its associated records.  When navigating to the project from a search, the project is fetched async _and_ pulled from the store by id in the meantime.  Since it has been fully serialized and is in the store, its relationship props have a length, but no items.  This is the cause of the blank page error in the issue.

2. Note the change to https://github.com/ManifoldScholar/manifold/blob/master/api/app/controllers/concerns/api/v1/resourceful.rb#L158.  Our serializers aren't namespaced, so this helper was never returning anything.  This only became apparent when removing the `each_serializer` specifications from the index methods.

3. Create methods are not returning the full serialization.  I believe this is ok, as we almost almost redirect to a show view afterwards or the item created is in a list and only needs partial serialization.  I haven't found any issues, but am happy to change these to the full serializer if desired.


zdavis:
Please resolve conflicts. Thanks!

api/app/serializers/project_full_serializer.rb
```diff
@@ -0,0 +1,54 @@
+# Provides a partial serialization of a project model.
```
zdavis: Comment seems inaccurate

api/app/serializers/project_full_serializer.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  meta(partial: false)
+
+  attributes :event_count, :metadata, :collections_count, :resources_count,
+             :uncollected_resources_count, :published_text_toc_id, :event_types,
+             :metadata_properties, :citations, :hide_activity, :metadata_formatted,
+             :slug
+
+  link :self do
+    "https://manifold-api.dev/api/v1/projects/#{object.id}"
```
zdavis: This has probably been here forever, but we may as well fix it with our links service.

zdavis:
Looks good other than the conflict and the couple minor comments I made.

SMaxOwok:
@zdavis I changed the base branch to master... I wasn't sure where we landed on whether this should be in `v2.0.x` or `master`.

zdavis:
Please open this against v2.0.x. Thanks!

zdavis:
Also, while you're at it, just remove that link: self bit in the full project serializer. I don't think we're using that in the client, and it's not accurate, so we should get it outta there.

PR #1451, opened 2018-10-04
[C] Adjust all projects route

SMaxOwok:


zdavis:
Could you adjust the commit message so it's clearer why this change is necessary?

SMaxOwok:
@zdavis added commit message

PR #1477, opened 2018-10-12
Implement IconComposer utility component

dananjohnson:
This PR adds a global utility component, `IconComposer`, which takes an `icon` attribute and retrieves the appropriate SVG icon component from `components/global/SVG/Icon`. The idea behind this is to abstract the existing `Resource.IconComposer` component into a more general component, which will in turn be used to retrieve icons for individual Project Collections.

With this, this PR also replaces all instances of `Resource.IconComposer` on the backend and frontend with `Utility.IconComposer`. It also creates new components for the corresponding SVG icons. These icons feature updated SVG markup based on improved designs. Moreover, they form an initial effort to implement a component- and SVG-based icon system, which could eventually replace the existing font-based approach.


zdavis:
This looks pretty good to me... I did notice that you're introducing a dependency on `pascal-case`. We are already using a very similar node module called `humps`. Could you look at this and see if humps does what you need, so as to avoid adding another dependency?

(also need to fix a conflict in a snapshot)

PR #1483, opened 2018-10-15
[B] Catch and ignore URI.parse errors in reader

### LINKED ISSUE #1476: Invalid URIs should not cause fatal errors
https://github.com/ManifoldScholar/manifold/blob/master/client/src/components/reader/Section/BodyNodes/LinkNode.js#L26

A link like `https://doi:10.17953/aicr.37.3.01130638k210j380` is throwing an error like 
```
TYPEERROR: PORT "10.17953" IS NOT A VALID PORT
```

SMaxOwok:
If a URL fails to parse (most common occurrence would be an
outdated DOI link format), we will now just render the url as
an absolute, external URL.
Fixes #1476

### DIFF
--- a/client/src/components/reader/Section/BodyNodes/LinkNode.js
+++ b/client/src/components/reader/Section/BodyNodes/LinkNode.js
@@ -19,7 +19,12 @@ class LinkNode extends Component {
     if (!this.hasUri()) {
       return false;
     }
-    return URI.parse(this.props.attributes.href).protocol;
+
+    try {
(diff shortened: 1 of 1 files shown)


zdavis:
What about the URI parse call at line 31? Won't that still trigger this bug?

ZD

SMaxOwok:
No because `adjustedAttributes` will only be called for relative URLs, which we're determining by parsing the URI. If the URI won't parse, we call it absolute (L26), so it gets rendered as an external link.

PR #1484, opened 2018-10-16
[B] Fix secondary header link styles

### LINKED ISSUE #1385: Backend dashboard project link style improvements
screen shot 2018-09-14 at 10 56 12 am
The `Projects` link is a different color than the other headers on the dashboard.  Those don't link, but it feels weird that this one is a different color when the significance is the same.

projects-link
Header and count both link to the same place, but are not tied to each other.  They hover individually, which seems weird because they're in line and go to the same place.

SMaxOwok:
Fixes #1385


zdavis:
Thanks for fixing this. It was bugging me.

PR #1502, opened 2018-10-23
[B] HasMany reports attributes errors

SMaxOwok:


zdavis:
Please fix commit message to use imperative voice: "HasMany reports attribute errors"

PR #1552, opened 2018-10-26
[B] Improve open graph meta tags

### LINKED ISSUE #1518: Homepage opengraph tags are not referencing settings
We aren't using the values in settings at the top level (home and following are what I've seen so far).

SMaxOwok:
Fixes #1518

### DIFF
--- a/api/app/controllers/concerns/validation.rb
+++ b/api/app/controllers/concerns/validation.rb
@@ -229,6 +229,7 @@ module Validation
           :installation_name,
           :default_publisher,
           :default_publisher_place,
+          :head_description,
           :social_share_message,
           :contact_email,
           :copyright,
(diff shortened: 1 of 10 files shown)


zdavis:
Max is going to address og:site_name and add default description meta tag field, then this is ready to accept.

SMaxOwok:
Ok, I've addressed the notes you mentioned above, but also several other things that I think were missing.  I added in the Twitter site name based on a top-level Twitter account being set and also the top-level installation logo.

Getting these fallback values to cascade the correct way involved a little refactoring to `HeadContent`, so please review as if this was a new PR.  Do you see any ways that component could be made a little cleaner?

zdavis:
Seems to me that fatal error shouldn't be imported here.

SMaxOwok:
Rebased on `v2.0.x` to remove `FatalError` import.

PR #1570, opened 2018-10-30
[B] Don't rescue StandardError during ingestion

### LINKED ISSUE #1355: Refactored ingestion should not catch StandardError
The recent refactoring of the ingestion service is assigning any error raised during ingestion to the active interaction errors. This needs to be narrowed considerably so that it is only rescuing from errors that are reasonably raised by our ingestion classes.

### LINKED ISSUE #1430: Manifest ingestion fails without useful feedback when the manifest references a missing file
Manifest ingestion fails without useful feedback when the manifest references a missing file
## Expected Behavior
A report that you are missing a file

SMaxOwok:
This commit adds a new IngestionError class and raises
it in places where ingestions tend to fail.  We now rescue
this error instead of StandardError in the CatchesException
concern.

Fixes #1355
Fixes #1430

### DIFF
--- a/api/app/services/ingestions.rb
+++ b/api/app/services/ingestions.rb
@@ -1,4 +1,6 @@
 module Ingestions
+  class IngestionError < StandardError; end
+
   mattr_accessor :configuration do
     Ingestions::Configuration::GlobalConfigurator.new
   end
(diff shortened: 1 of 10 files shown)


SMaxOwok:
@zdavis let me know if you remember any other places where we've been seeing ingestions fail or you want covered explicitly.

PR #1574, opened 2018-10-31
[B] Allow text titles to be formatted

### LINKED ISSUE #1563: Markdown doesn't render in title fields
Markdown doesn't render as expected in the project title field nor in the the text title field:

project-title
<hr/>
text-title

It does render as expected in the resource title field, but having an asterisk as the first character seems to break the resource, serving [only a blank page](https://edge.manifoldapp.org/projects/this-is-a-title-within-a-title/resource/title-in-title).

The samples here come from a [dummy project on Edge](https://edge.manifoldapp.org/projects/this-is-a-title-within-a-title).

SMaxOwok:
Fixes #1563

### DIFF
--- a/api/app/models/text.rb
+++ b/api/app/models/text.rb
@@ -78,6 +78,8 @@ class Text < ApplicationRecord
   delegate :creator_names_array, to: :project, prefix: true, allow_nil: true
   delegate :publication_date, to: :project, prefix: true, allow_nil: true
   delegate :title, to: :category, prefix: true
+  delegate :title_formatted, to: :main_title, allow_nil: true
+  delegate :title_plaintext, to: :main_title, allow_nil: true
 
   # Validation
   validates :spine,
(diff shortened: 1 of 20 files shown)


api/app/models/text.rb
```diff
@@ -199,6 +205,21 @@ def to_s
 
   private
 
+  def title_changed?
+    reload_title
```
SMaxOwok: Are there any suggestions on how to do this that doesn't require reloading every time _or_ that can be a little smarter about when to reload?  The `title` method get the value from the main title association, which will have changed during `title=`.  

Right now, to compare the old value with the current I need to refresh that association.

SMaxOwok:
@zdavis updated to delegate formatted values to the main text title association

zdavis:
Snapshot failing here:
```
FAIL  src/containers/backend/Text/__tests__/Wrapper-test.js
  ● Backend Text Wrapper Container › renders correctly
    expect(value).toMatchSnapshot()
    
    Received value does not match stored snapshot 1.
    
    - Snapshot
    + Received
    
    @@ -14,11 +14,11 @@
              >
                <i
                  aria-hidden="true"
                  className="manicon manicon-arrow-left"
                />
    -           Rowan Test
    +           Back
              </a>
            </div>
          </nav>
          <div
            className="wrapper"
      42 |   });
      43 | 
    > 44 |   it("doesn't render to null" () => {
      45 |     let tree = component.toJSON();
      46 |     expect(tree).not.toBe(null);
      47 |   });
      
      at Object.<anonymous> (src/containers/backend/Text/__tests__/Wrapper-test.js:44:6)

```

SMaxOwok:
@zdavis updated

PR #1579, opened 2018-10-31
[B] Fix duplicate request on projects search

### LINKED ISSUE #1577: All projects route is making two requests on every search
screen shot 2018-10-31 at 12 31 12 pm
screen shot 2018-10-31 at 12 31 35 pm

SMaxOwok:
Fixes #1577

### DIFF
--- a/client/src/containers/frontend/Projects.js
+++ b/client/src/containers/frontend/Projects.js
@@ -73,9 +73,6 @@ export class ProjectsContainer extends Component {
     ) {
       this.props.fetchData(this.props);
     }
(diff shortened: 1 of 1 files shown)


zdavis:
Was this reload there for a reason? There is a utility to accurately compare locations in the location helper.

SMaxOwok:
I think originally the filter change handler was just updating the url and then that check was used to refetch when the url changed.  We changed the change handler in ce126c1d5d1bb561d27806965f32c5a512518cb2 to make the request and update the url, but didn't remove the location check.

PR #1584, opened 2018-11-01
[B] Correct some longstanding routing issues

zdavis:


SMaxOwok:
FWIW, I don't think 94bebc3 is sufficient... I did the same thing for `/projects` children, but I think it needs to be programmatically added to routes rendered through `childRoutes`.  

For instance, in the backend `/records/pages/1/anything-in-here` will also mount an empty body rather than throwing a 404.

PR #1604, opened 2018-11-02
[B] Validate ProjectCollection#number_of_projects

### LINKED ISSUE #1596: Projects visible field allows non-number inputs
For smart collections the `Projects Visible` field generally doesn't accept non-number input, but four characters escape that condition: the period `.`, plus sign `+`, dash `-`, and the letter `e`:

visible

Attempting to proceed with any of these prompts an error message, but it's curious to be able to enter these and no other letters/symbols (beside numbers).

SMaxOwok:
Fixes #1596


SMaxOwok:
See my comment on #1596.  I personally don't think we should sanitize the actual input, but if that's desired, let me know and I will adjust accordingly.

Issue #1441, opened 2018-10-03
Long press logos collide with header nav links


SMaxOwok:
![screen shot 2018-10-03 at 9 14 40 am](https://user-images.githubusercontent.com/8389445/46424019-dac93100-c6ec-11e8-890a-62b5a5128926.png)

PR #1608, opened 2018-11-03
[C] Refactor app header to better handle logos

dananjohnson:
The app header layout was previously quite static, which led to problems with content collision now that custom logos can be supplied (see #1441). The solution required refactoring the markup and styles for this component to scale and position logos and layout header elements flexibly. Using CSS grid and `max-width` and `max-height` properties, header elements now scale to available space in the container.


zdavis:
From Dana: "There’s one remaining issue with positioning the notification bar when a user logs out. It’s getting cut off when a custom logo is used, and I’m not sure how to conditionally adjust it. Basically, the header was set up to be taller when a custom logo was used. I think this is the right choice, but in these situations, the notification bar has to be pushed down a bit further."

zdavis:
Left some comments on this via DM.

PR #1614, opened 2018-11-05
[B] Assign header IDs to all converted docs

### LINKED ISSUE #1613: Header IDs need to be assigned to all document ingestions
Currently we only assign IDs when a document goes through the HTML converter.  These need to be applied to documents that go through other conversions.

SMaxOwok:
Fixes #1613


SMaxOwok:
Also normalizes converter method names a little more.

PR #1621, opened 2018-11-08
[C] Only wrap project import in transaction

SMaxOwok:
With the latest changes to the dev loader, if something fails
at any point in the `#load` process the whole thing is rolled
back.  The rest of the seed data, besides projects, seems safe to
commit immediately.  This also enables being able to work with the
app while project data is loading.

### DIFF
--- a/api/app/services/demonstration/data_loader.rb
+++ b/api/app/services/demonstration/data_loader.rb
@@ -13,16 +13,14 @@ module Demonstration
     end
 
     def load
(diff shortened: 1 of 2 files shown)


zdavis:
Would it also work to just wrap each project import in a transaction? Do all of them need to be in a single transaction?

SMaxOwok:
Seems to be working alright wrapping individual projects instead.

SMaxOwok:
Will be part of #1630

PR #1629, opened 2018-11-12
[F] Format stylesheet raw_styles before saving

### LINKED ISSUE #1309: Raw styles should have line breaks
We should format raw styles a little before saving.  Currently if the input is an unbroken string, it's really hard to edit styles.

screen shot 2018-08-10 at 10 33 15 am

SMaxOwok:
Closes #1309


SMaxOwok:
Note that this transformation also happens during ingestion post processing.

The reason why this isn't universally taken care of in the model is that we associate text sections and stylesheets by the hash content of their raw_styles during ingestion.

If we format the raw_styles when stylesheets are created, the hashed content is changed and the association between stylesheet and text sections will be lost.  This PR adds the formatting to stylesheet post processing, which occurs after stylesheets and text sections are persisted and associated.

zdavis:
Does this affect reingestion? Do we rely on that hash during re-ingest to update instead of create records?

SMaxOwok:
I don't think so... 

We update/create stylesheets (before text sections) based on their `source_identifier`, which is related to the source file.  

The text sections associate stylesheets by `hashed_content`, which should be then updated on the stylesheet records at that point.

SMaxOwok:
We may be able to improve stylesheets more and be able to do this formatting before associations actually.  

We have access to the source files when we assign the text section stylesheet contents.  I don't know that there's a reason why we couldn't be using source identifiers there.  That would let us find stylesheets by their source identifiers instead of hashed content in the text section compiler.

Let me know if you want me to try and make this change as part of this branch.

zdavis:
Sure, give it a go.

SMaxOwok:
Ok, I might have been wrong about that.  I think this should stay as it is...

SMaxOwok:
@zdavis ok, should maintain stylesheet relationships now.

PR #1650, opened 2018-11-14
[B] PaperclipMigrator supports versionless uploaders

SMaxOwok:


api/lib/paperclip_migrator.rb
```diff
@@ -120,6 +120,8 @@ def execute
       # First we convert the data as it exists in paperclip
       shrine_data = shrine_data_for instance, attachment_name, *style_names, options.symbolize_keys
 
+      shrine_data = shrine_data[:original] unless uploader_supports_versions?
```
SMaxOwok: Not sure this is the best way, but couldn't come up with a more programatic way to determine what version to use besides `values.first`.  Ideally it would be whatever the `:default_style` is set to in `PseudoAttachment`.

PR #1663, opened 2018-11-14
[B] Scope reingestion websocket error to global

SMaxOwok:


zdavis:
Can I see a screenshot of what this error looks like in the global notification container?

SMaxOwok:
screen shot 2018-11-14 at 2 04 35 pm

PR #1675, opened 2018-11-16
[B] Improve unauthenticated authorization

### LINKED ISSUE #1670: Project#show does not authorize the resource
We need to look at the controllers in project/relationships as well as some top level controllers and make sure that they're authorizing the request correctly. With draft projects, I don't think users should be able to see child views.

SMaxOwok:
This commit addresses a couple issues:

1. We weren't previously authorizing Project/Text#show, which meant
that direct navigation to draft projects and those projects' texts
always succeeded, even when unauthenticated.

2. We were rescuing Authority::MissingUser errors with a 401 error
requiring login.  This is incorrect when an unauthenticated user is
attempting to view a Project or Text and we authorize the action.
Instead, we now create an `anonymous_user` that has reader privileges
and defers its `can_read?` check to the resource we're verifying.

Fixes #1670

### DIFF
--- a/api/app/authorizers/annotation_authorizer.rb
+++ b/api/app/authorizers/annotation_authorizer.rb
@@ -6,7 +6,7 @@ class AnnotationAuthorizer < ApplicationAuthorizer
   end
 
   def creatable_by?(user, _options = {})
-    return true unless Annotation::NOTATION_TYPES.include?(resource.format)
+    return known_user?(user) unless Annotation::NOTATION_TYPES.include?(resource.format)
     resource&.text&.notatable_by? user || false
   end
 
--- a/api/app/authorizers/application_authorizer.rb
+++ b/api/app/authorizers/application_authorizer.rb
@@ -66,6 +66,10 @@ class ApplicationAuthorizer < Authority::Authorizer
 
   protected
 
+  def known_user?(user)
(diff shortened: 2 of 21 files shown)


SMaxOwok:
Needs a little client adjustment for text auth errors

SMaxOwok:
I think that is an issue in the `AnnotationAuthorizer` rather than the anonymous user, at the moment.

https://github.com/ManifoldScholar/manifold/blob/master/api/app/authorizers/annotation_authorizer.rb#L9

Any thoughts on what we should check for here against a user?  Or maybe we need to add a separate role for the anonymous user and adjust from there.

zdavis:
Could you please summarize what you changed in response to my comments? If you include a comment saying how you addressed my concerns, it saves me the trouble of having to re-familiarize myself with the PR and reviewing it all again. Thanks.

SMaxOwok:
The changes I made were to remove the reader role/kind from the anonymous user and to instead return nil for those.  All the authorizers that don't rely on an authenticated user should still work, as they return true rather than a value based on a role.  

For the specific case you added a test for, I added a `known_user?` method to the application authorizer that looks for any role.  We then use that method to determine whether a user can leave an annotation or not.

PR #1677, opened 2018-11-19
[F] Improve ProjectCollection smart sorting

SMaxOwok:


api/app/models/event.rb
```diff
@@ -52,6 +52,8 @@ class Event < ApplicationRecord
 
   scope :search_import, -> { includes(:project) }
 
+  after_create :touch_project!
```
zdavis: Seems like this should happen in after_commit rather than after_save

PR #1678, opened 2018-11-19
[B] Display base model errors in FormContainer

### LINKED ISSUE #1643: FormContainer should display base model errors
The FormContainer component should include an error field that responds to errors on the model `:base`.  Right now these errors aren't displayed anywhere on the frontend.

Additionally, the existing `groupErrors` prop/errors in the form container should ignore `:base` errors after the above is added.

SMaxOwok:
Fixes #1643


SMaxOwok:
I only explicitly added `modelName` prop to the maker form (where we now return a base attribute error).  I'm happy to go through and it to the other forms, if desired.

PR #1683, opened 2018-11-20
[B] Fix ingestion file rejection regex

### LINKED ISSUE #1682: EPUB ingestion unzip regression
High priority.

It seems like the recent changes in https://github.com/ManifoldScholar/manifold/commit/c62373b8fa9d810e774faee15246f7e3a5b04ab2 seems to be preventing `book.opf` type files from being copied.

SMaxOwok:
Fixes #1682


SMaxOwok:
The idea is anything that matches `.`, `_`, or `~` any number of times _plus_ a single character.

Sometimes the dirname coming in here is `.` or `..`, which is causing some ingestions to not copy needed source files.

PR #1698, opened 2018-11-30
[B] Proxy IS attachments through API controller

### LINKED ISSUE #1697: Links to ingestion sources in text section bodies are not updated when ingestion source attachments are promoted

zdavis:
Fixes #1697

### DIFF
--- a/api/app/models/text.rb
+++ b/api/app/models/text.rb
@@ -178,7 +178,7 @@ class Text < ApplicationRecord
   def source_path_map
     map = {}
     ingestion_sources.each do |s|
-      map[s.source_path] = s.attachment_url
+      map[s.source_path] = s.proxy_path
     end
     map
   end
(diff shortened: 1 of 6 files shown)


api/app/models/text.rb
```diff
@@ -178,7 +178,7 @@ def section_source_map
   def source_path_map
     map = {}
     ingestion_sources.each do |s|
-      map[s.source_path] = s.attachment_url
+      map[s.source_path] = "#{prefix}/api/proxy/ingestion_sources/#{s.id}"
```
scryptmouse: Without including helpers, you could do something like:

```ruby
class IngestionSource < ApplicationRecord
  def proxy_path
    self.class.proxy_path self
  end

  class << self
    def proxy_path(ingestion_source)
      Rails.application.routes.url_helpers.api_proxy_ingestion_source_path ingestion_source
    end
  end
end
```

I don't think `Text` is the right place to know how to generate a path to an `IngestionSource`, it should just know to ask IngestionSource for its own path. This also avoids using the static text path and simplifies the test for this (since we don't have to test that Rails route helpers work as expected, as that's TTFW)

PR #1700, opened 2018-12-03
[B] Return nil when style value is empty

### LINKED ISSUE #1699: Attachment style hashes should return nil for missing style keys
https://github.com/ManifoldScholar/manifold/blob/master/api/app/models/concerns/attachments.rb#L216

This should be nil instead of always returning the original version

SMaxOwok:
Previously this was always returning the original attachment version
for _all_ style keys that did not have a version associated with them.
Fixes #1699


SMaxOwok:
@zdavis check to see if your project covers all appear.  Some of mine were broken, but I might have been in a weird state from before.  Running dev loader now everything is looking OK.

zdavis:
Discussed further in meeting—Max is going to take it a little further.

PR #1706, opened 2018-12-04
[B] Fix error boundary lifecycle usages

SMaxOwok:


client/src/global/components/fatal-error/Boundary.js
```diff
@@ -25,13 +25,11 @@ class FatalErrorBoundary extends Component {
   }
 
   componentDidCatch(error, info) {
-    // This will soon be deprecated.
```
SMaxOwok: I took this comment out because I didn't see anything about it being deprecated soon.  Not sure if I missed something or if this was an old comment or something.

zdavis: I think it should be left there. From the React.Component docs: https://reactjs.org/docs/react-component.html

Is getDerivedStateFromError in fact being called? I see the mistake in setting state instead of returning an object, but if memory serves, I was running into problems getting it to call when errors were caught. If that is in fact working, I don't think we need to set the state in `componentDidCatch`.

SMaxOwok:
Ok, @zdavis, I made some bigger changes here.  My thoughts are below, but I'm happy to adjust or remove these additional changes if you feel them unnecessary or unwanted.

As I understand it from the docs, `componentDidCatch` provides the component stack for purposes of logging to a service or outside of the render, which is why fallback rendering with that method will be deprecated.  To that end, I removed the `setState` occurring there and the component stack trace from the error component.

All the same information in the component stack trace is logged to the console by default, so `componentDidCatch` just needs to be present in the component, but does not need to be triggering re-renders with `setState`, especially since the component stack track is only rendered in development.  This varies in comparison to the other stack traces (API, client), in that those are already included in the returned error itself.

You might feel those extra renders are fine for this, in which case I will revert those changes.

Additionally, I added a `console.error` to the `ClientTrace` `componentDidMount` as the error component truncates the trace and points to the console, but the error wasn't logged there.

SMaxOwok:
**Development**
development

**Production**
production

PR #1710, opened 2018-12-05
[B] Improve upload file name previews

### LINKED ISSUE #1704: Images in upload components that are meant to accept files have unexpected thumbnail behavior
To reproduce, upload an image to the "download file" field at the bottom of project/appearance. You'll get output like this:

image

SMaxOwok:
Fixes #1704


SMaxOwok:
Probably only thing worth noting, though it would be expected given the solution we decided on.

image

SMaxOwok:
As discussed, will add `fileNameFrom` prop to specify an attribute to read the file name.

PR #1729, opened 2018-12-07
[F] Refactor client search containers

SMaxOwok:


SMaxOwok:
This ended up being a lot bigger than I intended, but I think it's better going forward.  In addition to using the store instead of state to manage search queries, I broke the main logic of when search containers should update into an HOC function.

client/src/hoc/with-search/index.js
```diff
@@ -0,0 +1,153 @@
...
+        setPage: this.setPage,
+        searchStateQuery: this.props.searchQueryState,
+        history: this.props.history,
+        location: this.props.location,
+        ref: this.searchComponent
+      };
+    }
+
+    /* eslint-disable no-console */
+    get searchParams() {
```
SMaxOwok: This is the only part I'm not excited about.  It works, and should continue to work, but it would be nice if there was a better way to define the logic for facets in a search container than calling the child component method.  Static methods would be slightly better, but we still need access to props in them.

zdavis:
@SMaxOwok when you've gleaned everything you need to glean from this PR, please close it, as I'm going to accept #1741 instead

zdavis:
Closing this to clear up the PR list, but it's still around if you want to look at it.

PR #1732, opened 2018-12-07
[B] Don't exclude retweets in twitter fetch

### LINKED ISSUE #1728: Retweets should be excluded from project activity

SMaxOwok:
Fixes #1728


SMaxOwok:
This will still work if the original query includes `-rt` anyways.

PR #1741, opened 2018-12-13
[F] Refactor client search to use store

zdavis:


SMaxOwok:
- Search results pagination isn't working

PR #1742, opened 2018-12-13
[B] Address OAuth regression

### LINKED ISSUE #1631: Investigate google oauth login
@jojokarlin reports CUNY installation users encountering issues trying to login to manifold with google credentials

zdavis:
When we installed Draper to provide decorators to our mailers, we
inadvertently broke the OAuth controller. This is likely due to the
following issues in Draper, which do not appear to have been fully
solved in Draper 3.0.1.

Because this is the only place we use ActionView outside of Mailers,
for now we will just render the content from the controller. In the
future, if we need more view functionality, we'll use cells.

See:
https://github.com/rails/rails/issues/27211
https://github.com/drapergem/draper/pull/793

Fixes #1631


api/app/controllers/oauth_controller.rb
```diff
@@ -11,11 +10,30 @@ def authorize
 
     @oauth_payload = ExternalAuth::Payload.new outcome
 
-    render layout: false
+    render html: body, layout: false
```
SMaxOwok: Needs `.html_safe` or equivalent to not just render as text on frontend.

SMaxOwok:
Disabled the warning on `html_safe` as we don't have view helpers there and because we're just rendering that block.

Issue #1752, opened 2018-12-19
Radio form component does not display errors


SMaxOwok:

SMaxOwok:
Hmm, doesn't seem to be used anywhere and pretty out of date with our other form components.  Might need a little more than just Errorable.

PR #1754, opened 2019-01-02
[F] Implement HOC to map form input options

SMaxOwok:


SMaxOwok:
Also fixes #1752

SMaxOwok:
Testing in my content blocks branch, but it's not working now.  Needs adjustment, I think.

SMaxOwok:
Fixed.

SMaxOwok:
options

PR #1765, opened 2019-01-15
[F] Add hero content blocks to projects

SMaxOwok:


api/db/migrate/20190115182442_add_hero_block_to_existing_projects.rb
```diff
@@ -0,0 +1,16 @@
+class AddHeroBlockToExistingProjects < ActiveRecord::Migration[5.0]
```
zdavis: Remove from this PR; we will tackle all migrations at once.

api/db/schema.rb
```diff
@@ -81,14 +81,14 @@
...
     t.uuid     "project_id"
-    t.string   "thumbnail_file_name_deprecated"
-    t.string   "thumbnail_content_type_deprecated"
-    t.integer  "thumbnail_file_size_deprecated"
-    t.datetime "thumbnail_updated_at_deprecated"
     t.string   "thumbnail_checksum"
     t.string   "fingerprint"
     t.datetime "created_at",                                     null: false
     t.datetime "updated_at",                                     null: false
+    t.string   "thumbnail_file_name_deprecated"
```
zdavis: This shouldn't be in this PR.

api/db/schema.rb
```diff
@@ -81,14 +81,14 @@
...
+    t.string   "thumbnail_content_type_deprecated"
+    t.integer  "thumbnail_file_size_deprecated"
+    t.datetime "thumbnail_updated_at_deprecated"
```
zdavis: Pretty sure this shouldn't be here either.

api/spec/authorizers/content_block_authorizer_spec.rb
```diff
@@ -71,4 +71,15 @@
 
     the_subject_behaves_like "instance abilities", ContentBlock, none: true
   end
+
+  context "when the block is a title block" do
```
zdavis: Should say hero block.

SMaxOwok:
One thing to note, the dragging functionality may be a little weird without title blocks.  Some changes to the ordering was necessary in order to anchor a block to the top of the list.  It should work correctly when the drag-disabled block is in the 0 position.

Saving a project before adding other blocks will create a HeroBlock and should prevent you from encountering the problem.

PR #1771, opened 2019-01-15
Add stubbed frontend content blocks

dananjohnson:


zdavis:
I fixed this branch (tests and linting) so that CI should pass. I've force pushed these changes.

Rather than migrate all project pages to content blocks, I'm inclined to take an approach where projects without content blocks get some default display (hero, activity, texts, resources, metadata). We'll have to mock empty content blocks for these, which should be fine, but think about where this will happen and consider even getting this in place now.

I imagine something like this in frontend/components/project/Content/index.js:

```
  get projectHasContentBlocks() {
    return Array.isArray(this.projectContentBlocks) && this.projectContentBlocks.length > 0
  }

  get projectContentBlocks() {
    return this.project.relationships.contentBlocks;
  }

  get defaultContentBlocks() {
    return [
      // an array of mocked content block models.
    ];
  }

  get contentBlocks() {
     return this.projectHasContentBlocks ? this.projectContentBlocks : this.defaultContentBlocks;
  }
```

I can't think of a case where we'd ever want to show an empty project page.

Also, be sure to update the relationships on the Project fixture in test/fixtures/entity.js... we should add an empty array to the content block relationships.

Think, too, about when props should be required. As you had it, the Content component would throw errors if props.project wasn't present (because it the content block relationships to be an array and calls map on it, and if it were null we'd have an exception). If you're not going to write code to deal with the absence of a prop (which is OK), then you need to set the prop to required in the propTypes. Make sense?

best,
Zach

Issue #1758, opened 2019-01-04
Improve Form.HasMany component


SMaxOwok:
Our current `Form.HasMany` component always fetches options.  In many cases, this is proving unnecessary.

For example, when an entire project is available, a select with options from relationship records does not need to be refetched.  These can be passed into the `HasMany` component directly.

As part of this, the list used in `FetchSelect` can be pulled out and reused as a component that serves both that and `HasMany`.

SMaxOwok:
Also seeing a flicker when saving between form data and entity data.

zdavis:
@bot accept

zdavis:
Thank you for taking the time to open this feature request. The Manifold team reviewed this issue during our bi-weekly meeting and the consensus is that this feature makes sense and is in keeping with our overall vision for the platform. Moreover, we see this request as a viable candidate for development under our current available funding. We’re adding an “accepted” label to this request to indicate that it’s within scope and possibly within budget.

The next step is for us to estimate the work involved with this and add it to our feature backlog. Our acceptance of the issue is not a promise that it will be implemented. We will balance this request against the other accepted requests and do our best to implement it within our current available funding.

_This was an automated message, but please don't hesitate to reply. Our team watches these issues closely and will respond as soon as we're able to!_

zdavis:
No longer needed, as we've removed this component and replaced it with the new Form.Picker component.

👋 Max!

SMaxOwok:
RIP 👻

PR #1781, opened 2019-01-16
[C] Allow HasMany to accept options array

SMaxOwok:
The HasMany/OptionsList/ConnectedSelect components can likely be
further refactored, but this simply allows HasMany to take an array
of options and avoid fetching when data is already available.


SMaxOwok:
This is necessary to continue building out some of the content blocks, but as mentioned, can probably be taken further and actually refactor these components a little more to address #1758

PR #1782, opened 2019-01-17
Improve hero position UI logic

zdavis:


client/src/backend/components/project/Content/index.js
```diff
@@ -176,30 +160,6 @@ export class ProjectContent extends PureComponent {
...
-
-  adjustedPosition(position) {
-    const max = this.state.blocks.length;
-
-    if (position >= max) return "bottom";
-    return position;
-  }
-
   handleAddEntity = type => {
     return this.insert(type, 1);
```
SMaxOwok: This has to be updated too since `insert` was removed.  Needs to make a pending block of whatever type and prepend to currentBlocks.  Or maybe just push to currentBlocks?

PR #1784, opened 2019-01-17
Add TextsBlock and backend form

SMaxOwok:
Depends on #1781 being accepted and merged into `feature/content-blocks`.


SMaxOwok:
@dananjohnson could you make a new icon component for the layered texts icon and maybe make some style adjustments to the switches in the drawer?

dananjohnson:
Will do @SMaxOwok !

PR #1788, opened 2019-01-17
[F] Add programatic project collection visibility

SMaxOwok:

Issue #1669, opened 2018-11-16
Allow users to specify how many projects a collection should include on the HP


mkgold:
## Feature Description
Right now, when collections with 8 or more texts appear on the homepage, 8 texts show up. This could start to be an issue when an installation has, say, dozens of collections (imagine the CUNY instance, for example, where the hope is that there will eventually be collections for every academic discipline). 

We should think about allowing admins to limit the number of texts that the collections section displays -- at least to 4, I would think, so that the vertical real estate taken up by the collection on the home page does not become too large.

Eventually, and possibly for larger installations, we may even need to think about a kind of directory of collections.

## *Why* is this feature important? *Who* does it help?
this feature will be especially important for large instances with lots of different collections. It will be difficult to scroll through the homepage of an instance with a large number of collections if each one shows 8 texts.

zdavis:
A directory of collections already exists: https://cuny.manifoldapp.org/projects/project-collections

zdavis:
I also think we should do this—it's a no brainer. The only thing I'm wondering is whether we should let users set an arbitrary number to display, or if we should let them select in multiples of four. Curious what others think.

tsmyre:
Why not just define it by rows? If I have five or eight projects it doesn't matter; in either case it's two rows.

zdavis:
Well, rows aren't really relevant on mobile, where these turn into a list... but maybe that's fine.

susanmpls:
Arbitrary numbers in collectionsajes most sense to me. For instance, if a series has only 7 titles, we might want a collection with 7 projects.

tsmyre:
@mkgold Just to clarify, are you suggestion an option for displaying a collection name on the home page but not revealing any of the projects it contains, regardless of what that number might be, e.g.:

***
Collection Name
Collection Description
Collection Link →
***

susanmpls:
How does everyone feel about offering three options for number of projects displayed for a collection from the home page or on the directory of collections page: 0, 1-4, or 4-8. 

I'm reluctant to show more than 8 because it takes up so much space on the HP. If a reader is interested in seeing all the books (or more than 8) then they should navigate to see the full collection.

tsmyre:
I like that. Unless there's an easy way to make the projects somehow smaller and made into a list view, which could be packed in there a bit more without sacrificing experience, but I think that'd be introducing something a bit more than necessary at the moment.

PR #1791, opened 2019-01-17
[F] Allow PC's to specify visible HP projects

SMaxOwok:
Built on #1788 because of the form refactor.


zdavis:
Issue number is wrong. Should be #1669

PR #1800, opened 2019-01-22
[C] Move project published text to Text

SMaxOwok:
Some notes:
- Migration down does not set a published text on projects.  Should it just take the first text marked published and use that?
- Stripped out `renderPublishedText` from `ProjectHero` component.  Should not affect anything as that component is being replaced by the `HeroBlock`.
- Removed published category from backend project texts.  Also adjusted arrow logic.  Both of these changes will be obsoleted by Zach's drag and drop interface.


zdavis:
I'm a little confused about this comment:

`Migration down does not set a published text on projects. Should it just take the first text marked published and use that?`

The migration definitely should not set an unpublished text to published, but maybe I'm reading this wrong.

If a project has a published text, the migration should set a "published" flag on that text to true. The goal is to move where the published state is stored from the project (which can only have one published text) to the text itself.

SMaxOwok:
The down being rolling back from having `published` on the text back to a `published_text_id` column on a project.  Are we ok with not setting a relation there?  Or if not, are we OK just taking the first text association marked published and using that.

We're moving to any text being able to be marked published, so rolling back would be going back to having a single published text.

SMaxOwok:
Note to self: I was going to split this up into things that we can merge now and things that need to be migrated with the Hero additions.

PR #1804, opened 2019-01-22
Misc. bug fixes

SMaxOwok:


client/src/theme/Components/browse/resource/_totals.scss
```diff
@@ -22,6 +22,10 @@
 
     &:hover {
       color: $accentPrimaryDark;
+
+      .resource-total__value {
```
SMaxOwok: Nesting OK here per @dananjohnson

PR #1809, opened 2019-01-23
Refactor frontend TextsBlock component

SMaxOwok:


SMaxOwok:
@dananjohnson these are the style things I think are outstanding:
- baseClass default props should be what the actual component is.  i'm passing down the parent when we're using the component with context.  i think i renamed them all, but worth a double check.
- class names should be updated to reflect component usage in isolation.  i.e., no more --grouped/--ungrouped.
- published badge should have border radius.
- last item in list has border now that class names have changed.
- dates should be updated/added at depending on which is newer.

PR #1817, opened 2019-01-24
[B] Lock ResourceImportRow import when executing

SMaxOwok:


SMaxOwok:
Requesting @scryptmouse takes a look at this branch quickly to see if you have a better idea why it's hanging up.

zdavis:
We solved these issues by updating Rails to 5.1, which included a fix for deadlocks in development.

PR #1830, opened 2019-01-25
Move project published text to Text

SMaxOwok:


api/db/migrate/20190122160027_move_published_text_from_projects_to_texts.rb
```diff
@@ -0,0 +1,30 @@
...
+        WHERE id IN (SELECT published_text_id FROM projects)
+    SQL
+
+    remove_belongs_to :projects, :published_text, type: :uuid
+  end
+
+  def down
+    add_belongs_to :projects, :published_text, type: :uuid
+
+    execute <<-SQL.squish
```
SMaxOwok: Can someone test this?

I first rolled back without the `INNER JOIN` so my data got into a weird state with mismatched texts.  It looked like it worked, but I just want to be sure with valid data.

SMaxOwok: Or maybe @scryptmouse can tell by just looking at it.

zdavis:
Please update the migration time stamps and I will test again.

api/db/migrate/20190122160027_move_published_text_from_projects_to_texts.rb
zdavis: Tested. Works.

PR #1838, opened 2019-01-28
[C] Drawers do not trigger page scrolling

SMaxOwok:
This commit changes drawer route behavior to never trigger parent
route scrolling. Since the drawer routes should be considered as an
interaction similar to an overlay, they should not change anything
about the page that opens them.


SMaxOwok:
One thing to note is that the focus trap tracks the last element focused _before_ the drawer is opened.   In a case like the makers list, if you click a name (drawer opens), then scroll down and click another name (drawer changes), the page will scroll back to the first element when the drawer is closed.

We can prevent behavior like this by scroll locking those usages or with adjustments to how we use focus trap there (not sure exactly how yet).

PR #1842, opened 2019-01-28
Build out Table of Contents block

dananjohnson:


api/app/serializers/text_serializer.rb
```diff
@@ -6,7 +6,7 @@ class TextSerializer < ApplicationSerializer
              :published, :annotations_count, :highlights_count, :bookmarks_count,
              :age, :position, :publication_date, :cover_styles, :subtitle,
              :slug, :section_kind, :title_formatted, :title_plaintext,
-             :subtitle_formatted, :subtitle_plaintext, :description
+             :subtitle_formatted, :subtitle_plaintext, :description, :toc
```
SMaxOwok: The `toc` attribute from https://github.com/ManifoldScholar/manifold/blob/master/api/app/serializers/text_full_serializer.rb#L7 can also be removed as part of this.  

Sometimes for big models we have two serializers: a partial one and a full one.  In cases where we just need some data for a model, we prefer the partial to speed things up.  The full serializer inherits everything defined in the partial one, so if an attribute is moved from one, it doesn't need to be defined in the other as well.

client/src/frontend/components/content-block/Block/types/TOCBlock/Contents.js
```diff
@@ -0,0 +1,151 @@
...
+          {/* eslint-disable-next-line no-shadow */}
+          {node.children.map(node => this.renderNode(node, depth))}
+        </ul>
+      );
+    }
+
+    const anchor = node.anchor ? `#${node.anchor}` : "";
+
+    return (
+      <li key={this.keyCounter++} className={`${blockClass}__node`}>
```
SMaxOwok: Just a suggestion: you could eliminate the need for this entire counter by using the node's id instead.

api/app/serializers/text_serializer.rb
dananjohnson: OK nice! thanks for the explanation.

client/src/frontend/components/content-block/Block/types/TOCBlock/Contents.js
dananjohnson: I was getting duplicate keys initially, but maybe after fixing how the depth was being tracked, this would work. Thanks.

dananjohnson: @SMaxOwok I switched this to `node.id` but am getting the duplicate keys warning. Is it clear to you what I'm doing wrong?

SMaxOwok: Hmm, I'll pull and take a quick look.  I dunno why they wouldn't be unique.

SMaxOwok: Ok, so it's because sometimes a line in the TOC is to a specific part of a page.  There can be more than one link to the same page.

`${node.id}${anchor}` works reliably, but I dunno if it's necessarily better than the counter you had set up.

Curious if @zdavis has any feelings, although this thread has probably turned into a useless suggestion... sorry!

zdavis: It's still possible that a TOC could include two links that were identical. In this case, because we can be pretty darn certain the JSX won't change after it's rendered, I'm OK with using an index-like counter, even though Eslint warns against it. That said, a better solution would be to include a UUID for each TOC entry on the server-side. That would involve writing a migration, of course, and updating some ingestion code.

PR #1843, opened 2019-01-28
[C] Rename Collection model to ResourceCollection

SMaxOwok:
This commit renames and refactors uses of Collection as ResourceCollection.
We're doing this because it is no longer immediately clear what collection
refers to.  With the addition of ProjectCollections and potentially future
collection kinds, we should take care of this now, rather than later as
things get more complicated.

This particular commit is limited to the change in the API, which will
facilitate a refactor of the frontend components and organization.


SMaxOwok:
I think this is ready for review.

The new container _does_ include pagination, but not filtering or query params.  I figured that could be easily added in, but for the time-being it was important to get this finished for the content block development.  

The other thing is after navigating to a collection, the back links go to the project.  We'd need to add additional logic to determine if it should go to the project or all collections page.

PR #1855, opened 2019-01-31
Let TOCBlock title be configured

SMaxOwok:


SMaxOwok:
I dunno how you all feel about 3391736.  This should let any block be able to render a title override attribute though.

zdavis:
@max FYI, I corrected the failing test and force pushed.

SMaxOwok:
Thanks!

zdavis:
The approach feels a little weird to me, because while the title is set in the static methods in each content block component, the actual determination comes from outside the block. I guess I feel like it would be simpler to have a method in each content block:

```
get title() {
    return this.props.contentBlock.attributes.title || "Foo"
}
```

Yes, there's a little redundancy around how the title is accessed, but that seems simpler and easier for future developers to understand than the indirect method proposed in this PR. And we already have to have a title() method in each component, so why not just make the decision there?

SMaxOwok:
Yeah, totally, I can drop that commit and adjust the other.

... On second though, I think I like your suggestion of removing the title prop and adjusting the getter methods to just return a default.

zdavis:
In looking at this, I noticed that most of the components use a <Header> element, but the texts list one uses a <Heading> element. Is that discrepancy necessary/intentional?

SMaxOwok:
It is because the description included in that block is not part of the generic content block `<Heading />` component.  It's possible this could be moved into the generic one, but right now the texts block is the only one with an additional description.

zdavis:
Instead of creating a new component, I'd be inclined to adjust the existing component. We could just make the text block accept children, so we could do this in the TextsBlock component. Or, if we think we'll have description on more blocks (which seems totally possible), we could add a desc. prop.

```
return (
  <Wrapper>
    <Header
      title={this.props.title}
      icon={this.props.icon}
      block={this.props.block}
      baseClass={baseClass}
    >
			{this.blockDescription && (
        <div className={`${this.props.baseClass}__details`}>
          <div
            className="description pad-bottom"
            dangerouslySetInnerHTML={{
              __html: this.blockDescription
            }}
          />
        </div>
      )}
    </Header>
  	[...]
	</Wrapper>
);

PR #1863, opened 2019-02-01
[F] Improve manifest ingestion TOC control

### LINKED ISSUE #1857: Better control of TOC generation in the manifest
I have some issues with TOC generation.

I have an HTML file which uses h1 and h2 headers. When I ingest the text by uploading the HTML file only the auto-generated TOC contains both types of headers. The resulting text, however, is to long to be shown at once. I therefore split the file into 3 HTML files - each file for one of the h1 headers uns use the manifest.yml ingest procedure. Everything works well, except, the TOC now only includes the three h1 headers without the h2 headers. Hence, it seems the HTML files are no longer parsed for headers beyond those I assigned with labels in the manifest.yml. This is unfortunate insofar it is also not reasonable in my case to split the document further. Consider a 400 pages book with 3 levels of headings which you will then have to split up into, let's guess, 50 documents for the only reason to have them recognized in the TOC.

Is there a way to combine TOC generation via parsing the documents and declaring sections in the manifest holding its own files. I tried out to declare sections in the manifest by giving them the same label in the manifest than the section's title within the document, but this didn't work. I think if something comparable is no possible two logics conflict here, one being the reason for splitting up a document for not having endless websites, the other being the whish to control what appears in the TOC.

Suggestion from discussion on Slack using URL fragments:

https://manifold-community.slack.com/archives/C8LFMHM0X/p1549030734053200

```
toc:
  - label: Foo
    source_path: foo.html
    start_section: true
    children:
      - label: Bar Part 1
        source_path: bar.md#part-1
      - label: Bar Part 2
        source_path: bar.md#part-2
      - label: Bar Part 3
        source_path: bar.md#part-3
  - label: Baz
    source_path: baz.html
```

SMaxOwok:
Prior to this commit, manifest ingestions would only include a TOC entry
for an entire source file.  This change allows a manifest source path to
include an anchor.  This will build out the TOC with links _within_ a source
file.
Closes #1857


zdavis:
I'm worried that there are unintended side effects in this PR. I changed an anchor to a line in the manifest.yml spec fixture:

```
meta:
  title: A Manifest Ingestion
  language: language
  rights: rights
  description: description
  creators:
    - name: Rowan Ono
  contributors:
    - name: Ida Davis
toc:
  - label: Title Set From TOC
    source_path: section_1.html
    start_section: true
    children:
      - label: Section 1.1
        source_path: section_1_1.html
        children:
          - label: Section 1.1a
            source_path: section_1_1a.md#foo
```

And when I do that, the source identifier changes in a spec, which seems wrong:
```
  2) Ingestions::Strategies::Manifest when collection of different file kinds the returned manifest its relationship attributes has the correct text section attributes
     Failure/Error: expect(manifest[:relationships][:text_sections]).to eq expected

       expected: [{"source_identifier"=>"02e9b3b8b5268c3fe8bd4150fd546a55", "name"=>"Title Set From TOC", "kind"=>"sec...on 1.1a", "kind"=>"section", "position"=>2, "build"=>"build/e2fe424ecf8e3c87da127ea617e57107.html"}]
            got: [{"source_identifier"=>"02e9b3b8b5268c3fe8bd4150fd546a55", "name"=>"Title Set From TOC", "kind"=>"sec...on 1.1a", "kind"=>"section", "position"=>2, "build"=>"build/8b7a46c8eee38aaaa70260f92eba6c8d.html"}]

       (compared using ==)

       Diff:

       @@ -8,9 +8,9 @@
          "kind"=>"section",
          "position"=>1,
          "build"=>"build/9fef2b7e781641e6861d1aa79a597a57.html"},
       - {"source_identifier"=>"e2fe424ecf8e3c87da127ea617e57107",
       + {"source_identifier"=>"8b7a46c8eee38aaaa70260f92eba6c8d",
          "name"=>"Section 1.1a",
          "kind"=>"section",
          "position"=>2,
       -  "build"=>"build/e2fe424ecf8e3c87da127ea617e57107.html"}]
       +  "build"=>"build/8b7a46c8eee38aaaa70260f92eba6c8d.html"}]

     Shared Example Group: "outcome assertions" called from ./spec/services/ingestions/strategies/manifest_spec.rb:87
     # ./spec/services/ingestions/strategies/manifest_spec.rb:67:in `block (5 levels) in <top (required)>'
     # ./spec/rails_helper.rb:90:in `block (2 levels) in <top (required)>'
     # -e:1:in `<main>'

Finished in 8.71 seconds (files took 0.53526 seconds to load)
13 examples, 2 failures

Failed examples:

rspec ./spec/services/ingestions/strategies/manifest_spec.rb:25 # Ingestions::Strategies::Manifest when collection of different file kinds the returned manifest its text attributes has the correct TOC
rspec ./spec/services/ingestions/strategies/manifest_spec.rb:63 # Ingestions::Strategies::Manifest when collection of different file kinds the returned manifest its relationship attributes has the correct text section attributes
```

I'd expect the TOC to change, but not the section identifier. This also will introduce some ambiguity around how we name text sections. For example:
```
          - label: Section 1.1a
            source_path: section_1_1a.md#foo
          - label: Section 1.1b
            source_path: section_1_1a.md#bar
```
What should that text_section model be named? I'm not sure we can resolve that, but I think we'd generally be more correct if we used the first label we came across rather than the last as we move recursively through the tree.

zdavis:
If the source identifier changes, this could lead to a manifest change causing a text section and all annotations to be deleted rather than updated.

I'd like to see some specs supporting this new behavior too, and additions to the manifest spec fixture that show that this change is safe.

SMaxOwok:
I updated the commit to preserve the source identifiers.  I can try to add some tests (I had some built out), but the output of the service is hard to test because the ids change every time.  I'll try to work around that.

SMaxOwok:
@zdavis I've added some tests to illustrate the anchor assignment in the TOC post processor.  I adjusted our existing fixtures to include hashed TOC entries, so I think that the ingestion source/text section counts remaining the same illustrates your first point.  The other two are addressed in the added TOC post processor spec.

zdavis:
Just adjusting the fixture is not enough. There should still be a spec:

"It creates one text section per ingestion source, even when it is included multiple times in the manifest"

Yes, that is implicitly covered, but I'm asking you to make the specification explicit. Thank you.

zdavis:
I didn't change the spec after our discussion yesterday; please update the language, @SMaxOwok if you have a chance, and I will accept this.

SMaxOwok:
Will do

PR #1868, opened 2019-02-01
Refactor and style frontend for project hero block

dananjohnson:


zdavis:
Some quick feedback on this...

- Light hero background should be #F7F7F7, not white.
- Hero always shows the generic project thumbnail, even when the project has a thumbnail.
- Reloading, and sometimes just navigating to the project detail page triggers a 500 error in the hero component. [FIXED]
- When the project thumbnail shows in the hero, it does not respect the color choice.
- Appears that we no longer show the author image; I don't think this is correct. None of the comps examples included this, but the logic around when we show author images should have remained as is.
- CTA Button text wraps too soon (too much padding). We should be able to fit more text on one line.
- Icon size on download CTA when rendered as button is silly large
- Twitter link is wrong. Looks like hashtag URL was copied for it.

zdavis:
Just need to remove the "by" if we're showing maker avatars, per my note in slack. Once that's fixed, ready to accept.

PR #1871, opened 2019-02-04
[C] Set default TOC block depth to 6

SMaxOwok:


SMaxOwok:
Updated to include client default

PR #1889, opened 2019-02-06
[B] Set default PC number of projects to 8

### LINKED ISSUE #1851: Improve usability of "number of projects" field on project collections
This appears to default to 0, which can lead to confusing behavior. If you make a smart collection and forget to set this, the collection won't get any projects added to it because of this limit. I think we should probably set this to a sane default, or treat 0 as unset. Thoughts?

SMaxOwok:
Fixes #1851

### DIFF
--- a/client/src/backend/containers/project-collection/New.js
+++ b/client/src/backend/containers/project-collection/New.js
@@ -25,7 +25,7 @@ export class ProjectCollectionNew extends PureComponent {
   get defaultModel() {
     return {
       attributes: {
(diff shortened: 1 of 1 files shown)


SMaxOwok:
I changed this on the client only because I think that 0 and unset are still valid values to have for this.

It's probably rare that someone would want a collection with 0 items, but I think it's less confusing than setting 0 and seeing it full of projects.

Happy to discuss if anyone feels differently.

zdavis:
I agree!

PR #1900, opened 2019-02-06
[B] Fix OAuth sessions not setting cookie

### LINKED ISSUE #1898: Refresh clears OAuth login
1. Log in with google (maybe others too)
2. Refresh any page

BUG: You're logged out.

SMaxOwok:
This commit refactors the auth cookie setting into its own function.
This function responds to the `LOGIN_SET_AUTH_TOKEN` action, which sets
the auth token in state.  With this change, the auth token will also be
set in a cookie to persist login sessions.  This will fix an issue where
logging in with OAuth would not set an auth token cookie and thus
would log users out on a refresh.

Fixes #1898

### DIFF
--- a/client/src/store/middleware/currentUserMiddleware.js
+++ b/client/src/store/middleware/currentUserMiddleware.js
@@ -1,4 +1,5 @@
 import actions from "actions/currentUser";
+import cookie from "cookie";
 import { ApiClient, tokensAPI, meAPI, favoritesAPI } from "api";
 import { notificationActions } from "actions";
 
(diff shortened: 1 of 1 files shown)


SMaxOwok:
There is a side-effect to this change:
>Every time a user refreshes or logs in through an existing auth token, it will change the expiration date of the cookie.  

If this is ok, then this should be good to go.
If not, this can be made more flexible and not set the cookie if one is present or a username/password was the login method.

zdavis:
When a user authenticates with a token, I don't think we should update the expiration date on the cookie.

PR #1902, opened 2019-02-06
[B] Fix backend event tile styles

SMaxOwok:
This is a temporary fix.  This bug is caused by the .vertical-list-primary
styles, which should be refactored to reflect our new styling conventions.
Completing that refactor should resolve the styling issues on the
event tiles.


SMaxOwok:
I spoke with @dananjohnson and we agreed that the best way to solve this right now was to add `!important` to styles we needed in event tiles that are being overwritten by `.vertical-list-primary` styles.  I've opened #1903 to reflect that.

PR #1906, opened 2019-02-07
[F] Add default content blocks on project creation

SMaxOwok:


zdavis:
Update mechanism should only operate on projects that have zero content block associations.

PR #1907, opened 2019-02-07
[C] Cleanup errors/warnings from Rails 5.0 -> 5.1 update

SMaxOwok:


SMaxOwok:
```
DEPRECATION WARNING: raise_on_unfiltered_parameters is deprecated and has no effect in Rails 5.1. (called from <top (required)> at /Users/mono/src/manifold/api/config/environment.rb:5)

DEPRECATION WARNING: ActiveSupport.halt_callback_chains_on_return_false= is deprecated and will be removed in Rails 5.2. (called from <top (required)> at /Users/mono/src/manifold/api/config/initializers/callback_terminator.rb:6)
```

PR #1945, opened 2019-02-15
[F] Add unsubscribe button to backend users edit

SMaxOwok:


SMaxOwok:
Also note that this changes the drawer utility buttons to always be stacked with text visible.

SMaxOwok:
screen shot 2019-02-15 at 11 16 53 am

PR #1962, opened 2019-02-28
[F] Add prompt and option instructions to Radios

SMaxOwok:
screen shot 2019-02-28 at 10 38 05 am


SMaxOwok:
@dananjohnson can you take a quick pass over the styles I added/adjusted for this and make whatever adjustments you think are improvements? 

I wasn't 100% on the best way to update the existing styles and use BEM for the new ones.

PR #1992, opened 2019-03-07
[WIP] Searchkick results

SMaxOwok:
- Added guards to `parents_for` methods in SearchResultsSerializer.
	- This works alright, but leads to client errors rendering each result item.

- Added guards to client components when relevant parent attributes are missing.
	- This works and resolves the above point, but can lead to states where the list is very misleading (empty pages, visibly incorrect items vs. count, etc.).

- Several changes to move SearchableNode indexing to Texts
	- As talked about, moves away from TextSections being editable.

- Filter SearchResults to remove those where `model` is nil.
	- Not sure I found the correct place to do this.  Results were still inconsistent, seemed to still need the serializer guards.

- Tried implementing `Searchkick::QueueProcessJob` to bulk reindex.
	- This seemed promising as it sped things up a bit, but it was hard to implement in a way that didn't fire off just as many as these as the normal SearchReindex jobs.
	- I think this is typically done through a timed job.

- Tried adjusting `TextSection::DestroySearchableNodes` to bulk reindex `SearchableNodes`.
	- Seemed to work, but also had inconsistent results.

Notes:
  - searchkick callbacks: :async enqueues a `Searchkick::ReindexV2Job` for each item
  - searchkick callbacks: :queue relies on a `Searchkick::QueueProcessJob` with the class name
  - `TextSections::ReindexSearchableNodes` resets `SearchableNode` records for the text section, then creates new ones and reindexes them. This uses `.reindex`, so it's performed in the same job.

### DIFF
--- a/api/app/serializers/search_result_serializer.rb
+++ b/api/app/serializers/search_result_serializer.rb
@@ -41,6 +41,8 @@ class SearchResultSerializer < ApplicationSerializer
   end
 
   def parents_for_text_section_child
+    return {} unless object.model.text_section.present?
+
     text_section = object.model.text_section
     {
       text_section: text_section_properties(text_section),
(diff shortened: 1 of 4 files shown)


SMaxOwok:
@zdavis is building on this stuff in #1999

Issue #1958, opened 2019-02-25
500 error when searching on CUNY instance for "Elaine"


jojokarlin:
When I type "Elaine" into the main search bar on cuny.manifoldapp.org, Manifold returns a 500 error.

## Expected Behavior

If the term exists in Manifold, those items should appear. If it is nowhere in any of the texts, Manifold should return a notice that the term can't be found. 

## Current Behavior

```
I, [2019-02-25T18:35:58.421650 #20181]  INFO -- : [c530f2c4-5309-4e31-81ac-7207f3709e69] Started GET "/api/v1/search_results?keyword=Elaine&scope=&search_num=1&all_facets=true&project=&text=&text_section=&page%5Bnumber%5D=1" for 65.157.26.82 at 2019-02-25 18:35:58 +0000
I, [2019-02-25T18:35:58.423434 #20181]  INFO -- : [c530f2c4-5309-4e31-81ac-7207f3709e69] Processing by Api::V1::SearchResultsController#index as HTML
I, [2019-02-25T18:35:58.423583 #20181]  INFO -- : [c530f2c4-5309-4e31-81ac-7207f3709e69]   Parameters: {"keyword"=>"Elaine", "scope"=>"", "search_num"=>"1", "all_facets"=>"true", "project"=>"", "text"=>"", "text_section"=>"", "page"=>{"number"=>"1"}}
I, [2019-02-25T18:35:58.639826 #20181]  INFO -- : [c530f2c4-5309-4e31-81ac-7207f3709e69] [active_model_serializers] Rendered ActiveModel::Serializer::CollectionSerializer with ActiveModelSerializers::Adapter::JsonApi (6.68ms)
I, [2019-02-25T18:35:58.640403 #20181]  INFO -- : [c530f2c4-5309-4e31-81ac-7207f3709e69] Completed 500 Internal Server Error in 217ms (ActiveRecord: 68.0ms)
F, [2019-02-25T18:35:58.641439 #20181] FATAL -- : [c530f2c4-5309-4e31-81ac-7207f3709e69]
F, [2019-02-25T18:35:58.641536 #20181] FATAL -- : [c530f2c4-5309-4e31-81ac-7207f3709e69] NoMethodError (undefined method `name' for nil:NilClass):
F, [2019-02-25T18:35:58.641577 #20181] FATAL -- : [c530f2c4-5309-4e31-81ac-7207f3709e69]
[...]

PR #1995, opened 2019-03-08
[C] Improve searchable node indexing

SMaxOwok:
Related to #1992 and #1958 

I think this is a slight, stable, improvement to searchable node generation. The main difference is the bulk queue of reindexing searchable nodes when a text section is destroyed.

Additionally some things have been renamed to be a little clearer of what they're actually doing.


zdavis:
Related to #1997

SMaxOwok:
@zdavis is building on this stuff in #1999

PR #1996, opened 2019-03-08
[F] Implement full reader dark scheme

### LINKED ISSUE #1696: Reader pagination should respect dark mode
Opinion: reader pagination should have darker version when in dark mode.  It's kind of intense since the footer is dark too.

screen shot 2018-11-29 at 3 10 12 pm

SMaxOwok:
Closes #1696

### DIFF
--- a/client/src/reader/components/Header/index.js
+++ b/client/src/reader/components/Header/index.js
@@ -95,21 +95,13 @@ export default class Header extends Component {
   };
 
   render() {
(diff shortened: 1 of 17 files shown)


SMaxOwok:
See what you all think of the colors. Can easily adjust.

zdavis:
Let's limit this to the header, text, the prev/next buttons, and the category bar. For now, let's leave the various drawers and menus in their current color.

PR #2003, opened 2019-03-12
[B] Use span tags for list item titles

### LINKED ISSUE #2002: Backend list item text should not be wrapped with h3s
List of subjects in subjects page is a list of headings, but these elements should not be headings as they do not contribute semantic information

SMaxOwok:
Fixes #2002

### DIFF
--- a/client/src/backend/components/subject/ListItem.js
+++ b/client/src/backend/components/subject/ListItem.js
@@ -21,7 +21,7 @@ export default class SubjectListItem extends PureComponent {
         <Link to={lh.link("backendSettingsSubject", subject.id)}>
           <header>
             <div className="meta">
-              <h3 className="name large">{attr.name}</h3>
+              <span className="name large">{attr.name}</span>
             </div>
           </header>
         </Link>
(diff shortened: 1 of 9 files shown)


client/src/backend/components/subject/ListItem.js
```diff
@@ -21,7 +21,7 @@ export default class SubjectListItem extends PureComponent {
         <Link to={lh.link("backendSettingsSubject", subject.id)}>
           <header>
```
zdavis: Seems like this <header> tag is also semantically incorrect and should be removed.

client/src/backend/components/user/__tests__/__snapshots__/ListItem-test.js.snap
```diff
@@ -29,13 +29,13 @@ exports[`Backend.User.ListItem component renders correctly 1`] = `
       <div
         className="meta"
       >
-        <h3
+        <span
```
zdavis: This is wrapped in a <header> tag as well, higher up.

client/src/backend/components/user/ListItem.js
```diff
@@ -56,9 +56,9 @@ export default class UserListItem extends PureComponent {
               )}
             </figure>
             <div className="meta">
-              <h3 className="name large">
+              <span className="name large">
```
zdavis: Header tag on line 40 should be removed as well?

SMaxOwok:
Looks like all of our vertical list item components are using these tags/styles.  <= Note to self

PR #2005, opened 2019-03-12
[B] Remove secondary skip-to links

### LINKED ISSUE #2004: Remove nested skip-to links
This page has a second skip nav between the main content area and the banner. This deducts from the functional accessibility of the page because it adds extra tabstops for keyboard users which are not needed. As Hadi would say, this is too helpful

SMaxOwok:
Fixes #2004


SMaxOwok:
skip

SMaxOwok:
Adjust styles, squarer or from bottom.

SMaxOwok:
@laelito @zdavis updated skip to nav styles

zdavis:
I'm Ok with this. Lael, for context—this will pop up from the lower left when the user focuses on the skip nav (by pressing tab). Most users won't ever see this. For an example, go to http://www.washington.edu/ and press the tab key.

zdavis:
Moved it back to the top left and gave it a little opacity.

PR #2032, opened 2019-03-15
[B] Improve manifest start section identification

### LINKED ISSUE #2029: Improve manifest start_section implementation
Given a manifest TOC like this:
```
toc:
  - label: Google Doc Demo
    source_path: some/url
    children:
      - label: Hartigan HTML
        source_path: sharing-digitally.html#second
        start_section: true
        children:
          - label: 1. The Eve of War
            source_path: wow-ch01.md
```

The reader will always start from the first whole source path.  This means that even if `Hartigan HTML` was in the first position, the reader will fetch `Google Doc Demo` as the first section.  

Unclear off hand where this is failing, but I think it's in the spine generation.

SMaxOwok:
This commit adds several improvements to start section identification,
primarily affecting manifest style ingestions.
- Handles identification when start section source is referenced multiple
times in TOC.
- Handles identification when start section source contains an anchor.
- Handles identification when start section is nested under another TOC
item.

Fixes #2029


api/app/services/ingestions/strategy/manifest/text_section.rb
```diff
@@ -19,7 +19,8 @@ def attributes
             source_identifier: source_identifier,
             name: name,
             kind: kind,
-            position: @position
+            position: @position,
+            full_source_identifier: Digest::MD5.hexdigest(@source[:source_path])
```
SMaxOwok: This is my solution for preserving the full path and matching it to the manifest's `start_section_identifier`.  Open to ideas for other approaches.

zdavis:
Discussed with Max on Slack and he's going to take another pass at it.

I think that the specs should show that:

- The correct start section is assigned regardless of where it’s specified in the manifest TOC
- If the same source exists multiple times in the manifest (with or without hashes) but only one is set as the start section, it still gets set correctly
- That the start section is also set correctly for epubs

SMaxOwok:
I think our existing tests for manifest ingestions actually spec'd the correct behavior, but had a failing value. 
https://github.com/ManifoldScholar/manifold/pull/2032/commits/3972b8ce191420f08df88cd596e3077ddb296cae#diff-0f522a79c77370665d89bfac315a5aaeR64

This test should reflect that ingestion sources are not duplicated in the text section attributes.  This is now the behavior of the manifest ingestion, but wasn't previously.

api/app/services/ingestions/compilers/text_section.rb
```diff
@@ -39,7 +39,7 @@ def adjusted_attributes
           hash[:ingestion_source] = ingestion_source
           hash[:source_body] = source_body
           hash[:stylesheets] = user_stylesheets + stylesheets
-        end.except(:build, :stylesheet_contents)
+        end.except(:build, :stylesheet_contents, :full_source_identifier)
```
zdavis: This revision isn't necessary. Please remove. Thanks!

PR #2039, opened 2019-03-18
[B] Remove italic styling from hero image credits

### LINKED ISSUE #2016: Image credits and italics
Having the image credits set to render in italics makes it difficult to refer to titles, names, works, etc. that should appear in italics. In the following, `Woman at a Window` should be set apart from the surrounding text. If reverse italics isn't possible, I'd suggest we change the general treatment here from italic to roman.

credits

SMaxOwok:
Fixes #2016

### DIFF
--- a/client/src/theme/Components/browse/project/_hero.scss
+++ b/client/src/theme/Components/browse/project/_hero.scss
@@ -424,6 +424,10 @@
     color: $neutral75;
     letter-spacing: 0.012em;
 
+    em {
+      font-style: normal;
+    }
+
     .project-hero--dark & {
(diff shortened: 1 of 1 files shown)


SMaxOwok:
Also dropped the font-size down a px as it looked a little too big without italics.

Screen Shot 2019-03-18 at 10 30 32 AM

zdavis:
I'd prefer an approach here that styled Italics as Roman text rather than removed them. Waiting for @tsmyre to sign off on this approach.

tsmyre:
Reverse italics works perfectly from my end. I'm good with that.

PR #2040, opened 2019-03-18
[B] Ignore attachment changes in resource versions

### LINKED ISSUE #2000: Exclude versions without changes in API response
In the versions list item component we have a guard:
```
    if (isEmpty(version.attributes.objectChanges)) return null;
```

This leads to items missing from the list, making the pagination per page look incorrect.  We should have a scope in the API that returns only versions where `object_changes` is not empty.

SMaxOwok:
Fixes #2000

### DIFF
--- a/api/app/models/resource.rb
+++ b/api/app/models/resource.rb
@@ -12,7 +12,10 @@ class Resource < ApplicationRecord
                   meta: {
                     parent_item_id: :project_id,
                     parent_item_type: "Project"
-                  }
+                  },
+                  skip:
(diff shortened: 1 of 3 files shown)


SMaxOwok:
As it turned out, we were adjusting `object_changes` in the serializer and then rendering null for any items without `object_changes`.  The effect of this adjustment was removing changes that occur on attachment columns (which we skip for the project itself).  The fix I went with here is to skip changes to resource attachment columns, so versions are not generated in those cases.  Then the only thing we remove from the changes in the serializer is the `updated_at` key.  

If we see versions appear with empty changes, we can investigate and continue adjusting, but I believe that those cases were limited to attachment changes.

PR #2067, opened 2019-03-22
[B] Fix drawer list items not truncating

### LINKED ISSUE #2052: Reader resource drawer scrolls horizontally with long titles
drawer

SMaxOwok:
Fixes #2052

### DIFF
--- a/client/src/theme/Components/backend/_vertical-list.scss
+++ b/client/src/theme/Components/backend/_vertical-list.scss
@@ -185,6 +185,7 @@
     .content {
       display: flex;
       flex-grow: 1;
+      min-width: 0;
     }
 
     .figure-wrapper,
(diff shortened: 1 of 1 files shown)


zdavis:
Can you say a few words about why/how this fix addresses the problem? It's not at all obvious why removing padding and adding min-width 0 is a solution.

SMaxOwok:
@dananjohnson found the solution in the `.long-and-truncated-with-child-corrected` example from https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/

zdavis:
Cool. I'll leave this open so I can work it into my list refactor and then close it.

SMaxOwok:
Will be fixed in #2075

PR #2090, opened 2019-04-01
[R] Refactor DrawerHeader buttons

SMaxOwok:


zdavis:
Could you please thoroughly check all uses of this component to ensure the buttons have the correct color. I gave the backend a cursory pass, and I see at least one inconsistency:

image

SMaxOwok:
@zdavis Should be good now

PR #2094, opened 2019-04-01
[B] Fix missed list search usages

SMaxOwok:


zdavis:
This will be addressed by zd/2085

PR #2096, opened 2019-04-02
[F] Add empty message to entities list

SMaxOwok:


client/src/backend/components/list/EntitiesList/List/Empty.js
```diff
@@ -0,0 +1,18 @@
...
+
+export default class ListEntitiesListEmpty extends PureComponent {
+  static displayName = "List.Entities.List.Empty";
+
+  static propTypes = {
+    message: PropTypes.node
+  };
+
+  get message() {
+    return this.props.message || "Sorry, no results were found.";
```
zdavis: Seems to me that the default prop should be specified using defaultProps in this case. I will fix.

Issue #2110, opened 2019-04-03
Design missing SVG icons


dananjohnson:
We need design assets for these icons to be able to finish replacing all icon font instances with SVG components. I'll update this list as I discover any other missing icons.

- [x] checkmark (16)
![Screen Shot 2019-04-02 at 3 07 44 PM](https://user-images.githubusercontent.com/12454203/55514780-d63ef000-561d-11e9-846e-6f874a3aaf6a.png)

- [x] ~bold close (16)~
![Screen Shot 2019-04-03 at 2 29 30 PM](https://user-images.githubusercontent.com/12454203/55514740-bd363f00-561d-11e9-95ab-909bcb4002e9.png)

- [x] zoom (see [resource collection slideshow](https://edge.manifoldapp.org/projects/the-adventures-of-sherlock-holmes/resource-collection/node-210631))
![Screen Shot 2019-04-12 at 2 58 32 PM](https://user-images.githubusercontent.com/12454203/57035517-1edece80-6c07-11e9-95f9-4e1086ed1395.png)

- [x] play outline (16) (see [resource collection](https://edge.manifoldapp.org/projects/the-adventures-of-sherlock-holmes/resource-collection/node-210631))
![Screen Shot 2019-04-12 at 3 38 57 PM](https://user-images.githubusercontent.com/12454203/57035581-42a21480-6c07-11e9-9d5f-58e9213a4aad.png)

- [x] eye open (16) (see [resource collection](https://edge.manifoldapp.org/projects/the-adventures-of-sherlock-holmes/resource-collection/node-210631))
![Screen Shot 2019-04-12 at 3 39 52 PM](https://user-images.githubusercontent.com/12454203/57035599-4d5ca980-6c07-11e9-8d51-a911010fabb8.png)

- [x] play (see [resource audio player](https://edge.manifoldapp.org/projects/ivanhoe/resource/outer-limits-audio-intro-e60a4111-2cc7-4716-82bc-9ee3cb7b5b1d))
![Screen Shot 2019-05-01 at 11 34 23 AM](https://user-images.githubusercontent.com/12454203/57035369-aed04880-6c06-11e9-8ccb-bcfd74e3bd71.png)

- [x] pause (see [resource audio player](https://edge.manifoldapp.org/projects/ivanhoe/resource/outer-limits-audio-intro-e60a4111-2cc7-4716-82bc-9ee3cb7b5b1d))
[...]

PR #2113, opened 2019-04-04
[R] Replace all icon font instances with SVG icons

dananjohnson:
This PR replaces nearly all remaining icon font instances in the client with SVG icons, which are rendered using the IconComposer component. It also introduces BEM conventions to nearly all contexts where the icons have been updated, and updates CSS to handle sizing and appearance changes related to the new icon format. Only icons still needing design (#2110) remain unimplemented.

In addition to icon changes, this PR refactors a number of commonly used CSS classes using the BEM approach in order to reduce selector specificity, improve stylesheet readability, and remove visual inconsistencies caused by how styles had been scoped in different contexts. This includes drawer and overlay headers, form elements, and most global button classes (`.button-primary`, `.button-secondary`, `.button-bare-primary`, `.button-icon-primary`, `.button-icon-secondary`, etc.). 

A few components have undergone more substantial refactoring. In particular, the DetailHeader component on the backend of the client has been rewritten using a new `.backend-header` class that styles headers across the backend in a consistent manner. CSS for the mobile site nav menu has also been refactored to clean up redundant code and fix visual inconsistencies.

Lastly, a handful of stale components and SCSS partials have been removed from the repository.


zdavis:
👏 👏 

Can we remote the icon font, or are we still waiting on @laelito for some icons?

dananjohnson:
There are a handful of icons we still need design for (#2110), mostly in the reader. I'll go through after and remove the font and any remaining `.manicon` styles in the CSS after that. 🙌

zdavis:
We also get to remove the webpack fontgen loader and some package
dependencies. This will also address the issue where installing node
packages hangs for a while, due to a bug in the font gen module.

PR #2115, opened 2019-04-04
[F] Add publish toggle to feature list

### LINKED ISSUE #2105: Toggle Feature publication from list view
Presently, to swap out the Feature on the site, I need to go into the feature that is now live and toggle it's published slider off and then nav to the feature I want live, enter its record, and move that slider to published. It would be nice if those sliders were exposed on the list of all of features for an instance so they could be changed directly from that view.

### LINKED ISSUE #2130: Hidden content blocks are visible on the FE

SMaxOwok:
Fixes #2105
Fixes #2130


zdavis:
I'm making some revisions to this. Not liking how the label + switch looks here.

PR #2124, opened 2019-04-05
[B] Remove duplicate paths from icon svgs

zdavis:


client/src/global/components/icon/64/ArrowBack.js
```diff
@@ -0,0 +1,81 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import classnames from "classnames";
+
+export default class ArrowBack extends Component {
```
SMaxOwok: I believe this is the icon we just removed from icons and comps

zdavis: It hasn't been removed from sketch yet, though, and ideally the state of Icons in the app matches exactly what's in the sketch file—avoiding manual adjustments means we can do full exports at any time. I do realize that there's an in-between period in which we've removed an icon from usage, but it's still available—that's just something we will have to be cautious about until we get the next iteration of the sketch file. If only the design files lived in SCM, this would be easier.

PR #2141, opened 2019-04-11
[B] Standardize FE project list markup

dananjohnson:
* Fixes inconsistencies in markup and class names across project and project collection lists
* Removes use of "additionalClasses" prop in favor of fixed class names
* replaces icon font instances with SVG icons and updates SCSS accordingly


dananjohnson:
@zdavis this PR now adds a default `<a>` color.

PR #2154, opened 2019-04-23
[F] Refactor footer for use with plugin replacement components

MollyJeanB:


zdavis:
Thanks for this PR. Before I can accept this, a few things need to be addressed:

1. The CI build is failing, which is probably because your changes require snapshots to be updated. Check with Dana or Max on how to address this.
2. The last commit messages on the last commit, 2878f09, is too long. 50 chars max, please. See https://chris.beams.io/posts/git-commit/
3. It's not clear to me why you need to introduce state in the footer component. I suspect it's not necessary, and that it's not the right solution to whatever you're trying to do. Explain your thinking to me and I can help you come up with a different solution.
4. The title of this PR should follow our PR conventions. This would classify as a feature, since you're adding new functionality to Manifold, so the PR should begin with [F]. PR titles flow into Manifold's changelog.

zdavis:
Hey @MollyJeanB I'm ready to accept this, but there are now a few merge conflicts—could you please address these? Thanks!

MollyJeanB:
@zdavis I've addressed the conflicts and it's ready for re-review

PR #2163, opened 2019-04-30
[B] Fix hidden resource thumbnails

### LINKED ISSUE #2144: ResourceThumbnails component hides all at narrower widths
I'm not totally sure offhand what is truncating these, but if there are fewer than 3 resources, the thumbnails don't appear on narrower screens.  It ends up leaving a weird space.

resource

ConnorChristensen:
Fixes #2144

### DIFF
--- a/client/src/theme/styles/components/frontend/project/_resource-thumbnail-list.scss
+++ b/client/src/theme/styles/components/frontend/project/_resource-thumbnail-list.scss
@@ -3,6 +3,7 @@
     @include listUnstyled;
     display: flex;
     flex-wrap: wrap;
+    // push it 20px to the left to negate the left padding on the resources
     margin-left: -20px;
   }
 
@@ -12,20 +13,8 @@
     padding-left: 20px;
     margin-bottom: 20px;
 
-    &:nth-last-of-type(-n + 2) {
-      display: none;
-    }
-
-    @include respond($break65) {
-      width: 25%;
-    }
-
-    @include respond($break90) {
+    @include respond($break80) {
       width: 20%;
-
-      &:nth-last-of-type(-n + 2) {
-        display: flex;
-      }
     }
   }
 


client/src/theme/styles/components/frontend/project/_resource-thumbnail-list.scss
```diff
@@ -1,30 +1,57 @@
...
-      width: 20%;
+      li {
+        width: auto;
+        // remove the width and padding from the fallback
+        padding: 0;
+      }
+
+      @include respond($break65) {
+        // make auto generated rows 0px tall
+        grid-auto-rows: 0;
```
dananjohnson: This is a really cool technique! 🎉

client/src/theme/styles/components/frontend/project/_resource-thumbnail-list.scss
```diff
@@ -1,30 +1,57 @@
...
+      // use grid gap instead
+      grid-gap: 20px;
+      grid-template-rows: repeat(2, auto);
+      grid-template-columns: repeat(2, 1fr);
+      // remove the margin from the fallback
+      margin-left: 0;
 
-    @include respond($break90) {
-      width: 20%;
+      li {
```
dananjohnson: Make sure all the `li` elements have matching heights.

Screen Shot 2019-05-01 at 9 22 38 AM

client/src/theme/styles/components/frontend/project/_resource-thumbnail-list.scss
```diff
@@ -1,30 +1,57 @@
...
     }
 
-    @include respond($break65) {
-      width: 25%;
-    }
+    // grid style if the browser supports it
+    @supports ( grid-area: auto ) {
+      display: grid;
+      // use grid gap instead
+      grid-gap: 20px;
```
dananjohnson: The margin below the list was lost switching to grid, so make sure to add it back. Otherwise the spacing beneath resources and resource collections becomes inconsistent.

Screen Shot 2019-05-01 at 9 27 13 AM copy

client/src/theme/styles/components/frontend/project/_resource-thumbnail-list.scss
ConnorChristensen: Thanks!

client/src/theme/styles/components/frontend/project/_resource-thumbnail-list.scss
ConnorChristensen: Good catch. While I was looking around at that I also found the grid was clipping the shadows on the left side. Fixed that as well.

dananjohnson: Good catch of yours too! It looks like that padding isn't quite sufficient to prevent clipping the shadow on the bottom of the items though.

Screen Shot 2019-05-01 at 4 15 15 PM

ConnorChristensen: I can't easily keep the grid from clipping the shadows without pushing the `Total Resources` tag further away than the 50px at which it currently sits. I could squeeze a little more space if I remove the top margin and padding on the total resources, but that would require a custom rule for that one element.

dananjohnson: And I'm seeing that there's clipping on the righthand side as well.

You're right that we should avoid writing one-off rules for different elements. Which leads me to think that in spite of very cool technique to hide overflowing grid items, we probably need to fallback to the other solution you proposed originally, of just dropping that 4-column intermediate layout. The shadow here just doesn't play well with `overflow: hidden`, and I don't think it's worth the CSS workarounds to resolve it.

ConnorChristensen: No worries. Often times the simplest solution is the best. That grid would have worked great if not for the shadows.

client/src/theme/styles/components/frontend/project/_resource-thumbnail-list.scss
```diff
@@ -12,20 +13,12 @@
     padding-left: 20px;
     margin-bottom: 20px;
 
-    &:nth-last-of-type(-n + 2) {
-      display: none;
-    }
-
     @include respond($break65) {
```
dananjohnson: Let's drop this breakpoint altogether, so the width goes straight from 50% to 25% at `$break90`.

client/src/theme/styles/components/frontend/project/_resource-thumbnail-list.scss
dananjohnson: Yeah exactly. And it's easy to miss one (or more) of the visual states an element might have before creating a solution.

client/src/theme/styles/components/frontend/project/_resource-thumbnail-list.scss
ConnorChristensen: It looks a little weird at 50% right after the breakpoint. Its about a 1:3 aspect ratio.
Screen Shot 2019-05-02 at 1 42 25 PM

dananjohnson: OK, let's set the breakpoint to `$break80` rather than `$break90`.

dananjohnson:
@zdavis this looks good to me

Issue #2068, opened 2019-03-23
Upgrade eslint config


zdavis:
Upgrade eslint-config-airbnb to 16 and fix things.

ConnorChristensen:
@zdavis 
Should I still use v16, or should I use v17 which is the newest one?

zdavis:
V17. Consult with max and Alexa if there are rules that require difficult
code changes, as we may choose not to enforce them.

Zach

zdavis:
This has been done. `eslint-config-airbnb` is currently at v18 in manifold. There are rules we should consider enabling and addressing, but I think this issue can be closed.

PR #2164, opened 2019-05-03
[C] Update eslint-config-airbnb to v17

ConnorChristensen:
## Linting
This pull request resolves issue #2068.

28 commits, 4 days, 677 files changed, ~734 errors fixed.

Almost all of them are syntax and file structure changes, and not many of them have a chance of altering the functionality of the code. Each commit was tagged with the linting error that it fixed. 

## react/sort-comp
Sort-comp is broken into part 1 and 2. Part 1 was automated and part 2 had to be done manually.

## react/no-access-state-in-setstate
react/no-access-state-in-setstate is the most likely to have introduced a bug, so they are broken up into three commits. 

* **Easy** - very simple changes that are likely to be much safer
* **Medium** - more complicated and should be carefully reviewed to make sure they have the same effect
* **Commented** - difficult ones that I didn't bother to alter and just turned off the linter. They can be converted, but only with a good reason to do so.

## Additional notes
* Around line 35 of `client/src/frontend/components/resource/Link.js` on `[C] Fix jsx-a11y/anchor-is-valid` I changed the download buttons so that it actually downloads the resources instead of opening a new tab with the content.
* I leaned more towards making commits often with the intention of squishing them later.
* I made some commits that introduced missing brackets and commas in the sort-comp section. These were all corrected in the `Minor style changes and fixes`. If you run a test in sort-comp, it will fail.
* Of course, this changes lots of files in very big ways, so it is likely to cause collisions with other branches being merged in if anyone altered the JS. The SCSS is pretty much untouched.


client/src/backend/containers/permission/Edit.js
```diff
@@ -64,7 +64,7 @@ export class PermissionEdit extends PureComponent {
   };
 
   removeAllPermissions = () => {
-    event.preventDefault();
+    window.event.preventDefault();
```
zdavis: This line should just be removed, or even should be added as an argument on removeAllPermissions handler if it's responding to an event.

client/src/reader/components/section/Text.js
```diff
@@ -104,6 +104,7 @@ export default class Text extends Component {
   // filteredAnnotations), but is in the unfiltered collection, then we can remove
   // it from the URL because the user has filtered it off the screen for now.
   checkRequestAnnotationHash() {
+    const { location } = this.props;
```
zdavis: This change requires some testing. Are you sure that this.props.location always === window.location?

client/src/store/middleware/oauthMiddleware.js
```diff
@@ -17,7 +17,7 @@ export default function oauthMiddleware({ dispatch, getStateIgnored }) {
   return next => action => {
     switch (action.type) {
       case `${O.oauthResponse}`:
-        tryToSignIn(dispatch, event);
+        tryToSignIn(dispatch, window.event);
```
zdavis: I'd want to double check that  window.event is the intention here. I'd see if Alexa has insight on this.

client/src/api/LowLevelApiClient.js
```diff
@@ -43,13 +43,15 @@ export default class LowLevelApiClient {
...
     const out = fetch(endpoint, fetchConfig).catch(error => {
       return new Promise((resolve, reject) => {
-        reject({
-          response: {
-            status: 503,
-            statusText: error.name,
-            exception: error.message
-          }
-        });
+        reject(
```
zdavis: We need to test that API failures are still resolved correctly following this change. Move this commit into a separate PR so it can be tested more carefully.

client/src/reader/containers/annotation/annotatable-components/Drawer.js
```diff
@@ -33,22 +24,12 @@ export default class AnnotatableDrawer extends PureComponent {
     return upperFirst(this.props.drawerState);
   }
 
-  setDrawerWrapperProps = drawerWrapperProps => {
```
zdavis: Please explain why this component is being changed in this way.

client/src/reader/components/notation/viewer/Group.js
```diff
@@ -30,7 +30,7 @@ export default class NotationViewerGroup extends PureComponent {
 
   constructor(props) {
     super(props);
-    this.state = { defaultActive: null };
+    this.state = {};
```
zdavis: Does this.state need to be set at all?

zdavis:
See my comments. Let's pull the prefer-promise-reject-errors change into a separate PR so I can just double check it

client/src/reader/components/notation/viewer/Group.js
ConnorChristensen: No it does not. Thank you.

client/src/reader/containers/annotation/annotatable-components/Drawer.js
ConnorChristensen: According to eslint:
`Unused state field: 'drawerWrapperProps'  react/no-unused-state`

Looking at the file, I could not find any reference to the `drawerWrapperProps` and it seemed safe to delete it.

Upon closer inspection, the drawer wrapper props were intended to be passed into the `Drawer.Wrapper` as `<Drawer.Wrapper {...drawerWrapperProps}>` instead of the current `<Drawer.Wrapper {...drawerProps}>`.

client/src/api/LowLevelApiClient.js
ConnorChristensen: Will do

client/src/reader/components/section/Text.js
ConnorChristensen: I double checked some documentation and it seems like it does, (especially with react-router) but I'll go with `const location = window.location` to be safe and clear.

client/src/store/middleware/oauthMiddleware.js
ConnorChristensen: Looks like we are good with this one and it should do exactly that

client/src/backend/containers/permission/Edit.js
ConnorChristensen: So it's safe to just delete this line? I'm not sure what type of event you would like to have trigger a preventDefault.

client/src/reader/containers/annotation/annotatable-components/Drawer.js
```diff
@@ -33,22 +24,12 @@ export default class AnnotatableDrawer extends PureComponent {
...
-    this.props.actions.closeDrawer();
-  };
-
   render() {
     const DrawerContents = Drawers[this.componentKey];
 
     let drawerProps = {
       style: "frontend",
       open: this.isOpen,
-      closeCallback: this.closeDrawer,
```
zdavis: Why remove this?

zdavis:
@ConnorChristensen I would like very much to accept this PR, but in its present form it introduces at least a few bugs that I don't have time to fix right now. I think many of the bugs are coming from the `[C] Fix react/no-unused-state`, so please remove that commit from this set. You'll have to deal with some merge conflicts when you remove it.

Some changes seem arbitrary to me, and I need to understand why they were made. In the commit "[C] Get location from window instead of props," you've made a decision to access `location` from window instead of props. Why? This seems wrong to me. In that `[C] Fix react/no-unused-state`, you're removing a closeCallback that's being passed to a child component. Why?

I'm not confident enough in this PR to accept it given that some of the changes I'm seeing don't make a lot of sense. I think we need to come at this a little differently to make the process easier. I want you to backup a bit and open a PR that updates ESLint (and associated modules) while making _minimal_ changes to the code base. In other words, update ESLint and then disable the new rules that are causing warnings/errors.

We can get that baseline update into Master, and then you can open smaller PRs that are more manageable and safer for me to accept. A lot of these commits are really straight forward, no-brainers and we can tackle those quickly, and then come back to the more complex changes where there's more risk of introducing bugs.

Make sense?

Zach

ConnorChristensen:
Yes that makes sense. It looks like the cases where `react/no-unused-state` causes the most problems is when a whole object is passed in to the render method and the linter can't figure out that it is actually being used. This is probably something that ESLint will have to improve over time.

I ran tests after every lint change I made, but I understand now that is by no means a guarantee against bugs and I need to be much more cautious.

On the commit "[C] Get location from window instead of props" that was changing back to what it was in its original form. You made a comment earlier that it got changed over to props instead of window and I was putting it back. Adding a commit to undo a bug introduced earlier on is not a great solution though.

I'll go over it much more carefully this time through. Anything outside of the things that are super safe, I'll be sure to just split them into more commits with an explanation attached. Thank you for looking though these commits closely and catching things I missed. Having an expert on the codebase check my work is very helpful.

zdavis:
I wish we had greater test coverage on the frontend (or the backend), but alas, we do not—there are definitely lots of cases, especially around user interaction, where we're missing tests.

PR #2165, opened 2019-05-06
[B] Correct project collection project sorting

### LINKED ISSUE #2146: Order of projects in collections
We can order projects in collections in various ways - alphabetically, by pub date, etc. - and they show up in this order on the homepage but not on the "See the Full Collection" page. Can this be set so the default is for projects to show up in the selected order on this page too? Thanks!

zdavis:
Fixes #2146

### DIFF
--- a/api/app/models/collection_project.rb
+++ b/api/app/models/collection_project.rb
@@ -5,6 +5,7 @@ class CollectionProject < ApplicationRecord
 
   belongs_to :project_collection, counter_cache: true
   belongs_to :project
+  has_one :ranking, class_name: "CollectionProjectRanking", inverse_of: :collection_project
   has_many :creators, through: :project, source: "makers"
   has_many :contributors, through: :project, source: "makers"
 
@@ -12,6 +13,8 @@ class CollectionProject < ApplicationRecord
   acts_as_list scope: :project_collection
 
   # Scopes
+  scope :globally_ranked, -> { joins(:ranking).merge(CollectionProjectRanking.globally_ranked) }
+  scope :ranked, -> { joins(:ranking).merge(CollectionProjectRanking.ranked) }
   scope :with_order, ->(order) { order(order) if order.present? }
 
   scope :projects_with_read_ability, lambda { |user = nil|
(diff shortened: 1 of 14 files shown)


zdavis:
Hey @ConnorChristensen, could you please do a few things for me on this PR?

First, please address the lint issues above.
Second, please rebase the branch and fixup my commit into Alexa's
Third, make some project collections in the backend and confirm that this fix seems to address the issue that was reported.
Fourth, if everything looks good, add a "ready to accept" label to this PR.

Thanks!

ConnorChristensen:
I addressed the lint issues and hound is happy now.

The test on travis CI failed because it looks like a snapshot for a particular React component needs to be updated every month. I'm going to see if I can fix that.

So you want your commit squashed into Alexa's commit?

Can you point me to some more resources or examples for your third item?

zdavis:
Well, I want it fixed-up, rather than squashed, because I don't want to maintain my commit message. In case you're not familiar with the (minor) difference:

`The difference between squash and fixup is that during the rebase, the squash operation will prompt you to combine the messages of the original and the squash commit, whereas the fixup operation will keep the original message and discard the message from the fixup commit.`

As for my 3rd item, you should definitely take some time to familiarize yourself with the Manifold backend. I want you to login to your local Manifold instance, create a project collection, and confirm that the projects are sorted correctly in both contexts, as described in the bug. We did this at my desk, remember? I had a project collection in the backend and we could see that the project sorting wasn't the same on the project collection detail view and the home page. I'm asking you to check and make sure that we've addressed the original problem. If you don't have a backend user, you can create one through the console, or there is a rake task to do it in the API dir: `rake manifold:user:create:admin['email@example.com','test123!','First','Last']`. If it's still not clear what I'm asking for, let me know.

zdavis:
@ConnorChristensen Did you see my feedback above? Have you had a chance to test this changeset?

ConnorChristensen:
I did see it. Thank you for telling me about fixup. I had never heard of it before, so I went and did some research about it. It looks like the way its done is by adding a `--fixup` to a `git commit`. Since you already made the commit, and it happened before Alexa's commit, I'm confused about how you want me to fixup the commit. What I tried to do is rebase the branch, reorder the commits, so yours goes after Alexa's then do a soft reset at Alexa's commit and recommit your changes with the `--fixup` flag. This doesn't seem like the right approach though.

I ran a manual test by making a collection of resources. It looks like the bug is still present. I ran a `db:migrate` before I did the test to make sure I had everything up to date before I did the test, so I think I did the test correctly.

zdavis:
I adjusted the Project model to use Alexa's new ranking mechanism.

Thanks @ConnorChristensen @scryptmouse for your contributions on this! Yay, it's done!

PR #2176, opened 2019-05-14
Simple ESLint update

ConnorChristensen:
Partly fixes issue #2068 

This is just the most simple and safe changes that can be made while following new ESLint rules. There are other branches that contain more difficult fixes, and they can be incorporated at any time.


zdavis:
👏 Thanks, Connor! Looks great!

PR #2300, opened 2019-07-29
[B] Refactor FetchSelect

### LINKED ISSUE #2294: Grant Permissions dropdown isn't focusable
screencast 2019-07-25 14-55-29

Consider refactoring as a native `<select>`.

dananjohnson:
This commit refactors the FetchSelect component to use the native HTML `<select>` element, thus enabling keyboard interaction and improving accessibility.

Resolves #2294

### DIFF
--- a/client/src/backend/containers/form-inputs/connected-inputs/FetchSelect.js
+++ b/client/src/backend/containers/form-inputs/connected-inputs/FetchSelect.js
@@ -2,7 +2,8 @@ import React, { PureComponent } from "react";
 import PropTypes from "prop-types";
 import { connect } from "react-redux";
 import debounce from "lodash/debounce";
-import classNames from "classnames";
+import find from "lodash/find";
+import { UID } from "react-uid";
 import { ApiClient } from "api";
 import IconComposer from "global/components/utility/IconComposer";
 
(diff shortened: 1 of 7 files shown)


dananjohnson:
@zdavis I’d appreciate some feedback on this one.

There was a challenge in that previously the component had that placeholder text by default, which would go away once a selection was made. With a `<select>`, you can’t really do that; if you want a placeholder, it has to be the first option in the list, which means that your handler needs to account for if the placeholder is selected. I opted instead to just remove the placeholder, and set the first option in the list as the initial selected option. I don’t feel 100% confident in my decision, though, or in the way I implemented it, so your feedback would be very much appreciated!

dananjohnson:
@zdavis re-upping this, as it might be good to merge in alongside #2601.

zdavis:
I don't really think the solution you have in place works. It looks like the FetchSelect component is mostly used by OptionLists by way of the BelongsTo component. A good example of these components in use is on the project permissions form, where the user has to select a Manifold user to receive permission.

The current implementation looks like this:
image

With your changes, we get this:
image

If we cut over to using a standard select input here, which I'm totally fine with, we can't lose the label. We'd need to make sure the label appears above the select input, like this:
image

The easy way to do this would be to just make the placeholder prop render as the label, but in the long run that will just be confusing. Instead, you'll need to go through the various places where this component is used (maybe only through Form.BelongsTo, but maybe through other components) and change the placeholder prop to a label prop.

Make sense?

dananjohnson:
Yep, makes sense, and totally agree.

zdavis:
Rebased on master

Issue #2304, opened 2019-07-31
Revisions to standalone header feature


dananjohnson:
- [ ] apply shadow on standalone header to library header as well
- [ ] standardize background color on dropdown menus

zdavis:
Comps:
![image](https://user-images.githubusercontent.com/255909/62725578-42fe3880-b9ca-11e9-9ac4-4d4ecad0e5d5.png)

Implementation:
![image](https://user-images.githubusercontent.com/255909/62725604-51e4eb00-b9ca-11e9-9fd2-ebb4c54d0df4.png)

tsmyre:
The Standalone mode tool tip reads incorrectly at present:

> If enabled, this project will always render in standalone mode when “&mode=standalone” is appended to the URL.

The ampersand should be a question mark: `?mode=standalone`.

tsmyre:
Also, when toggled to Enabled a project does not persist the `?mode=standalone` in the URL when a user clicks on a text. For example, see this project: `https://edge.manifoldapp.org/projects/the-adventures-of-sherlock-holmes?mode=standalone`.  

If you click on any of the texts and enter the reader, the menu you see will still provide all the options available for library view: Project Home, Projects, Following. If you nav back that way, Standalone is no longer visible.

Likewise, if you return fresh to `https://edge.manifoldapp.org/projects/the-adventures-of-sherlock-holmes?mode=standalone` and click on any resource, when you arrive at the resource page the standalone suffix will be gone, though the standalone header will persist. If you use that to nav back to the project home, the standalone header will be there even though the URL suffix is not showing. However if you click refresh, you'll be out of standalone.

zdavis:
![image](https://user-images.githubusercontent.com/255909/63063602-2c684d80-beb2-11e9-9677-835cd4f1b1ab.png)

tsmyre:
I'm wondering if there should be a default header bar color when a user doesn't specify one in settings. Right now the bar turns white and is only visible on hover if no color is provided.

tsmyre:
I'm also wondering if it would be possible to have a toggle that allows a public project to be in standalone and not appear in the library. I suspect that will be a highly sought after means of conveying projects when in standalone.

dananjohnson:
> I'm wondering if there should be a default header bar color when a user doesn't specify one in settings. Right now the bar turns white and is only visible on hover if no color is provided.

Yeah there definitely needs to be a default color here. Without one specified, the BG is actually just transparent.

[...]

PR #2333, opened 2019-08-28
[B] Revise notification styles

dananjohnson:
* fixes issues with incorrect position in some cases relative to header components
* fixes style regressions for notifications in backend drawers
* makes header notifications full-width
* rewrites SCSS using BEM conventions

Issue #2304


dananjohnson:
I'm gonna do a bit more work to improve this before it's ready for review.

PR #2334, opened 2019-08-28
[F] Allow backend users to access ingestion source

### LINKED ISSUE #2177: Allow download of ingested files from the backend
## Feature Description

It would be great if users who have project creator status on a project could download the file that was ingested in Manifold from the backend.

This would make it easier for users who are collaborating to be able to share the most up to date file between themselves, and it would make it easier for someone who created the file but wants to update it to locate the most up to date version without having to find it on their computer.

## User Stories

I am a user who created a text for ingestion into Manifold. I want to update the text, but I'm not sure which computer and which folder I was working on when I created the file. I wish I could download it from Manifold instead of having to find it in my files.

I am an instructor who is collaborating with several other instructors on a textbook we are writing. We are each adding a section to one text, or a chapter to a manifest ingestion. It would be great if I could download the text or zipped file, add my work to it, and then re-ingest instead of having to email the file between the collaborators.

dananjohnson:
Resolves #2177

### DIFF
--- a/api/app/models/text.rb
+++ b/api/app/models/text.rb
@@ -72,6 +72,7 @@ class Text < ApplicationRecord
           -> { where(kind: TextSection::KIND_NAVIGATION) },
           class_name: "TextSection",
           inverse_of: :text
+  has_one :last_finished_ingestion, -> { where(state: "finished").order(updated_at: :desc) }, class_name: "Ingestion"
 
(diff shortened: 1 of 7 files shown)


dananjohnson:
@zdavis this handles the frontend portion of this feature so far

PR #2405, opened 2019-10-19
[WIP] Construct API documentation

zdavis:


zdavis:
@ConnorChristensen 

Let's cutover to this branch moving forward. I've refactored and reorganized the code in a number of ways. We can review this on Monday, but the high level changes are as follows:

1. I've removed all of your schemas and specs—so you will want to keep your branch around as reference and so you can move your work in as needed. I want the conventions I've established here to be how we move forward, and expect you to build resource definitions and resource spec files one at a time going forward. I am less concerned about the model definitions being 100% correct, and more concerned that we have passing tests and that all the routes are covered. We can talk more about what that means.

2. All the resources are defined in api/spec/api_docs/definitions/resources. These resource definition files define what a response and request for that resource looks like. We define four constants in these files: ATTRIBUTES, READ_ONLY, WRITE_ONLY, and RELATIONSHIPS. We can talk through what these do. There are various ways to override behavior, but I want you to talk to me before you go down those roads. Be aware that you should use underscore keys in these definitions, not camelcase. It's important that when we're in ruby land, we use ruby conventions.

3. All the types are defined in api/spec/api_docs/definitions/types. Check with me before adding a new type, because I saw a number of types that were really one-offs in what you had prior to this. Most of the types are self explanatory, but there are a couple to note. `attachment_styles` represents an attachment on a resource (eg, an image). `Collection` refers to a collection of resources in a relationship. `Resource` refers to a single resource relationship. `Pointer` is the `{id, type}` structure that we use to identify a related resource in both `collection` and `resource` types. `Upload` is used for requests when a file is being uploaded.

4. All specs are in `api/spec/request/api/v1`. Going forward, I don't want you to change any request specs—and I have reverted all of them to how they are in master, with a few fixes from your branch re-applied. I'd rather you added new specs in this new directory. That way I can easily see what you're doing without having to grok changes to existing specs. Note how much simpler these specs are. And notice also that I'm flattening out the specs. Comments are a good example of what I mean here. This resource can be accessed through a top-level API, and also nested under Annotations and Resource (we have a model called "resource" in manifold, which can be confusing because we use resource to talk about objects in our REST API). Rather than having a bunch of nested specs, I want us to have a single spec file that includes all the routes that deal with that resource. 

5. As a proof of concept, I've added some filters to the Projects spec. Don't worry about those for now—I needed to be sure we could handle them correctly, but I don't need you to suss out what all the filters are.

6. There are various ways to override this little API doc framework we've produced. But generally speaking, you shouldn't need to. We're aiming for convention over configuration here, which is the Rails way. 

7. I've looked carefully at how the documentation renders, and will continue to do so. I've made language consistent in various places, ensured that resource names are capitalized where they should be and not capitalized where they should not be. It's possible I've missed a spot or two, so please just let me know if something looks wrong—but generally, this is working how it should be working, so please don't make changes without discussion.

8. I've removed your READMEs, but feel free to keep them somewhere outside of the project. I can see how those were helpful to you, but they don't belong in the project repo. 

9. Most fields in the API are nullable, and I've made that the default for now. We can circle back and specify what's not nullable and what's required as we iterate over this.

10. There are still a few small cleanup tasks I'd like to do—but these are about the internal structure and behavior of these classes, and won't change the interfaces you use. I will do that this wknd or Monday.

Let's go through these together on Monday and build out a spec or two. My hope is that this new approach will allow you to focus more on what matters (accurately describing the inputs and outputs) and less on the nuts and bolts of how everything fits together in the testing framework.

Have a great weekend!

best,
Zach

zdavis:
Opening a clean PR

PR #2423, opened 2019-10-31
[B] Add workaround for Windows CSV upload bug

### LINKED ISSUE #2409: Reports of inability to upload CSV on Windows + Chrome
Error report is from a surface tablet. User reports that when trying to upload a CSV file in the backend on the bulk resource import interface, the file is not recognized.

"We’re having a weird issue here where we can’t load a CSV file for bulk upload on a Windows machine. Same file works just fine on my Mac. Works as in it loads to the backend. With PC it looks like we never selected any file of any kind."

dananjohnson:
Since we do server-side validation of CSV files, we can safely use [this workaround](https://github.com/react-dropzone/react-dropzone/issues/276#issuecomment-278357061)
for react-dropzone that allows CSV files to be uploaded in MS Edge.

Resolves #2409

### DIFF
--- a/client/src/global/components/form/Upload/index.js
+++ b/client/src/global/components/form/Upload/index.js
@@ -31,7 +31,7 @@ export class FormUpload extends Component {
       extensions: "doc, docx, txt, odt"
     },
     csv: {
-      accepts: "text/plain, text/csv",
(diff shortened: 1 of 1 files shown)


zdavis:
Rebased on master

PR #2447, opened 2019-11-15
[WIP] Document Texts API endpoints

ConnorChristensen:


zdavis:
I don't think this PR makes sense at this point. If there's work here that needs to be included in our feature/api-docs branch, please extract it and make a new PR.

PR #2450, opened 2019-11-25
[F] Add ability to export Texts as V3 Epubs

scryptmouse:
Add service infrastructure for exporting `Text` instances as V3 Epubs from Manifold.

The key supporting services are located in `Packaging::EpubV3`, with an interaction located at
`Packaging::Exportation::ExportTextToEpubV3`, which takes a `Text` on the `:text` input. There
is also a job, `Packaging::Exportation::ExportTextToEpubV3Job` that can perform the export in
the background.

A rake task can be called to export a text by its ID:
`bin/rake packaging:epub_v3:export_text[text_id]`.
It will print the path to the exported epub.


zdavis:
Rebased on master to address failing CI.

zdavis:
Rebased on master

PR #2458, opened 2019-12-01
[F] Improve API performance on large model collections

### LINKED ISSUE #2443: Poor performance serializing projects with hundreds of texts
- [x] Incorrect messaging in resource import
- [x] Incorrect field names for metadata
- [x] Reading group: https://edge.manifoldapp.org/my/groups/f4427bd2-3f81-4a96-8aa2-8a584c3b94fb/
- [x] Missing title on backend text header
- [x] Error on backend texts / peoples tab
- [x] Texts blocks without texts show when they shouldn't
- [x] Search results types contain an underscore do not render in results list
- [x] Full text nodes not rendering in search results
- [x] Includes are being passed to the error serializer which causes exception

zdavis:
Fixes #2443


zdavis:
PC tests will fail in this commit because they are address in #2463. These two PRs are heavily connected, but I'm pulling them in separately so we get two entries in the CHANGELOG.

PR #2463, opened 2019-12-01
[B] Remove pagination from BE PC containers

### LINKED ISSUE #2424: Unable to add more than 25 projects to manual collection
This was reported by Indiana: When you try to add a 26th project to a manual collection, the most recent selection gets added while the one just previously added gets unchecked and removed from the collection.

zdavis:
Fixes #2424


zdavis:
Accept only after AMS > Fast JSON API PR is accepted and only after rebasing on master.

PR #2478, opened 2019-12-09
[WIP] Integrate API docs with new serializer type system

zdavis:


zdavis:
@ConnorChristensen Please correct the commit message on b693905. It is missing correct punctuation on the extended commit message.

api/spec/api_docs/definitions/resources/project.rb
```diff
@@ -16,80 +16,10 @@ class Project
...
-          creators: Type.collection(contains: "creators"),
-          events: Type.collection(contains: "events"),
-          published_texts: Type.collection(contains: "texts"),
-          resource_collections: Type.collection(contains: "resource_collections"),
-          resources: Type.collection(contains: "resources"),
-          subjects: Type.collection(contains: "subjects"),
-          text_categories: Type.collection(contains: "text_categories"),
-          texts: Type.collection(contains: "texts"),
-          twitter_queries: Type.collection(contains: "twitter_queries")
+          pending_slug: ::Types::String,
```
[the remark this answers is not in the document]
ConnorChristensen: Yep, they are request attributes. I'll rename it to `REQUEST_ATTRIBUTES` and make appropriate changes

api/spec/api_docs/definitions/resource.rb
```diff
@@ -1,78 +1,146 @@
...
+        resource_data(
+          response_attributes(attributes),
+          response_relationships(relationships)
+        )
       end
 
-      def create_request
-        definition = make_request(__callee__, create_attributes || request_attributes)
-        transform_keys(definition)
+      def response_relationships(realtion)
```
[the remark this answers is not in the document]
ConnorChristensen: Oh, I see the confusion. I was changing the name from `relationships` to `relation` not because the input changed, but because I wanted to make the distinction between the variable `relationships` and the method `relationships`. While I understand that ruby's scoping ensures that in the method it will use the local variable `relationships`, I felt it made it more confusing than it needed to be. 

I do now see with your comment how this just switched one confusion for another. Do you have any recommendations for situations like that?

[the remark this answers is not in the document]
zdavis: I'd like you to solve this problem without my input—given my responsibilities at CIC, I don't have the bandwidth to solve problems that are this granular. I need you to be able to solve this. I can give you some guidance, however. Method names should clearly describe what the method is doing. Your explanation just doesn't make sense—if we have a method called "relationships", I would expect it to return a collection of relationships. If that method is changed to "relation," I would expect it to return a single relation. It's weird to change it to "relation" since we universally refer to these things as relationships throughout the code, right? It's weird to name a method that returns multiple things in a way that implies it only returns one thing. It's weird to change methods on a class to something confusing because you don't want a local variable to overlap. Please go through this and ensure that your changes to not make it harder to understand what's happening in this class.

api/spec/api_docs/definitions/resource.rb
```diff
@@ -1,78 +1,149 @@
...
+        resource_data(
+          response_attributes(attributes),
+          response_relationships(relationships)
+        )
       end
 
-      def create_request
-        definition = make_request(__callee__, create_attributes || request_attributes)
-        transform_keys(definition)
+      def response_relationships(relation)
```
zdavis: This method is still named poorly. The name should say what it does. In this case, it looks like it's taking a relationship (don't call things one thing in one place and another thing in another place) and type casting it somehow. If that's what it's doing, call the method `cast_relationship`.

api/spec/api_docs/definitions/resource.rb
```diff
@@ -1,78 +1,149 @@
...
+        ::Types::Hash.schema(relation)
       end
 
-      protected
+      def response_attributes(attr)
```
zdavis: See my comment above for `response_relationships`. Same problem here.

api/spec/api_docs/definitions/resource.rb
```diff
@@ -1,78 +1,149 @@
...
-          properties: {
-            data: Type.object(properties: {
-                                id: Type.id,
-                                type: Type.string(example: type),
-                                attributes: Type.object(properties: response_attributes),
-                                relationships: response_relationships,
-                                meta: Type.meta
-                              })
-          }
+      ####################################
```
zdavis: Part of why this class is now so much more confusing is that you've greatly expanded its public interfaces. When I wrote this class, it had four interfaces: resource_response, collection_response, update_request, and create_request. It now has over 15 additional public interfaces. Why? Are the necessarily public, or did you just not consider whether they were public or private? Can you make it clear which of these methods are internal implementation details, and which are part of its public interface?

api/app/services/dry_types_to_json.rb
```diff
@@ -0,0 +1,89 @@
+class DryTypesToJson
+  class << self
+    def convert(type)
+      mapping = {}
+
+      if type.respond_to?(:type)
```
zdavis: The structure of this method is still too complex for me to accept it. It don't think it would pass rubocop cyclomatic complexity linting (https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Metrics/CyclomaticComplexity). Methods and method names are how we communicate what's happening in code. Restructure this so I can look at this and easily understand what it's doing. If/else should be used sparingly and is usually a code smell. Here's a hint:

```
# Public interface to the class.
def convert(type)
  mapping = {}
  maybe_convert_simple_dry_type!(mapping, type)
  maybe_convert_multi_type!(mapping, type)
  maybe_map_enum_values!(mapping, type)

  # I have no idea from reading your code _why_ these two things are happening,
  # so I can't refactor them into methods. Can you think of method names that
  # make the intention more clear?
  mapping[Dry::Types["date_time"]] = base_date_time
  mapping[Dry::Types["bool"]] = { type: "boolean" }

  # Same complaint here. Why is a method called "convert" doing all these things
  # to the mapping object that it creates, and then not returning that mapping
  # object. Why even have this mapping object if you're just going to return it's
  # type property? The code needs to be simple, clear, and easy to understand.
  mapping[type] || type
  
end

private

def values?(type)
  type.respond_to :values
end

def meta?(meta)
  type.respond_to :meta
end

def type?(type)
  type.respond_to? :type
end

def multi_type?(type)
  type.respond_to?(:name) && split_multi_type(type).count > 1
end

def maybe_convert_simple_dry_type!(mapping, type)
  return unless type?(type)
  // do the things to mapping
end

def maybe_convert_multi_type!(mapping, type)
  return unless multi_type?(type)
  // do the things to mapping
end

def maybe_merge_enum_values!(mapping, type)
  return unless values?
  // do the things to mapping
end

def maybe_merge_meta!(mapping, type)
  return unless meta?
  // do the things to mapping
end

```

api/spec/api_docs/config.rb
```diff
@@ -10,7 +10,7 @@ def swagger_root
 
       def definitions
         defined = {}
-        ApiDocs::Definition::Resource.constants.each do |resource_class|
+        ApiDocs::Definitions::Resources.constants.each do |resource_class|
```
zdavis: I can't understand why this change is happening. Is this even an actual class?

api/app/services/dry_types_to_json.rb
```diff
@@ -0,0 +1,89 @@
+class DryTypesToJson
```
zdavis: I'd put this in a module inside of services. For example, `ApiDocumentation::DryTypesToJson`

api/spec/api_docs/config.rb
ConnorChristensen: I changed the name of some of the modules to match the folders in which they were located (eg. there was a `Definition` module located in the `defintions` folder). I was under the impression that this was following the rails convention. Is that correct?

api/spec/api_docs/definitions/resource.rb
ConnorChristensen: That's a good point. I will be more cognizant of what methods are public and private in a class. I was trying to organize the methods so I could track the distinction between request and response functionality, but in doing so, I messed up the private and public statuses.

api/spec/api_docs/config.rb
zdavis: Yes, you have it right—I was confused about the module/class structure. Thanks for checking.

zdavis:
Because this PR is attempting to fixup commits into a commit that's not part of this PR, it can't be accepted. I will manually rebase and fixup the commits into feature/api-docs and open a new PR for that branch.

PR #2483, opened 2019-12-12
[C] Add dry types to comment serializer

ConnorChristensen:
I made a couple more changes than just updating the comment serializer, so I opened up this PR to get some feedback.

I made multiple commits to make it easier to see what I changed and why, and if you like all the changes, I would be happy to squish them down or organize them however you like.


api/spec/api_docs/definitions/resource.rb
```diff
@@ -68,7 +71,12 @@ def write_only_attributes
         const_defined?(:WRITE_ONLY) ? self::WRITE_ONLY : []
       end
 
+      def writable_attributes
+        const_defined?(:WRITEABLE) ? self::WRITEABLE : []
```
ConnorChristensen: Is `WRITABLE` a good name and clear enough description of what these attributes should be?

api/spec/requests/api/v1/comments_spec.rb
```diff
@@ -1,4 +1,3 @@
-require "rails_helper"
```
ConnorChristensen: Rails helper is included in the swagger_helper, so I don't think it's needed here

api/spec/services/api_documentation/dry_types_to_json_spec.rb
```diff
@@ -178,7 +178,7 @@
           items: {
             type: "object",
             properties: {
-              name: { type: 'string' }
+              name: described_class.convert(types['string'])
```
ConnorChristensen: If the `Types::String` JSON output is changed in the future, this test will not require an update. It should have been written like this when I first wrote these tests.

api/spec/api_docs/config.rb
```diff
@@ -24,7 +24,7 @@ def swagger_docs
               title: "Manifold",
               version: "v1"
             },
-            host: "localhost:3020", # TODO: base this off the .env file
+            host: ENV["DOMAIN"],
```
zdavis: Environment should not be accessed directly like this. Instead, if you want the URL of the API, get it from here: `Rails.application.config.manifold.api_url`

api/spec/api_docs/helpers/request.rb
```diff
@@ -164,6 +169,7 @@ def request
       end
 
       def response
+        # TODO: Denote whether resposne is paginated or not
```
zdavis: Typo in comment.

api/spec/api_docs/definitions/resource.rb
zdavis: I don't understand why we need WRITABLE at all. Didn't I see a mechanism for setting attributes to READ_ONLY in a previous commit? Explain why we need READ ONLY and WRITABLE. We should only need one or the other, but not both.

ConnorChristensen: In the case of the comments at least, I had to write out all attributes except for the 3 in the `READ_ONLY` section.
With `WRITABLE`, I can just get the attributes that the validation accepts. Because there are less writable attributes than read only, it makes it easier to see what is there rather than what is not.

I was thinking for some serializer resources, it will make more sense to use `READ_ONLY` and for others `WRITABLE` depending on the ratio between the two.

api/spec/api_docs/config.rb
ConnorChristensen: Thanks!

ConnorChristensen:
I just pushed a commit that scraps the `WRITABLE` interface and will determine whether the attributes are read_only or write_only based on meta values on the dry types. Hopefully I executed this well based on your suggestion in slack:

> It would be even better to just capture this in the serializer
> ...
> Seems like you could capture it there
> ...
> If the type system allows metadata about the type, that would be a good place to put it

Feel free to take a peek if you like but don't go for a full review until I give it another look tomorrow morning with a fresh brain. I don't waste your time with little mistakes if I can help it.

zdavis:
I like that interface a lot better, Connor. It makes a lot of sense to me to include the read-only information in the serializer that way, so please proceed in that direction.

That said, we now have some attributes that are in fact write_only. For example, I think that if you want to delete an attachment called "cover", the request could include a remove_cover attribute (or something like that). These attributes can't live in the serializer, since they are never serialized in the response, so they have to live in our resource definitions. I'm not sure I'd call them WRITE_ONLY though—instead, I think we'd just call them request_attributes or something like that. Do you have this mechanism or something like it for handling attributes that only show up in the request?

api/app/serializers/serializer_registry.rb
```diff
@@ -53,7 +53,7 @@ def typed_attribute(attribute, type, options = {}, &block)
   end
 
   def abilities
-    typed_attribute :abilities, ::Types::Hash do |object, params|
+    typed_attribute :abilities, ::Types::Hash.meta(read_only: true) do |object, params|
```
ConnorChristensen: The attributes are always read_only, correct?

zdavis: You mean the abilities? Yes, abilities are always read only.

PR #2488, opened 2019-12-19
[C] Update Projects API documentation

ConnorChristensen:
Updates the projects documentation generator to use the new internal API

* Add paginated flag on show request
* Move read only attributes to serializer
* Expand on request attributes

### DIFF
--- a/api/spec/api_docs/definitions/resources/project.rb
+++ b/api/spec/api_docs/definitions/resources/project.rb
@@ -3,8 +3,6 @@ module ApiDocs
     module Resources
       class Project
 
-        # TODO: Accepts metadata(Project)
-
         REQUEST_ATTRIBUTES = {
           avatar: Types::Serializer::Upload,
           hero: Types::Serializer::Upload,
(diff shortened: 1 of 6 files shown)


api/spec/api_docs/definitions/resources/project.rb
```diff
@@ -3,43 +3,15 @@ module Definitions
...
-          :citations,
-          :resource_tags,
-          :event_count,
-          :resource_count,
-          :event_types,
-          :metadata_formatted,
-          :metadata_properties,
-          :slug
-        ].freeze
+        # TODO: Accepts metadata(Project)
```
ConnorChristensen: How would I represent this project as a resource that it accepts?

ConnorChristensen:
The resources module felt like it was getting too big, so I created a helper class to handle everything related to documenting metadata. The handler will do validation to make sure the metadata attribute types are defined if the model has metadata values on them. It will also return the metadata structure that should be merged into the request and response attributes.

zdavis:
Please be sure to add a "ready for review" label when it's ready to be accepted into feature/api-docs.

PR #2499, opened 2019-12-29
[B] Restore height to annotation popup wrapper

### LINKED ISSUE #2469: Pop-Up Menu Hovers Outside Viewscreen Area
When highlighting text near the top of text sections the pop-up modal inconsistently hovers outside clickable screen area:

menu

This comes from this text: `https://edge.manifoldapp.org/read/76161e0e-5a04-441d-8c44-29c52e643319/section/6d57c48b-8ffb-4c97-a9b3-355e80fe8006`.

zdavis:
When the panel is positioned absolutely, the .annotation-popup ancestor
element does not have height. The popup positioner needs to be able to
accurately determine the height off the .annotation-popup container so
that it can determine whether to position it above or below the
selection. One of the jobs of the popup positioner is to prevent the
popup from colliding with the edges of the window, and it cannot do this
without an accurate height measurement.

Fixes #2469

### DIFF
--- a/client/src/theme/styles/components/reader/annotations/_popup.scss
+++ b/client/src/theme/styles/components/reader/annotations/_popup.scss
@@ -27,7 +27,7 @@
 
   &__panel {
     @include boxShadow(0, 12px, 22px, -3px, $neutralBlack, 0.39);
-    position: absolute;
+    position: relative;
     left: 50%;
     background-color: $panelBgColor;
     border-radius: $panelRoundedRadius;


zdavis:
Would appreciate a quick review from @dananjohnson, as I believe this bug was introduced by your refactoring of the annotation popup panels during reading group development. I've tested this change, and it seems to work, but if you have any concerns, I'm open to other solutions.

PR #2513, opened 2019-12-30
[WIP] Document ActionCallout API endpoints

ConnorChristensen:
On PATCH for an `ActionCallout`, I tried sending in data like this:

```
{
  "data": {
    "attributes": {
      "attachment": {
        "filename": "profile_pic.jpg",
        "data": "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNkYPj/HwADBwIAMCbHYQAAAABJRU5ErkJggg==",
        "contentType": "image/jpg"
      }
    }
  }
}
```

The `data` value is a base-64 1x1 blue pixel. This works for setting an avatar image for a user and I was assuming it would work similarly in an `ActionCallout,` but all the `AttachmentStyles` in the `ActionCallout` response are still null.

### DIFF
--- a/api/app/serializers/v1/action_callout_serializer.rb
+++ b/api/app/serializers/v1/action_callout_serializer.rb
@@ -3,13 +3,14 @@ module V1
 
     include ::V1::Concerns::ManifoldSerializer
 
-    typed_attribute :title, NilClass
-    typed_attribute :kind, NilClass
-    typed_attribute :location, NilClass
-    typed_attribute :position, NilClass
-    typed_attribute :url, NilClass
-    typed_attribute :button, NilClass
-    typed_attribute :attachment_styles, Types::Hash
+    typed_attribute :title, Types::String
+    typed_attribute :kind, Types::String.enum("link", "read", "toc", "download")
+    typed_attribute :location, Types::String.enum("left", "right")
+    typed_attribute :position, Types::Integer
+    typed_attribute :url, Types::Serializer::URL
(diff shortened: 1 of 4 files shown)


zdavis:
Failing due to invalid base64 encoded data. Should be:

```
                "attributes": {
                  "attachment": {
                    "filename": "profile_pic.png",
                    "data": "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z/C/HgAGgwJ/lK3Q6wAAAABJRU5ErkJggg==",
                    "contentType": "image/png"
                  }
                }
```

ConnorChristensen:
Thank you for the feedback! The tests were all working fine, this was just a problem I was running into while manually testing the route to make sure I understood it.

As such, no changes to the logic needed to be made, but I updated the example upload fields so a user can send a default example on the swagger docs and have it be valid.

zdavis:
I understand what you're saying, but in the future I'd still like to see a test to help me help you. The spec doesn't have to be an API doc spec. When I started working on this, I just wrote a normal request spec that posted the params that you included, which made it very easy to reproduce the issue and see why the attachment wasn't getting set. So, even if your tests are passing, if you're claiming in your review that there's a problem with the route, I'd like to see a spec that demonstrates the problem in addition to your comment. That way, I don't have to spend time reproducing the issue. Thanks.

PR #2514, opened 2019-12-30
[WIP] Document Annotation API endpoints

ConnorChristensen:
Two issues for this group of routes:

1. It appears the show method on the controller that manages the `/annotations/{id}` route requests does not exist
2. I get a 500 response on an index request to `/me/relationships/annotations`. I am not sure if there is something wrong with that one or if I am just accessing the route incorrectly.

### DIFF
--- a/api/app/serializers/serializer_registry.rb
+++ b/api/app/serializers/serializer_registry.rb
@@ -69,7 +69,7 @@ class SerializerRegistry
   end
 
   def current_user_is_creator?
-    typed_attribute :current_user_is_creator, ::Types::Bool do |object, params|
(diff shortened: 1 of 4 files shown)


zdavis:
For #1, open a new issue asking that the /annotations/:id show route be removed, since it's not supported by the controller. Exclude it from documentation.

For #2, tell me what you see in the rails log or in the exception, or write a test to reproduce this. Keep in mind that there is already a spec for /me/relationships/annotations in specs/requests/me/relationships/annotations, and that test is passing, which suggests that you're doing something wrong in your swagger spec.

PR #2523, opened 2020-01-03
[WIP] Document Favorite API endpoints

ConnorChristensen:
I had to make a few adjustments to the resource module in order to be able to set relationships as a required field in the request. 

I also figured it wise to not attempt to programmatically document the response behavior for creating a favorite.  In the favorites controller, there is a comment that says the response behavior is rather unexpected and should be refactored at some point, so I documented the response behavior with a simple description on the swagger route.

### DIFF
--- a/api/spec/api_docs/definitions/resource.rb
+++ b/api/spec/api_docs/definitions/resource.rb
@@ -2,35 +2,39 @@ module ApiDocs
   module Definitions
     module Resource
       def update_request
-        definition = make_request(__callee__, update_attributes || request_attributes)
+        attributes = assign_required_attributes(
+          (update_attributes || request_attributes),
+          required_update_attributes
+        )
 
-        definition = ApiDocumentation::DryTypesParser.convert(definition)
-        transform_keys(definition)
+        relationships = assign_required_attributes(update_relationships, required_update_relationships)
+
+        definition = make_request(attributes, relationships)
+        parse_dry_types(__callee__, definition)
       end
 
       def create_request
-        definition = make_request(__callee__, create_attributes || request_attributes)
+        attributes = assign_required_attributes(
+          (create_attributes || request_attributes),
+          required_create_attributes
+        )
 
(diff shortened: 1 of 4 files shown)


api/spec/requests/api/v1/favorites_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    path "/me/relationships/favorites" do
+      include_examples "an API index request",
+                       model: Favorite,
+                       auth_type: :admin
+
+      include_examples "an API create request",
+                       model: Favorite,
+                       auth_type: :admin,
+                       description: "If the favorite is valid, it will return the current user. "\
```
ConnorChristensen: Here is the spot where I documented the unique response behavior of the route. Hopefully the description is easily understood.

api/spec/api_docs/definitions/resource.rb
```diff
@@ -80,14 +108,20 @@ def write_only_attributes
         self::REQUEST_ATTRIBUTES
       end
 
-      def make_request(callee, attributes)
-        attributes = ::Types::Hash.schema(attributes)
-        attributes = attributes.meta(required: required_create_attributes) if required_create_attributes.present? && callee == :create_request
-        attributes = attributes.meta(required: required_update_attributes) if required_update_attributes.present? && callee == :update_request
+      def build_value(contents, required_values)
```
ConnorChristensen: I could not think of a great name for this method, if you have a suggestion, that would be appreciated.

zdavis: I'm not sure what this method is doing. Can you explain what it's doing? Maybe the word you're looking for here is `hydrate`, which refers to the process of filling an object with data?

ConnorChristensen: I cleaned up the code a bit, so even though there are changes, it should be more straightforward to understand.

PR #2524, opened 2020-01-03
[WIP] Upgrade authorization rswag tests

ConnorChristensen:
I documented why I made each change in the commit descriptions, so do look those over for more information on the changes. Feel free to look over all the changes in one go, or take it one commit at a time, but I suggest looking by commits to get a better sense of how the changes relate to each other.

The summary of these changes are that I made some clarification and enhancement changes to how response codes are tested and documented in rswag tests.

* Clarity changes involved the interface for passing information into rswag routes for testing response codes
* Enhancement changes involve expanding on the response codes tested/documented for each CRUD operation, as well as verifying which users can access which resources in the rswag tests.

Apart from adding a couple new 403 and 404 response tests, these changes cause some of the routes within the tag grouping on the swagger.json file to be rearranged, I think due to run time changes on the tests (the tests that have loops around them take longer to execute), but this has no effect on how the routes are displayed using the swagger interface.

### DIFF
--- a/api/spec/api_docs/examples/create.rb
+++ b/api/spec/api_docs/examples/create.rb
@@ -15,11 +15,11 @@ shared_examples_for "an API create request" do |options|
     description api_spec_helper.response_description if api_spec_helper.response_description?
     produces api_spec_helper.content_type
     consumes api_spec_helper.content_type
-    security [apiKey: []] if api_spec_helper.with_auth
+    security [apiKey: []] if api_spec_helper.requires_auth?
     tags api_spec_helper.tags
 
     response "201", api_spec_helper.success_description, focus: api_spec_helper.focus do
-      let(:Authorization) { get_user_token(api_spec_helper.auth_type) }
+      let(:Authorization) { get_user_token(api_spec_helper.authorized_user) } if api_spec_helper.requires_auth?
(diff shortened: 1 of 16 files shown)


zdavis:
This looks pretty good, although I have some concerns. When you give a test a set of say 5 user roles, is it going to run the request for each role? I'm not sure it's desirable to run each request against each role type, since that will significantly increase the time needed to run the suite, and in many cases the response is the same. Am I reading this correctly in thinking that you run the request for each type, but only one request figures into the generated documentation? If that's true, wouldn't it be better to test this stuff in the request specs?

All in all, I'm kind of curious what prompted this refactor, and while the work looks pretty good, I'm not sure I see why it's necessary. We should just chat about this when we have a chance.

ConnorChristensen:
Those are some good points and I would like to chat about this sometime. I think I might have too ambitious with this refactor, so I'm scaling it back and just making the refactor change on the authenticated_request shared context and changing `auth_type` to `authorized_users` to make the interface on the rswag tests a little more clear. 

I'm saving my other work on another branch and we can talk about it at a later time. Since the other work will all be internal changes, my work moving forward will not be affected by including or excluding those changes.

PR #2528, opened 2020-01-06
[WIP] Document CollectionProjects API endpoints

ConnorChristensen:
I'm running into a problem with creating a CollectionProject on a ProjectCollection.

To reproduce the error,  run `rspec -fd spec/requests/api/v1/collection_project_spec.rb` from the `api` folder.

### DIFF
--- a/api/app/serializers/v1/collection_project_serializer.rb
+++ b/api/app/serializers/v1/collection_project_serializer.rb
@@ -3,7 +3,7 @@ module V1
 
     include ::V1::Concerns::ManifoldSerializer
 
-    typed_attribute :position, NilClass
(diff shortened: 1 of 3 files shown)


zdavis:
Removing question label.

zdavis:
Please remove the use of faker from this PR, unless you can explain why you made that change. In general, I think factories should be deterministic, not returning random data.

ConnorChristensen:
I made the change to the factory bot because the title for a project_collection must be unique, and while I was playing around with factory bot, it was difficult to create more than one collection project. 

I should not have made that a permanent change though, as I reverted the factory, and my rswag tests still pass. I updated my PR and left the factory in it's original form.

PR #2533, opened 2020-01-08
[WIP] Document Events API endpoints

ConnorChristensen:
As far as I know, there are no routes that create an event. I'm assuming that it goes on behind the scenes when a user does something to their project. So I marked each field as read_only.


zdavis:
That's correct.

PR #2538, opened 2020-01-09
[WIP] Document ReadingGroup API endpoints

ConnorChristensen:


api/app/serializers/v1/reading_group_serializer.rb
```diff
@@ -5,26 +5,25 @@ class ReadingGroupSerializer < ManifoldSerializer
...
 
-    typed_attribute :name, NilClass
-    typed_attribute :privacy, NilClass
-    typed_attribute :invitation_code, NilClass
-    typed_attribute :notify_on_join, NilClass
-    typed_attribute :memberships_count, NilClass
-    typed_attribute :annotations_count, NilClass
-    typed_attribute :highlights_count, NilClass
-    typed_attribute :created_at, NilClass
-    typed_attribute :current_user_role, NilClass
```
ConnorChristensen: I removed this line when I added types to the serializer. It is my impression that because the attribute is defined twice in the file, rails will only use the the last occurrence of the definition in the file (so this line is redundant). Is that correct?

api/app/serializers/v1/reading_group_serializer.rb
```diff
@@ -5,26 +5,25 @@ class ReadingGroupSerializer < ManifoldSerializer
...
-    typed_attribute :creator_id, NilClass
-    typed_attribute :texts, NilClass
-    typed_attribute :all_annotations_count, NilClass do |object, _params|
+    typed_attribute :name, Types::String
+    typed_attribute :privacy, Types::String.enum("public", "private", "anonymous")
+    typed_attribute :invitation_code, Types::String.meta(example: "E80RSYY6", unique: true)
```
ConnorChristensen: I added a new "unique" meta tag to be used from here on out for attributes that have a requirement for uniqueness  on the model.

api/app/services/api_documentation/dry_types_parser.rb
```diff
@@ -51,11 +55,21 @@ def merge_enum_values(hash, type)
...
-        filtered_meta = type.meta.select do |k, _|
+      def allowed_meta_values(type, allowed_swagger_meta_types)
+        type.meta.select do |k, _|
           allowed_swagger_meta_types.include? k
         end
-        hash.merge(filtered_meta)
+      end
+
+      def add_description_for_uniqueness(converted_hash)
+        converted_hash[:description] = (converted_hash[:description].to_s + " Must be unique.").strip
```
ConnorChristensen: I went with this implementation of string concatenation on the attribute description since it is the most concise and I am only combining two strings, but  if I need to do any more string building on the description in the future, I will be sure to change this to the array `.join(" ")` approach.

api/app/serializers/v1/reading_group_serializer.rb
zdavis: That is correct.

api/app/serializers/v1/reading_group_serializer.rb
zdavis: Sounds good.

PR #2539, opened 2020-01-10
[WIP] Document Ingestion API endpoints

ConnorChristensen:
I'm having a hard time getting an exact definition or example of the structure of the `source` request attribute for an ingestion. Any help would be appreciated.

The structure I have commented out right now was gathered from the validation for an ingestion:

```
  def ingestion_params
    params.require(:data)

    attributes = [
      :external_source_url,
      :ingestion_type,
      {
        source: [
          :id,
          :storage, { metadata: [:filename, :size, :mime_type] }
        ]
      }
    ]
    relationships = [:text]

    param_config = structure_params(
      attributes: attributes,
      relationships: relationships
    )

    params.permit(param_config)
  end
```

### DIFF
--- a/api/app/serializers/v1/ingestion_serializer.rb
+++ b/api/app/serializers/v1/ingestion_serializer.rb
@@ -3,17 +3,16 @@ module V1
 
     include ::V1::Concerns::ManifoldSerializer
 
-    typed_attribute :state, NilClass
-    typed_attribute :external_source_url, NilClass
-    typed_attribute :log, NilClass
-    typed_attribute :source_file_name, NilClass
-    typed_attribute :ingestion_type, NilClass
-    typed_attribute :available_events, NilClass
-    typed_attribute :strategy, NilClass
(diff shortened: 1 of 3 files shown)


ConnorChristensen:
From Zach on slack:
> Connor, you should leave a note for me to come back to the ingestion source attribute
> It’s a little different from other uploads because first there is a post to http://manifold.lvh/api/files which is our TUS upload endpoint. From that, the client gets a file ID, which is included in the post to ingestions. The source on that post looks like this:
>
> ```
> "source":{ 
>  "id":"http://manifold.lvh/api/files/338756a5557ed13485a42f7b931c3cab",
>  "storage":"cache",
>  "metadata":{ 
>    "filename":"anton-chekhov_the-duel.epub",
>    "size":683346,
>    "mimeType":"application/epub+zip"
>  }
> }
> ```

I made adjustments accordingly, and left a note to circle back to this at a later time.

PR #2548, opened 2020-01-14
[WIP] Document Me API endpoints

ConnorChristensen:
While trying to document the `Me` resources, I ran into an error on assigning a dry type to a `typed_attribute`.

I made a note in the serializer_registry where the error is being generated. To replicate the error in a test, run `rspec -fd spec/requests/api/v1/me_spec.rb` from the api folder.

### DIFF
--- a/api/app/serializers/v1/current_user_serializer.rb
+++ b/api/app/serializers/v1/current_user_serializer.rb
@@ -6,7 +6,9 @@ module V1
 
     set_type :user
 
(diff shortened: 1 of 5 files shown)


zdavis:
See my two commits for fixes here. When you fixup this branch, keep 4d3d0ab as is, but roll up 333d913 into your work. I left an extended commit message on 333d913 that explains why this route is different.

PR #2555, opened 2020-01-14
[WIP] Improve Form.Upload a11y

### LINKED ISSUE #2439: Project Thumbnail "Upload a File" Button Lacks Focus Indicator
As mentioned in the title, and in a previous issue, this specific instance of the "Upload a File" Button has no focus indicator. It can be found on the  General Attributes Page in the Admin Interface for an editable project.

image

This is likely due to its unique style as a sort of merged button with the button to the left of it.

dananjohnson:
Makes the following improvements to Form.Upload components:
* removes upload button (redundant with upload `input`)
* provides visual indication that `input` is focused. This required
updating `react-dropzone` from `4.x` to `10.x`, as newer versions allow
more granular control of dropzone markup. Unfortunately this
library lacks some accessibility features out of the box (namely
incorrect `tabindex` preventing focus on `input`), so
improvements were made here as well.
* properly associates labels with corresponding inputs. A similar
improvement is made to the ColorPicker component, which is used
alongside the Form.Upload components in selecting a project thumbnail.

Resolves #2439


dananjohnson:
The `react-dropzone` update seems to be causing a number of broken tests. I suspect something needs to be changed in `client/src/global/components/form/__tests__/Upload-test.js` to resolve this, but I'm not certain. @zdavis could you take a look when you have a moment?

zdavis:
It looks like there are a few issues, @dananjohnson. First, there are a number of snapshot tests failing, probably because you changed the markup in the component. You can fix those with `jest -u`. It looks like there are also some test that make assertions around how the component behaves. For example:

```
it("should contain a remove link", () => {
  expect($wrapper.find('[data-id="remove"]')).toHaveLength(1);
});
```
If you've changed the markup in the component, these tests may need to be changed or updated. Have you tried adjusting the tests? Or do you think there's a more widespread problem here?

dananjohnson:
Thanks for looking at this, @zdavis. With the newer version of `react-dropzone`, the markup of the dropzone component just gets rendered like this by the test (whereas before the inner markup was actually rendered):
```jsx
<react-dropzone
  onDrop={[Function]}
>
  <Component />
</react-dropzone>
```

Tests like `expect($wrapper.find('[data-id="remove"]')).toHaveLength(1);` that look for certain DOM nodes in the dropzone are failing as a result.

My hunch is that this relates to the new version of `react-dropzone` using hooks to render::
```jsx
<Dropzone onDrop={this.handleFileDrop}>
  {({ getRootProps, getInputProps }) => (
    ...
  )}
</Dropzone>
```
But I'm not sure if this is correct, nor how the tests should be updated if this is the case.

dananjohnson:
I'm actually seeing this pattern in a lot of other snapshots (e.g. [here](https://github.com/ManifoldScholar/manifold/blob/master/client/src/global/components/form/__tests__/__snapshots__/FieldGroup-test.js.snap)), beginning with [this commit](https://github.com/ManifoldScholar/manifold/commit/d090e04db9da19d24094810c9ca939b1db18f577) it seems. So maybe tests like the one below can safely be removed? 
```javascript
it("should contain a remove link", () => {
  expect($wrapper.find('[data-id="remove"]')).toHaveLength(1);
});
```

zdavis:
Rebased on master

PR #2564, opened 2020-01-16
[STALE] Update elasticsearch

zdavis:
A few changes that were made to deal with breaking changes in elasticsearch:

* Transport type now requires security.
* If you want to set a low watermark, you must also set a high
  watermark and a flood_stage value. Reverting to default values.
* Elasticsearch no longer provides a basic TAR file for all operating
  systems. The only difference between the elasticsearch packages
  is that they contain an installer that is specific to each operating
  system. Since we are not doing a system wide install, the macOS
  tar should work fine for every operating system.
* The query and response data structures in elasticsearch have changed.
  Some of the manifold code that makes the queries and parses the results
  have been updated accordingly.


zdavis:
Resolved in master. PR no longer useful.

PR #2569, opened 2020-01-17
[WIP] Document Stylesheet API endpoints

ConnorChristensen:


api/app/serializers/v1/stylesheet_serializer.rb
```diff
@@ -3,15 +3,23 @@ class StylesheetSerializer < ManifoldSerializer
 
     include ::V1::Concerns::ManifoldSerializer
 
-    typed_attribute :name, NilClass
-    typed_attribute :source_identifier, NilClass
```
ConnorChristensen: What is `source_identifier` used for? Do people upload a CSS file to manifold?

api/app/serializers/v1/stylesheet_serializer.rb
```diff
@@ -3,15 +3,23 @@ class StylesheetSerializer < ManifoldSerializer
...
-    typed_attribute :styles, NilClass
-    typed_attribute :ingested, NilClass
-    typed_attribute :position, NilClass
-    typed_attribute :created_at, NilClass
+    typed_attribute :name, Types::String
+    typed_attribute :source_identifier, Types::String.optional.meta(read_only: true)
+    typed_attribute :styles, Types::String.optional.meta(
```
zdavis: Please change the description to:

"Validated, scoped, and transformed styles that will be applied in the reader."

ConnorChristensen: Done

api/spec/api_docs/definitions/resources/stylesheet.rb
```diff
@@ -0,0 +1,16 @@
+module ApiDocs
+  module Definitions
+    module Resources
+      class Stylesheet
+
+        REQUIRED_CREATE_ATTRIBUTES = [:name].freeze
+
+        READ_ONLY_RELATIONSHIPS = [:text].freeze
```
ConnorChristensen: What do you think of this as an interface for noting which relationships are read only (and consequently which ones are writable)? Eventually, I think we should include writable relationships in the request examples on the documentation, which means that the `ApiDocs::Definitions::Resource` module and the `ApiDocs::Helpers::Request` module will need to know which relationships are not writable.

This makes an assumption that every route that has included resources attached to it in it's response, will be able to write to those relationships that are allowed through the validator. Is this a correct assumption?

For example, a stylesheet has included `text` and `text_sections` relationships, so if I pass in a `READ_ONLY_RELATIONSHIPS` value of `text`, then the `text_sections` relationships will end up as part of an example request on the documentation.

zdavis: I think this is OK for now. Some inconsistency will creep in because we're effectively copying logic from the validation concern into the API definition, so I expect this will change. With this type system we can introspect our serializers and understand what they're returning. We don't have a similar system for the request side of things, as the question of what is writable is answered in a rather ad hoc way in that validation concern. Adding this interface to the API definitions is OK for now, and as far as I'm concerned whether we add it as READ_ONLY_RELATIONSHIPS or WRITABLE_RELATIONSHIPS is a six of one half dozen of another question.

PR #2572, opened 2020-01-17
[WIP] Document ResourceImport API endpoints

ConnorChristensen:
I'm getting this error on the GET for the `/projects/{project_id}/relationships/resource_imports/{id}` route:

```
Rswag::Specs::UnexpectedResponse:
      Expected response code '500' to match '200'
      Response body: {
        "status": 500,
        "error": "Internal Server Error",
        "exception":"#RuntimeError: The typed attribute :data on SerializerRegistry is not a hash. 
                     Consider changing the type to AttachmentUploader::UploadedFile",
```

Here is a snippet from a sample response on the the GET for the `/projects/{project_id}/relationships/resource_imports/{id}` route:
```
"attributes": {
  "source": "attached_data",
  "data": {
    "original": {
      "id": "resourceimport/151747a1-ec84-492d-bf1f-7cb407881f5a/data/original-248172aeb6805275060ad0f6283afee1.csv",
      "storage": "store",
      "metadata": {
        "size": 17816,
        "width": null,
        "height": null,
        "sha256": "d77ff3c1d5cf4958b371fa7f38f256b651b7846c915b5dea7651f3adba91c236",
        "filename": "resources.csv",
        "mime_type": "application/octet-stream"
      }
    }
  },
```

Running `rspec -fd spec/requests/api/v1/resource_imports_spec.rb` from the api folder will reproduce my issue.

### DIFF
--- a/api/app/serializers/v1/resource_import_serializer.rb
+++ b/api/app/serializers/v1/resource_import_serializer.rb
@@ -3,23 +3,37 @@ module V1
 
     include ::V1::Concerns::ManifoldSerializer
 
-    typed_attribute :source, NilClass
-    typed_attribute :data, NilClass
-    typed_attribute :column_map, NilClass
-    typed_attribute :column_automap, NilClass
-    typed_attribute :header_row, NilClass
-    typed_attribute :headers, NilClass
-    typed_attribute :available_columns, NilClass
-    typed_attribute :data_filename, NilClass
-    typed_attribute :storage_type, NilClass
-    typed_attribute :storage_identifier, NilClass
-    typed_attribute :url, NilClass
-    typed_attribute :parse_error, NilClass
-    typed_attribute :import_results, NilClass do |object|
+    typed_attribute :source, Types::String.enum("google_sheet", "attached_data")
+    typed_attribute :data, Types::Hash.schema(
+      original: Types::Hash.schema(
+        id: Types::String.meta(
+          example: "resourceimport/151747a1-ec84-492d-bf1f-7cb407881f5a/data/original-248172aeb6805275060ad0f6283afee1.csv"
+        ),
+        storage: Types::String.meta(example: "store"),
+        metadata: Types::Hash.schema(
+          size: Types::Integer,
+          width: Types::Integer.optional,
+          height: Types::Integer.optional,
+          sha256: Types::String,
+          filename: Types::String,
+          mime_type: Types::String.meta(example: "application/octet-stream")
+        )
+      )
+    ).meta(description: "Required if source is an attachment", read_only: true)
+    typed_attribute :column_map, Types::Hash
+    typed_attribute :column_automap, Types::Hash.meta(read_only: true)
+    typed_attribute :header_row, Types::Integer
+    typed_attribute :headers, Types::Hash.meta(read_only: true)
+    typed_attribute :available_columns, Types::Array.of(Types::String).meta(read_only: true)
+    typed_attribute :storage_type, Types::String
+    typed_attribute :storage_identifier, Types::String
+    typed_attribute :url, Types::Serializer::URL.optional.meta(description: "Required if source is a google sheet")
+    typed_attribute :parse_error, Types::Bool.meta(read_only: true)
+    typed_attribute :import_results, Types::Array.of(Types::String).meta(read_only: true) do |object|
       object.import_results.map { |r| camelize_hash(r) }
     end
-    typed_attribute :data_filename, NilClass, &:data_file_name
-    typed_attribute :state, NilClass do |object|
+    typed_attribute :data_filename, Types::String.meta(read_only: true), &:data_file_name
+    typed_attribute :state, Types::String.meta(example: "pending") do |object|
       object.state_machine.current_state
     end
 
--- a/api/app/serializers/v1/resource_import_serializer.rb
+++ b/api/app/serializers/v1/resource_import_serializer.rb
@@ -4,22 +4,6 @@ module V1
     include ::V1::Concerns::ManifoldSerializer
 
     typed_attribute :source, Types::String.enum("google_sheet", "attached_data")
-    typed_attribute :data, Types::Hash.schema(
-      original: Types::Hash.schema(
-        id: Types::String.meta(
-          example: "resourceimport/151747a1-ec84-492d-bf1f-7cb407881f5a/data/original-248172aeb6805275060ad0f6283afee1.csv"
-        ),
-        storage: Types::String.meta(example: "store"),
-        metadata: Types::Hash.schema(
-          size: Types::Integer,
-          width: Types::Integer.optional,
-          height: Types::Integer.optional,
-          sha256: Types::String,
-          filename: Types::String,
-          mime_type: Types::String.meta(example: "application/octet-stream")
-        )
-      )
-    ).meta(description: "Required if source is an attachment", read_only: true)
     typed_attribute :column_map, Types::Hash
     typed_attribute :column_automap, Types::Hash.meta(read_only: true)
     typed_attribute :header_row, Types::Integer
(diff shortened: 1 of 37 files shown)


ConnorChristensen:
Update on this after looking into it more:

### Runtime Error
I got a runtime error that said `#RuntimeError: The typed attribute :data on SerializerRegistry is not a hash. Consider changing the type to AttachmentUploader::UploadedFile`. This error did not make sense to me, as every type definition on a serializer should be a DryType and it was suggesting I give it a class or module. 

I followed the instructions regardless, as there are still many parts of this application that I don't understand. Based on this suggestion, the typed attribute in the resource_import serializer would look like this: `typed_attribute :data, AttachmentUploader::UploadedFile`. If I add an `UploadedFile` type on the serializer, I get a failure in the process that converts DryTypes to swagger JSON, as that process expects a DryType as input.

I took a step back and found where the runtime error `SerializerRegistry is not a hash` was being raised. I found it [here](https://github.com/ManifoldScholar/manifold/blob/master/api/app/serializers/serializer_registry.rb#L91) in the serializer_registry. I still don't fully understand how this proc works, but I do know that if in every instance that the `object.public_send(attribute)` line is run, it returns an empty hash, *except* for on this resource_import test, where `object.public_send(attribute)` returns:

```
#<AttachmentUploader::UploadedFile:0x00007fad30484998 @data={"id"=>"resourceimport/data/6a492c415e43ee7d19dec73461dbf3d8.csv", "storage"=>"cache", "metadata"=>{"size"=>17816, "width"=>nil, "height"=>nil, "sha256"=>"d77ff3c1d5cf4958b371fa7f38f256b651b7846c915b5dea7651f3adba91c236", "filename"=>"resources.csv", "mime_type"=>"application/octet-stream"}}>
```

So, this `AttachmentUploader` does not respond to the `keys` method, as it is not a hash, and it rasises an error.

I can keep working on this to find a solution, but I think I need some help understanding how this code actually works, so I don't break something in my fixes. 

For example, if I make a change on the `build_camelize_proc` method like so:

```
  def build_camelize_proc(attribute, block = nil)
    proc { |object, params|
      value = if block
                block.arity.abs == 1 ? block.call(object) : block.call(object, params)
              else
                {} # used to be: object.public_send(attribute)
              end
      unless value.respond_to? :keys
        raise "The typed attribute :#{attribute} on #{self.class} is not a hash.
              Consider changing the type to #{value.class}"
      end
      klass.camelize_hash(value)
    }
  end
```

Then the test passes. But I can't explain why `object.public_send(attrbiute)` was there in the first place.

### No implicit conversion of nil into String
Even after moving past this first error, I get this error on the `update` test:

```
 2) Resource Imports /projects/{project_id}/relationships/resource_imports/{id} patch Returns the updated ResourceImport returns a 200 response
     Failure/Error: example.run

     Rswag::Specs::UnexpectedResponse:
       Expected response code '500' to match '200'
       Response body: {"status":500,"error":"Internal Server Error","exception":"#\u003cTypeError: no implicit conversion of nil into String\u003e","traces":{"Application Trace":[{"id":4,"trace":"app/lib/updaters.rb:95:in `attachment_from_params!'"},{"id":5,"trace":"app/lib/updaters.rb:79:in `block in attachmentize_attributes!'"},{"id":6,"trace":"app/lib/updaters.rb:78:in `each'"},{"id":7,"trace":"app/lib/updaters.rb:78:in `attachmentize_attributes!'"},{"id":8,"trace":"app/lib/updaters.rb:69:in `block in assign_attributes!'"}
```

Which is happening on this line [here](https://github.com/ManifoldScholar/manifold/blob/master/api/app/lib/updaters.rb#L94). Again, I'm not sure what the expected behavior is or what this updater is trying to do, as out of all the tests, the `patch` on a resource_import is the only one that even makes it past [the guard clause](https://github.com/ManifoldScholar/manifold/blob/master/api/app/lib/updaters.rb#L92) two lines above where the failure occurs.

I am trying to be respectful of your time, but also trying not to spend too much time trying to hunt for bugs/understand a code base that is over my head. Do feel free to send me back to work on it if you are too busy or are ok with me taking more time to understand the code base more.

zdavis:
I'll take a look at this as soon as I can and try to figure out what's up! Thanks for investigating further, but I think you can move on to other tasks until I investigate.

ConnorChristensen:
Sounds good. Thank you! I'm curious about what's going on, so I'll be sure to look over whatever you come up with.

api/app/serializers/v1/resource_import_serializer.rb
```diff
@@ -4,22 +4,6 @@ class ResourceImportSerializer < ManifoldSerializer
     include ::V1::Concerns::ManifoldSerializer
 
     typed_attribute :source, Types::String.enum("google_sheet", "attached_data")
-    typed_attribute :data, Types::Hash.schema(
```
zdavis: This is a weird case in that it's handling uploads differently from other parts of the application. Take an "avatar" upload field, for example. Normally the request would include something like this in the avatar attribute:
```
    {
      content_type: text/csv,
      data: "data:text/csv;base64,foobar",
      filename: resources.csv
    }
```
And the response would include `avatar_styles` with info about the avatars.

In this case, we're using attributes.data for the request _and_ the response, which makes things hard because the factories are mostly setup to generate a structure that we can look at for the response, not the request. 

There are a few things I've done to address this:

1. Refactored the 'param_helpers' methods. They were messy and hard to understand with poor method names.
2. Made the methods context sensitive by including a `type` keyword argument that can be :request or :response
3. If we're generating a json request structure, the helper will loop through the attributes and see if any of the values are file-like. If they are, it will replace the object with a suitable request datastructure.
4. I've also removed the `data` attribute from the resource import serializer response. It is never used in the client application, and trying to support the same property that looks one way in the request and another way in the response is more work than I want to tackle right now.

api/spec/api_docs/examples/create.rb
```diff
@@ -5,7 +5,7 @@
   api_spec_helper = ApiDocs::Helpers::Request.new(options, :create)
 
   # TODO: Convert serializer information into a request
-  let(:body) { json_structure_for(api_spec_helper.factory) }
+  let(:body) { json_structure_from_factory(api_spec_helper.factory, type: :request) }
```
zdavis: Tell the helper that it's a request

api/spec/api_docs/definitions/resources/resource_import.rb
```diff
@@ -5,6 +5,11 @@ class ResourceImport
 
         REQUIRED_CREATE_ATTRIBUTES = [:header_row].freeze
 
+        REQUEST_ATTRIBUTES = {
+            data: Types::Serializer::Upload
```
zdavis: Data is an upload

api/spec/support/requests/param_helpers.rb
```diff
@@ -2,38 +2,50 @@
...
 
-  def relationship_payload(name, models)
-    Hash[name, models]
+  def build_json_payload(**options)
+    build_json_structure(options).to_json
   end
 
-  def json_payload(attributes: {}, relationships: {}, meta: {})
-    json_structure(attributes: attributes, relationships: relationships, meta: meta).to_json
+  def adjust_attributes(attributes, type)
```
zdavis: Now we make some adjustments to the attributes if we're dealing with a request.

api/spec/support/requests/param_helpers.rb
```diff
@@ -2,38 +2,50 @@
...
+    return adjust_attributes_for_request(attributes) if type == :request
+    adjust_attributes_for_response(attributes)
+  end
+
+  def adjust_attributes_for_request(attributes)
+    attributes.map { |k,v|
+      next [k, v] unless v.respond_to? :path
```
zdavis: Does the attribute value quack like a duck?

PR #2579, opened 2020-01-22
[WIP] Document ReadingGroupMembership API endpoints

ConnorChristensen:


api/spec/api_docs/examples/destroy.rb
```diff
@@ -4,6 +4,11 @@
 
   api_spec_helper = ApiDocs::Helpers::Request.new(options, :destroy)
 
+  let(:resource_instance) do
+    defined?(resource) ? resource : FactoryBot.create(api_spec_helper.factory)
+  end
```
ConnorChristensen: I brought this code from the `update` request shared context so the delete test will create a resource to attempt to delete if it is not already provided.

PR #2580, opened 2020-01-22
[WIP] Document Settings API endpoints

ConnorChristensen:


api/spec/requests/api/v1/settings_spec.rb
```diff
@@ -0,0 +1,22 @@
+require "swagger_helper"
+
+RSpec.describe "Settings", type: :request do
+
+  let(:resource) { Settings.first }
```
ConnorChristensen: Is this a reasonable assumption that there will be a setting that already exists? I think the app in configured so there can only be one setting, so when I tried to run a `FactoryBot.create` on the setting, it came back with an error that one already existed and there could only be one at a time.

zdavis: Yes! Settings is a singleton object, since a Manifold installation can only have one set of settings. To get the settings, we use `Settings.instance` and you should use that instead of Settings.first. Good catch though, Connor.

ConnorChristensen: Thank you! I'll make that change now.

PR #2581, opened 2020-01-22
[WIP] Document TextSection API endpoints

ConnorChristensen:


api/app/serializers/v1/text_section_serializer.rb
```diff
@@ -3,19 +3,24 @@ class TextSectionSerializer < ManifoldSerializer
...
     typed_belongs_to :text
     typed_has_many :stylesheets
 
     when_full do
-      typed_attribute :body_json, Types::Hash
-      typed_attribute :citations, Types::Hash
+      typed_attribute :body_json, Types::Hash.schema(
+        tag: Types::String.meta(example: "div"),
+        children: Types::Array.of(Types::Hash).meta(
+          description: "A tree of tag and child arrays, each containing a node with a tag, nodeType and attributes"
```
ConnorChristensen: Is this a good description for what the `body_json` does? Since the depth can be variable, it makes more sense to just describe this with a description rather than a DryTypes structure.

api/app/serializers/v1/text_section_serializer.rb
```diff
@@ -3,19 +3,24 @@ class TextSectionSerializer < ManifoldSerializer
...
-    typed_attribute :text_slug, NilClass
-    typed_attribute :text_title, NilClass
-    typed_attribute :name, NilClass
-    typed_attribute :social_image, NilClass
-    typed_attribute :source_identifier, NilClass
-    typed_attribute :kind, NilClass
+    typed_attribute :text_slug, Types::String
+    typed_attribute :text_title, Types::String
+    typed_attribute :name, Types::String.optional
+    typed_attribute :social_image, Types::Serializer::URL
```
ConnorChristensen: Do you happen to know what type the `social_image` is It looks like it's just a single URL, but I'm not confident.

api/app/serializers/v1/text_section_serializer.rb
zdavis: I would say: "A JSON representation of the text section's HTML contents"

api/app/serializers/v1/text_section_serializer.rb
zdavis: Yes, It's a single URL. See `project.social_image`. The TextSection delegates the request to the Text, and the Text delegates it to the project.

api/app/serializers/v1/text_section_serializer.rb
ConnorChristensen: Sounds good.

api/app/serializers/v1/text_section_serializer.rb
ConnorChristensen: Thank you!

PR #2583, opened 2020-01-22
[WIP] Document Statistics API endpoints

ConnorChristensen:
This PR:
* Documents the Statistics API endpoints
* Creates a new optional parameter called `instantiate_before_test`, which allows the default behavior of creating a resource with FactoryBot before attempting a `show` request. In this case, the app generates the statistics and there is no factory that would create a statistics database entry, so `instantiate_before_test: false` turns off the attempt to create a statistic.
* To make the response success descriptions line up a little better with the route summary (since they are usually very closely related), if the route does not have a custom success description, it will fall back on the provided summary, and finally the default success description

### DIFF
--- a/api/spec/requests/api/v1/twitter_queries_spec.rb
+++ b/api/spec/requests/api/v1/twitter_queries_spec.rb
@@ -23,7 +23,7 @@ RSpec.describe "Twitter Queries", type: :request do
                         model: TwitterQuery,
                         authorized_user: :admin,
                         tags: "Projects",
-                        summary: "Gets all twitter queries associated with a project",
+                        summary: "Returns all twitter queries associated with a project",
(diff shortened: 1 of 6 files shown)


api/spec/requests/api/v1/twitter_queries_spec.rb
```diff
@@ -23,7 +23,7 @@
                         model: TwitterQuery,
                         authorized_user: :admin,
                         tags: "Projects",
-                        summary: "Gets all twitter queries associated with a project",
+                        summary: "Returns all twitter queries associated with a project",
```
ConnorChristensen: This change was made to be more consistent with the other summaries in the documentation.

zdavis:
This looks good, but there's a merge conflict to resolve.

PR #2590, opened 2020-01-24
[WIP] Document /tokens API endpoint

ConnorChristensen:


api/spec/support/requests/authenticated_request.rb
```diff
@@ -29,7 +29,7 @@ def get_user_token(user_type)
 
   Role::ALLOWED_ROLES.each do |role|
     let("#{role}_email".to_sym) { "#{role}@castironcoding.com"}
-    let(role.to_sym) {
+    let!(role.to_sym) {
```
ConnorChristensen: I made this change so the admin user exists in the database before the tokens route searches for it. Is this a good way to go about accomplishing this task, or is this a brute force solution?

Is there a way to make sure that an admin user is created in the database just for the purposes of the /tokens route test?

api/spec/requests/api/v1/tokens_spec.rb
```diff
@@ -0,0 +1,33 @@
+require "swagger_helper"
+
+RSpec.describe "Tokens", type: :request do
+
+  include_context("authenticated request")
+
+  before(:each) {
```
zdavis: @ConnorChristensen you asked me how to avoid turning the `let` into `let!` — I'd use a before callback that runs before each assertion to touch the admin model.

ConnorChristensen: Nice! Thanks for the tip!

PR #2598, opened 2020-01-27
[F] Expose UI for project export & preservation

zdavis:


zdavis:
I left a `binding.pry` statement in api/app/controllers/api/v1/export_targets_controller.rb. Don't let that trip you up.

client/src/backend/containers/export-targets/Edit.js
```diff
@@ -0,0 +1,207 @@
+import React, { PureComponent } from "react";
```
zdavis: This was copy-pasted — clean it up or replace with an empty component.

client/src/backend/containers/project/Export.js
```diff
@@ -0,0 +1,149 @@
...
+    return (
+      <Authorize
+        entity={project}
+        ability="managePermissions"
+        failureNotification
+        failureRedirect={lh.link("backendProject", project.id)}
+      >
+        <FormContainer.Form
+          className="form-secondary"
+          name="name"
```
zdavis: name prop should be tied to a requests string. Clearly should not be "name". For example:

```
name={requests.beProjectExportationsCreate}
```

client/src/backend/containers/project/Export.js
```diff
@@ -0,0 +1,149 @@
...
+          model={{
+            attributes: {
+              project_id: project.id,
+              export_target_id: ""
+            }
+          }}
+          update={() => null}
```
zdavis: Can this be left blank? Does the form container take issue with this being empty?

client/src/backend/containers/project/Export.js
```diff
@@ -0,0 +1,149 @@
...
+          create={projectExportationsAPI.create}
+          onSuccess={() => this.fetchExportations(1)}
+          doNotWarn
+        >
+          <Form.FieldGroup label="New Project Export" horizontal>
+            <Form.Select
+              name="attributes[export_target_id]"
+              label="Export Target for the Export"
```
zdavis: Simplify label: "Export Target"

client/src/backend/containers/project/Export.js
```diff
@@ -0,0 +1,149 @@
...
+              options={this.exportTargetSelectOptions()}
```
zdavis: make the method a get accessor and remove the ()

client/src/backend/containers/project/Export.js
```diff
@@ -0,0 +1,149 @@
...
+            />
+            <Form.Save text="Export" wide={false} />
+          </Form.FieldGroup>
+        </FormContainer.Form>
+        <EntitiesList
+          entityComponent={ExportationRow}
+          entityComponentProps={{ active }}
+          title="Project Exportations History"
```
zdavis: Change title to "Past Project Exports"

client/src/backend/containers/project/Export.js
```diff
@@ -0,0 +1,149 @@
...
+          titleStyle="section"
+          showCount
+          pagination={pagination}
+          callbacks={{
+            onPageClick: this.pageChangeHandlerCreator
+          }}
+          entities={exportations}
+          unit="export targets"
```
zdavis: Need unit and plural unit, which should be "export" and "exports"

client/src/backend/containers/records/export-targets/Edit.js
```diff
@@ -0,0 +1,112 @@
...
+  static propTypes = {
+    match: PropTypes.object.isRequired,
+    confirm: PropTypes.func.isRequired,
+    dispatch: PropTypes.func.isRequired,
+    exportTarget: PropTypes.object,
+    history: PropTypes.object.isRequired
+  };
+
+  static defaultProps = {
+    exportTarget: {}
```
zdavis: Default should be null not an empty hash. An empty hash is not a state this prop would ever have. It could be there (in which case it would be a full hash) or it could be absent, in which case it would be null.

client/src/backend/containers/records/export-targets/Edit.js
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  componentDidUpdate(prevProps) {
+    if (prevProps.match.params.id !== this.exportTargetID) {
+      this.fetchExportTarget(this.exportTargetID);
+    }
+  }
+
+  componentWillUnmount() {
+    this.props.dispatch(flush([requests.beExportTargetUpdate]));
```
zdavis: You're flushing a request that's not really generated in this component. You should be flushing requests. beExportTarget

client/src/backend/containers/records/export-targets/Edit.js
```diff
@@ -0,0 +1,112 @@
...
+
+  fetchExportTarget(id) {
+    const call = exportTargetsAPI.show(id);
+    const exportTargetRequest = request(call, requests.beExportTarget);
+
+    this.props.dispatch(exportTargetRequest);
+  }
+
+  render() {
+    if (!this.exportTarget) return null;
```
zdavis: With your current default prop, this won't ever be null.

client/src/backend/containers/records/export-targets/Form.js
```diff
@@ -0,0 +1,96 @@
...
+import { camelCase } from "lodash";
+import Form from "global/components/form";
+import FormContainer from "global/containers/form";
+
+export default function ExportTargetsContainerForm(props) {
+  const { model, onSuccess, options } = props;
+  return (
+    <FormContainer.Form
+      model={model}
+      name={requests.beExportTargetUpdate}
```
zdavis: When I'm reusing a form like this, I'll try to pass in prop that tells the form if it's new or edit and use the appropriate name based on that prop. Alternately, you can check if this.props.model has an ID or not as a way to see if you're creating or updating.

client/src/backend/containers/records/export-targets/Form.js
```diff
@@ -0,0 +1,96 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { exportTargetsAPI, requests } from "api";
+import { camelCase } from "lodash";
+import Form from "global/components/form";
+import FormContainer from "global/containers/form";
+
+export default function ExportTargetsContainerForm(props) {
```
zdavis: Would rather you followed general manifold style and went with a class component instead of a functional component.

client/src/backend/containers/records/export-targets/Form.js
```diff
@@ -0,0 +1,96 @@
...
+      name={requests.beExportTargetUpdate}
+      update={exportTargetsAPI.update}
+      create={exportTargetsAPI.create}
+      className="form-secondary"
+      notificationScope="drawer"
+      onSuccess={onSuccess}
+      options={options}
+    >
+      {getModelValue => {
+        const strategy = camelCase(getModelValue("attributes[strategy]"));
```
zdavis: Once this is a class, move this into a class instance method.

client/src/backend/containers/records/export-targets/List.js
```diff
@@ -0,0 +1,93 @@
...
+  static propTypes = {
+    exportTargets: PropTypes.array,
+    exportTargetsMeta: PropTypes.object,
+    match: PropTypes.object.isRequired,
+    route: PropTypes.object.isRequired,
+    dispatch: PropTypes.func.isRequired
+  };
+
+  static defaultProps = {
+    exportTargets: [],
```
zdavis: I'd have it be null. Empty array is a confusing state for not having received data from the server.

client/src/backend/containers/records/export-targets/List.js
```diff
@@ -0,0 +1,93 @@
...
+    exportTargetsMeta: {}
```
zdavis: See comment above. Should not have a default prop on this.

client/src/backend/containers/records/export-targets/List.js
```diff
@@ -0,0 +1,93 @@
...
+    return (
+      <>
+        {childRoutes(route, { drawer: true, drawerProps })}
+        <EntitiesList
+          entityComponent={ExportTargetRow}
+          entityComponentProps={{ active: active || "" }}
+          title="Manage Export Targets"
+          titleStyle="bar"
+          entities={exportTargets}
+          unit="export target"
```
zdavis: Need plural unit and singular unit

client/src/backend/routes.js
```diff
@@ -424,6 +424,13 @@ const routes = {
                 }
               ]
             },
+            {
+              name: "backendProjectExports",
+              exact: false,
+              component: "ProjectExport",
+              path: "/backend/projects/:pId/exports",
```
zdavis: The route here is OK and does not need to be changed to project-exportations. That's because this is part of the user interface, and it's OK for users to think of these as "exports". But, in code, which users do not see, we want to use names that are close to the model/resource names. Subtle, but important distinction.

client/src/helpers/router/navigation.js
```diff
@@ -222,6 +228,13 @@ class Navigation {
         ability: "manageSocials",
         args
       },
+      {
+        label: "Exports",
+        route: "backendProjectExports",
+        entity: project,
+        ability: "manageExports",
```
zdavis: ability name should change.

client/src/global/components/form/Save.js
```diff
@@ -9,17 +9,21 @@ export default class FormSave extends Component {
   static propTypes = {
     text: PropTypes.string,
     cancelRoute: PropTypes.string,
-    theme: PropTypes.oneOf(["frontend", "backend", "reader"])
+    theme: PropTypes.oneOf(["frontend", "backend", "reader"]),
+    wide: PropTypes.bool
```
zdavis: Please remove this wide prop and the associated changes. I don't think the current select + button lockup is working. I realize that once this is removed the button will appear under the input, but that's fine because I'm going to refine the select + button interface in a different direction.

client/src/api/requests.js
```diff
@@ -30,6 +30,14 @@ export default {
   beUserCreate: "backend-user-create",
   beUserUpdate: "backend-user-update",
   beUserDestroy: "backend-user-destroy",
+  beExportation: "backend-exportation",
```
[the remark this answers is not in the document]
zdavis: With the one exception that as far as the user interface goes, we actually do want to call these things "exports". This distinction is subtle, so if you have questions, ask.

client/src/backend/containers/project/Export.js
ConnorChristensen: The form container does take issue with it being empty. It is marked as a [required prop](https://github.com/ManifoldScholar/manifold/blob/master/client/src/global/containers/form/Form.js#L41).

client/src/backend/containers/records/export-targets/List.js
ConnorChristensen: Is it better in this case to set the `defaultProp` value on an `exportTarget` to null, or just to not have a `defaultProp` value at all?

client/src/backend/containers/project/Export.js
zdavis: Ok, I'll think about how to address it in the Form container.

client/src/backend/containers/records/export-targets/List.js
zdavis: Better not to have the default prop value at all if it's null.

client/src/backend/containers/Records/export-targets/Form.js
```diff
@@ -0,0 +1,110 @@
...
+        model={model}
+        onSuccess={onSuccess}
+        options={options}
+        name={this.formKind}
+        update={exportTargetsAPI.update}
+        create={exportTargetsAPI.create}
+        className="form-secondary"
+        notificationScope="drawer"
+      >
+        {getModelValue => this.renderFormContents(getModelValue)}
```
ConnorChristensen: Is this what you meant when you suggested moving this into a class instance method?

client/src/helpers/router/navigation.js
```diff
@@ -274,6 +287,12 @@ class Navigation {
         route: "backendRecordsFeatures",
         entity: "feature",
         ability: "update"
+      },
+      {
+        label: "Export Targets",
+        route: "backendRecordsExportTargets",
```
ConnorChristensen: I noticed under the Records label on line 57 that there was an array of objects that are also defined here. Am I doing this wrong by adding the Export Targets route information in two places on the same file?

client/src/backend/routes.js
```diff
@@ -514,6 +521,29 @@ const routes = {
...
+              exact: true,
+              component: "ExportTargetsNew",
+              path: "/backend/records/export-targets/new",
+              helper: () => "/backend/records/export-targets/new"
+            },
+            {
+              name: "backendRecordsExportTargetEdit",
+              exact: true,
+              component: "ExportTargetsEdit",
+              path: "/backend/records/export-targets/:id",
```
ConnorChristensen: Some of these paths include an `/edit` at the end (eg. `"/backend/projects/:id/texts/ingestion/:ingestionId/edit"`) but a majority of them do not. I went with what was most common and left off the `/edit` at the end of the path. Is that alright?

client/src/backend/components/list/EntitiesList/Entity/ProjectExportationRow.js
```diff
@@ -0,0 +1,32 @@
...
+    entity: PropTypes.object.isRequired
+  };
+
+  render() {
+    const {
+      entity: {
+        attributes: { createdAt, exportedAt, currentState },
+        relationships: {
+          exportTarget: {
+            attributes: { name: exportTargetName }
```
ConnorChristensen: I haven't seen a lot of advanced es6 destructuring in this project, but I think that in this case, it is a way of assigning variables from this prop that makes it a little easier to read. Any thoughts on this? I want to make sure I'm not breaking any conventions here.

client/src/backend/containers/Records/export-targets/Edit.js
```diff
@@ -0,0 +1,110 @@
...
+      exportTarget: select(requests.beExportTarget, state.entityStore)
+    };
+  };
+
+  static displayName = "ExportTargets.Edit";
+
+  static propTypes = {
+    // export target information is required, but can be null/undefined until
+    // the async network request is fulfilled
+    exportTarget: PropTypes.object, // eslint-disable-line react/require-default-props
```
ConnorChristensen: I know that the eslint rule `react/require-default-props` is turned off for this project, but I think this is a helpful bit of information for any developer that is not familiar with the way manifold uses `mapStateToProps`. When trying to build this component, I did get tripped up that the value of exportTarget would start out as `null`, and then fill in the data after the network request was fulfilled.

client/src/backend/containers/Records/export-targets/List.js
```diff
@@ -0,0 +1,97 @@
...
+    const currentModified = get(this.props, "exportTargets.modified");
+    const previousModified = get(prevProps, "exportTargetsMeta.modified");
+    if (!currentModified) return false;
+    return !(currentModified && previousModified);
+  }
+
+  render() {
+    const { exportTargets, match, route } = this.props;
+    if (!exportTargets) return null;
+    const active = match.params.id || "";
```
ConnorChristensen: Should I do anything to guard against `params` being undefined on the `match` object?

client/src/backend/containers/project/ProjectExportations.js
```diff
@@ -0,0 +1,167 @@
...
+          name={requests.beProjectExportationCreate}
+          model={{
+            attributes: {
+              project_id: project.id,
+              export_target_id: ""
+            }
+          }}
+          update={() => null}
+          create={projectExportationsAPI.create}
+          onSuccess={() => this.fetchExportations(1)}
```
ConnorChristensen: Is it ok that on successful creation of a project exportation, the form fetches the first page so the new project exportation can be seen in the project exportations list immediately after?

client/src/backend/containers/Records/export-targets/Form.js
zdavis: Honestly, looking back at it, I'm not sure what I meant.

client/src/backend/routes.js
zdavis: I think it's fine for now, although I think it's better to include /edit, since it leaves the other path free for a show or detail view in the future should we need it.

client/src/helpers/router/navigation.js
zdavis: This file is confusing—I'm not sure why it was built the way it was. I think what you did is fine, but we need to refactor this navigation at some point.

client/src/backend/containers/Records/export-targets/Edit.js
zdavis: Having this one-off rule disable here doesn't help. I'm sorry you found this confusing, but it hasn't been a problem for other devs on the process. Null is exactly the state that we want the prop to be in, because null tells us that there is no data there. If we wanted to change this behavior/convention, we'd be changing it at a much higher level, not as a one-off on this one component.

client/src/backend/containers/Records/export-targets/List.js
zdavis: You shouldn't need to.

client/src/backend/containers/project/ProjectExportations.js
zdavis: Yes, this is OK.

client/src/backend/containers/Records/export-targets/Form.js
ConnorChristensen: Would you like me to put it back the way it was, or is this a fine change?

zdavis: I can move it back, as I'm continuing to tweak these components currently.

ConnorChristensen: Thank you for finishing this up on my behalf.

PR #2605, opened 2020-01-31
[F] Add pagination to route response documentation

ConnorChristensen:
The first commit adds pagination to the responses and the following two commits just clean up the code. I have been documenting which routes are paginated and this connects the pagination parameters to the documentation.


zdavis:
@connor — there are a number of changes here that don't seem directly related to pagination. I'm happy to look this over, but I've since made my own cleanups to the request helper, so this needs to be rebased on feature/api-docs before I can accept it.

ConnorChristensen:
I removed the code cleanup, so this PR just adds pagination to responses. I will open the unrelated code cleanup in a different PR.

PR #2606, opened 2020-01-31
[F] Document SearchResult API endpoint

ConnorChristensen:


ConnorChristensen:
I'm getting a failure on this test that re-indexing is not working. It looks like it is looking for a connection to an elastic search instance that is not working. I'm not sure how to proceed with this problem in a timely manner. These tests work locally.

I see that the tests for the [search serializer](https://github.com/ManifoldScholar/manifold/blob/master/api/spec/serializers/search_result_serializer_spec.rb#L10) re-indexes for the tests, and those seem to work fine on the github actions test. I originally thought that there might not be an instance of elastic search running, but it seems like the search serializer test shows that it can work. I tried to copy the search serializer as best I could, but there are likely things going on behind the scenes that a more experienced ruby person could figure out faster.

If you have any hints for me, I'm happy to make another attempt at solving the failing test.

zdavis:
We don't run tests that require a connection to ES on CI — we haven't gotten around to getting ES spun up in CI env. Tag the test with :elasticsearch and it won't run on CI. You'll see this in other places in the specs. For example:

```
  context "can be searched", :elasticsearch do
  end
```

api/spec/requests/api/v1/search_results_spec.rb
```diff
@@ -0,0 +1,44 @@
+require "swagger_helper"
+
+RSpec.describe "Search Results", type: :request do
+
+  context "can search a project", :elasticsearch do
```
ConnorChristensen: Did I do this right? It looks like this test is still trying to run and failing.

zdavis:
Rebased on Master.

zdavis:
Changed base to master.

api/spec/requests/api/v1/search_results_spec.rb
zdavis: The tag was correct. The issue was with your spec — you were reindexing models outside of a before block and that code was being executed whether or not the examples were included. Generally, doing things outside of let and before/after blocks in specs is an antipattern and you shouldn't do it.

zdavis: See line 9 below for the correct way to do what you were trying to do.

PR #2624, opened 2020-02-21
[B] Skip some invalid ingestion sources

### LINKED ISSUE #2612: DS_Store Invalidates Manifest Archives
On the Minnesota instance I just reingested a Manifest Archive. The process failed on account of a validation exception. Manifold's error message indicated the file types it was expecting, but not which was flagging the error. I noticed I had a rogue DS Store file in my archive. After removal the reingest worked as expected. Given the prevalence of these to creep up when using hosting services like Google Drive, it would be useful if Manifold could ignore or simply remove the DS Store file(s) present in such uploads.

zdavis:
When ingesting, we can safely skip ingestion sources that are invalid
and of the type “publication_resource”. If we dont skip these, all it
takes is an errant DS_Store file to blow up a manifest ingestion, which
will try to include every file it finds in the archive, whether or not
its referenced in the manifest. This is less of a problem with EPUBs,
which require an explicit reference to every file in the package manifest.

Fixes #2612


tsmyre:
As a matter of context, I actually just encountered the same with an EPUB and a TIFF file. The EPUB was created using Calibre, and it passed through the TIFF, even though that's not technically in line with the EPUB spec. Manifold failed out the ingest generally.

PR #2628, opened 2020-02-26
[WIP] Access Entitlements

scryptmouse:
- [x] Consider collapsing entitlements and permissions into a single tab. Users will be confused by the distinction between the two.
- [x] Think about how to message the "restricted access" switch and where it should go. 
- [x] The entitlement authorizer doesn't have the right logic in it — I don't think it should be referencing the project_resource_editor and should instead be giving the user the right to manage entitlements if they can manage the project the entitlement is being added to
- [ ] Need to wrap some <Authorize> logic around the various entitlement views in the UI so that users who can't create them don't see the option to do so.
- [x] Add a type-ahead input for selecting the user or reading group. We've been missing a good user selection interface for a long time, so we may as well build it. Can also be used on the permissions interface.
- [x] Ensure that we show some more info in the entitlement row—who created the entitlement, when it expires, etc. Need to show an icon for reading groups as well.
- [x] It seems like when I try to access a restricted project directly while I don't have access, I'm redirected to the HP. Wonder where this redirect is coming from. Need to at least notify the user as to lack of access.
- [x] Currently the projects controller is scoping `Project.with_read_ability` on the index and show view. Is this what we want? I was thinking we would show projects that can't be read in the list view so that users could be enticed to purchase the project, and then we'd restrict access in the show view. Need to discuss with team.
- [x] Need to confirm that project child models can't be accessed if the user can't access the project. Should work, because in most cases the children delegate read access to the parent, but would be good to confirm.


client/src/helpers/router/navigation.js
```diff
@@ -181,6 +181,13 @@ class Navigation {
         ability: "managePermissions",
         args
       },
+      {
+        label: "Entitlements",
+        route: "backendProjectEntitlements",
+        entity: project,
+        ability: "managePermissions",
```
zdavis: Probably should be manageEntitlements.

client/src/backend/containers/entitlements/List.js
```diff
@@ -0,0 +1,77 @@
...
+            entityComponentProps={{
+              active,
+              linkName: listUrl
+            }}
+            buttons={[
+              <Button
+                path={lh.link(newUrl, entity.id)}
+                text="Grant new entitlement"
+                type="add"
+                authorizedTo="createPermissions"
```
zdavis: Probably should be createEntitlements

PR #2657, opened 2020-03-13
[F] Allow disabling of library pages

### LINKED ISSUE #2652: Improve Standalone Mode
**Add setting for disabling library mode**

Add a new setting to Manifold that allows an installation to disable Library mode. When library mode is disabled, Manifold will adopt the following behaviors:

- [x] New settings will be exposed: "Library Redirect URL" and "Manifold Home Page Redirect URL"
- [x] Home, Projects, and Following are removed from the primary navigation.
- [x] The Manifold logo will redirect to the Manifold home page redirect URL specified in settings.
- [x] Visits to non-project, non-transactional pages will redirect to the library redirect URL specified in settings.
- [x] All projects will render in standalone mode, regardless of how they are configured.

**Improve Manifold logout mechanism**

Current when users logout of Manifold we redirect them to the home page. We will adjust the logout mechanism so that users remain on the current page. This change will also involve redirecting users to the home page (or home page redirect url) in cases where a project cannot be accessed by an unauthenticated user (such as when it's in draft mode)

- [x] Implement logout adjustments

zdavis:
Fixes #2652


tsmyre:
On Edge the homepage redirect isn't working as expected; instead the HP is redirecting to the target identified in the library redirect field.

zdavis:
Thanks, Terence. Fixed in 581598f

PR #2658, opened 2020-03-14
[A] Improve a11y of DnD interfaces

### LINKED ISSUE #2208: Keyboard navigation inadequate for BE project collection list
The sidebar list of project collections needs improvements to keyboard navigation:
- [x] currently a project collection can't be opened by keyboard
- [ ] ~visibility button loses focus when toggled~
- [x] DnD functionality doesn't work with keyboard. This is a more complex refactor potentially requiring a new DnD library. We'll revisit this at a later time.

### LINKED ISSUE #2415: Project Collections grid is not keyboard accessible
The library we use here is different from our other D&D interfaces. We used a different library because we needed grid drag and drop support. AFAIK, this library is not keyboard accessible, and will likely need to be replaced. I also wonder if a list interface like what we have on the other D&D components isn't a better UI choice anyway.

### LINKED ISSUE #2437: Admin Interface Drag and Drop Interactions Unclear
For content creation in Manifold, Drag and Drop is an important feature. This may be seen in the Layout page for a book. While a keyboard user can interact with these drag and drop elements, there are issues with the interaction for Screen Reader Users

## Expected Behavior
Users should be aware of how to interact with elements, as well as how their interactions affect those elements. For Drag and Drop specifically, a user should be told how to interact with the element. Further, the user should be given status updates, such as if an element is being used, and where it is being moved to.

Hadi recommends adding documentation for this feature, as well as a link to that documentation, so that a reader may understand how it works.

## Current Behavior
A keyboard user may interact with Manifold's Drag and Drops, but it is not initially clear how to do so. A Screen Reader User is told that they are dragging something but not told where it is being dragged to/what elements are moving.

## Steps to Reproduce
1. Navigate to Admin Interface (via Enter Admin Mode)
2. Navigate to an editable project
3. Navigate to the Layout page
4. Navigate to the Layout subsection in the Content Blocks section
5. Navigate to the "Button Draggable" button on one of the Layout elements
6. Note how there are no instructions for how to use this element
7. Press Space and notice how the user is told that they are dragging the element
8. Press up or down arrow keys. Note how the element successfully moves, but a Screen Reader user is told nothing of this interaction.
9. Press space again, and notice how the user is not told that the action has been completed.

## Testing Environment
NVDA 2019.2.1
ESR Firefox 68.2.0 esr

dananjohnson:
@zdavis, I think I’ve taken this PR as far as I can for now. All DnD interfaces now use `react-beautiful-dnd`, which itself has been updated to v13. This adds screen reader announcements for all user interactions with a DnD interface. We can customize the language of these, but I've found the default ones to be solid. I also revamped the Manual Project Collection interface to use a list rather than a grid, made Content Blocks keyboard accessible, and tweaked the CSS for these interfaces to a more consistent appearance during the dragging state.

There are a couple issues left that I could use your eyes on:
1. The re-rendering issue we've discussed in `<SortableEntities />`.
2. Keyboard accessibility remains only partially functional for text category and hero action callout blocks (as discussed [here](https://github.com/ManifoldScholar/manifold/issues/2437#issuecomment-596004827)). I've experimented with your suggestion to capture key events and move the blocks where we want them to go, but didn't have any luck. I was able to capture the event in the `<Draggable />` and pass it up through a function prop up to `<Droppable />` and `<DragDropContext />`, but I couldn't get the necessary information about the position of the block to know where to move it. It seems that this data isn't actually updated until dragging is released. I also tried using the [`onDragUpdate` responder](https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/guides/responders.md), which is called when a block is moved, and actually provides the positional information we would need. The problem is that `onDragUpdate` isn't called when a block can't move, which is exactly the situation we're trying to address. Let me know if you think there's a solution 🤞, or if we need to figure out a different approach.

Resolves #2208
Resolves #2415 
Resolves #2437


zdavis:
#1 addressed in 201f3f4

zdavis:
@dananjohnson please clean up my WIP commit which shows you how I think this should be laid out. There is still an image with the text not lining up (see image) and we should crop the button text at 2 lines:

image

I'll continue to work on the project list—this is real close!

dananjohnson:
Just a heads-up @zdavis—since I don't think the commits fully reflect this right now—that this PR should close #2208, #2415, and #2437.

zdavis:
Ok, this is good to go as soon as tests pass. I've added an issue for keyboard a11y on text lists because we haven not yet addressed that. This PR does, however, improve D&D a11y in every other place, I think.

PR #2665, opened 2020-03-20
[B] Fix missing poster display on resource interactive slides

### LINKED ISSUE #2418: Poster images not displaying in resource slideshow
When I add a poster image to a resource it is not displaying as expected in the resource's slideshow. An example can be found here:

**Frontend**
`https://manifold.umn.edu/projects/cut-copy-paste/resource-collection/humphrey-moseley`.

**Backend**
`https://manifold.umn.edu/backend/projects/resource/7ad9da5f-d382-4272-87b7-b2ba4f68c64a/variants`

ConnorChristensen:
Resolves #2418. This PR involves some refactoring of the components that relate to a resource slideshow.

This work is built on top of the small [PR#2664](https://github.com/ManifoldScholar/manifold/pull/2664). PR#2664 should be merged in before this.


client/src/frontend/components/resource-slide/Slide/Info.js
```diff
@@ -0,0 +1,41 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import FormattedDate from "global/components/FormattedDate";
+
+import classnames from "classnames";
+import IconComputed from "global/components/icon-computed";
+
+export class Info extends Component {
```
zdavis: Please include a display name: `ResourceSlide.Slide.Info`

client/src/frontend/components/resource-player/Iframe.js
```diff
@@ -1,125 +1,32 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
-import throttle from "lodash/throttle";
-// TODO: Fix this cyclical dependency
-// import ResourceSlide from "frontend/components/resource-slide";
-import { closest } from "utils/domUtils";
 
 export default class ResourcePlayerIframe extends Component {
   static propTypes = {
```
zdavis: Add display name to component.

client/src/frontend/components/resource-slide/Slide/index.js
```diff
@@ -0,0 +1,55 @@
+export * from "./Info";
+export Zoom from "./Zoom";
```
zdavis: These exports here are weird. Why are you doing this?

client/src/frontend/components/resource-slide/Slide/Info.js
zdavis: Why is this not the default export?

client/src/frontend/components/resource-slide/Slide/Info.js
```diff
@@ -0,0 +1,41 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import FormattedDate from "global/components/FormattedDate";
+
```
zdavis: Remove unnecessary space.

client/src/frontend/components/resource-slide/Slide/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import Zoom from "./Zoom";
+import { Info } from "./Info";
```
zdavis: Why are these components exported differently? Make consistent please.

client/src/frontend/components/resource-slide/Slide/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+export default class ResourceListSlide extends Component {
+  static propTypes = {
```
zdavis: Missing display name.

client/src/frontend/components/resource-slide/Slide/index.js
```diff
@@ -0,0 +1,55 @@
...
+    } = this.props;
+
+    if (!resource) return null;
+
+    const bgImage = backgroundUrl || "/static/images/resource-splash.png";
+
+    let slideStyles = {};
+    if (React.Children.count(children) === 0 || !!backgroundUrl) {
+      slideStyles = { backgroundImage: `url(${bgImage})` };
+    }
```
zdavis: Too much logic in the render method that's too hard to easily parse. You have a tendency to do this—just move small bits into class instance methods. The act of naming those methods really makes the code easier for the next person to read. For example:

```
get background() {
  return this.props.backgroundUrl || "/static/images/resource-splash.png"
}

get hasBackgroundUrl() {
  return !!this.props.backgroundUrl;
}

get hasChildren() {
  return React.Children.count(children) === 0
}

get hasBackground() {
  if (this.hasChildren) return true;
  if (this.hasBackgroundUrl) return true;
  return false;
}

render() {
  [...]
  const slideStyles = this.hasBackground ? { backgroundImage: url(this.background) } : {} 
}
```

The difference between this and what you wrote is really important to me—my version might be more verbose, but whoever comes to this next can understand the logic without having to really grok the code. Note, too, that when we use method names to explain what's happening here, we see that the logic doesn't actually make much sense. If I read your code right, `this.background` would always have a value (either the provided value or the fallback in /static), yet you're only showing a background if the provided value is present. It seems to me that the logic is wrong, and it should be simplified to this:

```
get background() {
  return this.props.backgroundUrl || "/static/images/resource-splash.png"
}

get hasChildren() {
  return React.Children.count(children) === 0
}

get hasBackground() {
  return this.hasChildren;
}

render() {
  [...]
  const slideStyles = this.hasBackground ? { backgroundImage: url(this.background) } : {} 
}
```

client/src/frontend/components/resource-slide/Slide/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+    return (
+      <figure>
+        <Zoom show={!!zoom} resource={resource} label={zoomLabel} />
+        <div className="figure-default" style={{ ...slideStyles }}>
+          {info && (
+            <Info resource={resource} withBackground={!!backgroundUrl} />
```
zdavis: Per above, refactor these background checks into method names so that it's super clear what conditions need to be met for this to have a background, and so that the actual check is only happening in one place.

client/src/frontend/components/resource-slide/SlideAudio.js
```diff
@@ -3,6 +3,8 @@ import PropTypes from "prop-types";
 import ResourcePlayer from "frontend/components/resource-player";
 
 export default class ResourceListSlideFigureAudio extends Component {
```
zdavis: Class name should be refactored — I don't think you need "figure" in there.

client/src/frontend/components/resource-slide/SlideImage.js
```diff
@@ -1,38 +1,28 @@
...
 
-  static defaultProps = {
-    enableZoom: true
+  static propTypes = {
+    resource: PropTypes.object.isRequired
   };
 
   render() {
-    const attr = this.props.resource.attributes;
+    const background = get(
```
zdavis: Prefer get methods on the class instance to declaring constants in the render method, IMO.

client/src/frontend/components/resource-slide/SlideImage.js
```diff
@@ -1,38 +1,28 @@
...
+      this.props,
+      "resource.attributes.attachmentStyles.medium"
```
zdavis: As I've worked on manifold more, I've come to think that using lodash get is wasteful because it has to do string parsing to traverse the object.

I'd do this instead:

```
get background() {
  const { resource } = this.props;
  if (!resource.attributes.attachmentStyles) return null;
  return resource.attributes.attachmentStyles.medium;
}

File: client/src/frontend/components/resource-slide/SlideInteractive.js [hunk 12]
```diff
@@ -1,21 +1,125 @@
...
+            <ResourcePlayer.Iframe
+              resource={resource}
+              styleProps={{ height: "100%" }}
+            />
+          </>
+        ) : (
+          <Slide
+            resource={this.resource}
+            zoom
+            info={!this.state.poster}
```
zdavis: Unclear why poster should have anything to do with info. Methods would help me understand intent.

client/src/frontend/components/resource-slide/Slide/index.js
ConnorChristensen: It was so I could access these default slide components from other slides, such as the Interactive and Image slides which use the zoom button, however it is more important that I keep the code consistent. I moved the files so the exports could be more consistent with the other slide components.

ConnorChristensen:
### Summary
I addressed all the changes that you requested on the code review (most of them were small ones like some name changes and code cleanups). Some of the components I wrote were overly complex (for example the SlideDefault component), so I simplified them, and I think they will be much more straightforward to understand this time.

I fixed the bug you mentioned at the top where the sizing logic on a resource hero was not tall enough. It involved [setting the hero contents height to 100%](https://github.com/ManifoldScholar/manifold/pull/2665/files#diff-16ecf2f486db264eeb789acc1e79d0feR46) and [making interactive component containers flexible so the iframes fill the container](https://github.com/ManifoldScholar/manifold/pull/2665/files#diff-2d8a032b6c38974a211abeb27e61ee04R106).

### Changes from the original implementation
* Interactive element slides are zoomable. I figured that if there was an element that was too small to be displayed in a slideshow (because the slide window is smaller than the minimum size constraints) that it made sense to be able to open the iframe in a zoom window to get a look at it.
* The bug where posters on interactive resources not being displayed is fixed.

### Some use questions to make sure I understand how the site works

* What happens if a user sets the min height to be larger than 52vh when looking at the resource? Is that something we trust the admins not to do?
* Should the minimum height and width of the iframes be respected on the  "view resource" page?

zdavis:
Rebased on master

PR #2695, opened 2020-04-04
[F] Support phrasal searching

### LINKED ISSUE #2668: Allow users to search for exact phrases on Manifold
Expected Behavior

If I put a phrase that's in quotation marks into Manifold's search bar, then I should be able to find that exact phrase. The phrase should not get broken up. 

Current Behavior

If I am searching a e-book version of War and Peace, and if I plug "with her daughters and a large number of guests" into the search bar, then I get results that bring me nowhere close to those words. Also, the search results do not acknowledge that I put the phrase in quotation marks, which in most search bars means that the phrase should be treated as one unit. It should not be broken into its individual parts. This is an example of how my phrase was broken up: https://www.screencast.com/t/WeyHDnbytie

Steps to Reproduce

1. Go to this Manifold version of War and Peace: https://uw.manifoldapp.org/projects/war-and-peace
2. Plug in "with her daughters and a large number of guests." This should bring you to a line that starts section XVIII. The same results happen with any phrase. What this means is that if a group of people are reading the same Manifold book, then it can be very hard for them to get onto the same page because there are no pages numbers in Manifold. 

Context (Environment)

This happened in a reading group on Twitter. War and Peace has many different translations with many different ways of representing section breaks. A group of people was reading one translation, and another group was reading the Manifold book. We discovered that our chapters were out of alignment, and I wanted to be able to tell people how to search for phrases in order to find an exact section or line.

People who use academic catalogs and databases are very used to being able to find exact phrases (using quotation marks) when they search. If this feature is not in Manifold, then I worry that people will become frustrated.

Highly functional search bars are also crucial when it comes to accessibility. Disabled people rely on effective search bars to go exactly where they want to go instead of having to scramble with different forms of navigation or having to read absolutely everything to find one little thing.

I'm not sure how to add labels here, but I'd say that this needs a "bug" issue and an "accessibility" issue.

scryptmouse:
* Change our approach to use custom elasticsearch
  query based on default searchkick methods
* Add `keyword_search` to parse quoted strings
* Support negated phrases: `-"some phrase"`

Resolves #2668


zdavis:
Looks like this needs some more work. The SearchResults API should return a search result object with a "model" relationship that points to the indexed model. This is then serialized by the SearchResult Serializer (see https://edge.manifoldapp.org/api/v1/search_results?keyword=franklin) for an example.

zdavis:
Searching for negation only throws an exception — eg a search for "-madame".

Phrasal searching works well, but not having much luck with negation.

PR #2712, opened 2020-04-30
[B] Additional a11y fixes to Form.Picker

### LINKED ISSUE #2436: Admin Interface Widget That Adds Badges to a "Bucket" Is Missable by Screen Readers
Apologies for the weird title, below is the element that I am referring to:
image

This element, while it can be interacted with properly, can be missable by a Screen Reader user. For example, the "Add a Subject" box is completely missed, reading only the title, and the contents of the bucket (as it is labelled in the HTML).

## Expected Behavior
All elements of a page should be read in proper succession by a Screen Reader while browsing the page.

## Current Behavior
The widget that adds badges to buckets (eg, the "Add a Subject" field in Taxonomy) are missed by Screen Readers Browsing the page.

## Steps to Reproduce
1. Navigate to Admin Interface (via Enter Admin Mode)
2. Navigate to an editable project
3. On the General page, navigate to the Taxonomy Section
4. With a screen reader that has Browse Mode enabled (NVDA Modifier + Space if using NVDA), press the down arrow to read the following contents.
5. Note how the "Add a Subject" field is not read.
6. Navigate to the "Taxonomy" heading (NVDA can navigate headings using H or Shift + H)
7. Press Tab to navigate to the next interactable element. The next element should be the "Add a Subject" field, but focus jumps to the first badge ("Pretty", in my example)

## Testing Environment
NVDA 2019.2.1
ESR Firefox 68.2.0 esr

dananjohnson:
- makes combobox and listbox siblings
- prevents default on up and down arrow key events
- fixes use of `aria-selected` on options
- moves open, reset buttons and live region outside of combobox
- improves button label for selected items

Fixes #2436

Screen Shot 2020-04-30 at 11 31 11 AM


zdavis:
Fixes #2436

PR #2716, opened 2020-05-01
[F] Document Entitlements

ConnorChristensen:
Only two things out of the ordinary:

1. The rswag docs were out of sync when I started, so there is a single commit that just updates the docs before I made any documentation changes.
2. There are three GET requests relating to entitlements, `/entitlements`, `/entitlements/{id}` and `/projects/{project_id}/relationships/entitlements`. Only `/entitlements/{id}` requires authorization, out of these three GET requests. The reasoning behind restricting this route does not seem obvious to me, so if you happen to know why the route is restricted, I can add that to the route documentation.

### DIFF
--- a/api/app/serializers/v1/entitlement_serializer.rb
+++ b/api/app/serializers/v1/entitlement_serializer.rb
@@ -4,7 +4,8 @@ module V1
 
     ID = Types::Serializer::ID.meta(read_only: true)
 
-    CURRENT_STATE = Types::Coercible::String.enum(EntitlementState.map(&:to_s)).meta(example: "active", read_only: true)
+    ENTITLEMENT_STATE_OPTIONS = EntitlementState.map(&:to_s)
+    CURRENT_STATE = Types::Coercible::String.enum(*ENTITLEMENT_STATE_OPTIONS).meta(example: "active", read_only: true)
 
     typed_belongs_to :entitler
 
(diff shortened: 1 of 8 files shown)


api/app/serializers/v1/entitlement_serializer.rb
```diff
@@ -4,44 +4,43 @@ class EntitlementSerializer < ManifoldSerializer
 
     ID = Types::Serializer::ID.meta(read_only: true)
 
-    CURRENT_STATE = Types::Coercible::String.enum(EntitlementState.map(&:to_s)).meta(example: "active", read_only: true)
+    ENTITLEMENT_STATE_OPTIONS = EntitlementState.map(&:to_s)
+    CURRENT_STATE = Types::Coercible::String.enum(*ENTITLEMENT_STATE_OPTIONS).meta(example: "active", read_only: true)
```
ConnorChristensen, line 8: I changed this because the enum method takes several parameters rather than an array of strings.

api/spec/requests/api/v1/entitlements_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  context "for a project" do
+    path "/projects/{project_id}/relationships/entitlements" do
+      include_examples "an API create request",
+        parent: "project",
+        model: Entitlement,
+        url_parameters: [:project_id],
+        authorized_user: :admin,
+        description: "Creates an entitlement to be associated with the project "\
+        "ID provided in the endpoint. The subjectUrl in this case is not needed." do
```
ConnorChristensen, line 47: The route will always accept a subjectUrl as an argument, but it has no function in this route, so I just documented that it is not needed here. Is that overkill? Should the subjectUrl have documentation in the resource information at the bottom of the page and then leave the user to understand when it is required and when it is unneeded data?

scryptmouse: I wouldn't document it as part of this route. Parameters that are accepted but not used don't need to be documented.

api/app/serializers/v1/entitlement_serializer.rb
scryptmouse: Thank you for catching that!

zdavis:
Added authorization to entitlement index actions.

PR #2777, opened 2020-06-15
[C] Remove unneeded wrapper from BE ProjectsList

SMaxOwok:
As I was getting the app set up on my local, I noticed these two backend pages (Projects/Records) were rendering with slightly different padding.

As far as I can tell, `Layout.BackendPanel` isn't doing much in this case and seems like it shouldn't be here.  Other way to go might be adding that wrapper to these record child components.

projects


zdavis:
Hey @dananjohnson Could you review this when you get a chance? Thanks!

PR #2786, opened 2020-07-09
[F] Update Rails and Shrine

zdavis:


zdavis:
Picking this work up in #2813.

PR #2790, opened 2020-07-15
[B] Truncate project collection titles that are too long

### LINKED ISSUE #2764: Project collection title doesn't wrap correctly
The names of project collections don't wrap to their container:

collection-wrap

ConnorChristensen:
Resolves #2764

### DIFF
--- a/client/src/theme/styles/components/backend/project-collection/_collection-list.scss
+++ b/client/src/theme/styles/components/backend/project-collection/_collection-list.scss
@@ -41,7 +41,7 @@
   &__inner {
     @include draggable;
     display: flex;
-    align-items: baseline;
+    align-items: center;
     justify-content: space-between;
     width: 100%;
 
@@ -57,8 +57,10 @@
 
   &__link {
     @include buttonUnstyled;
+    display: flex;
     flex-grow: 1;
     padding: $_inner-padding $_inner-padding $_inner-padding $_inner-padding * 2;
+    overflow: hidden;
     text-align: left;
 
     .project-collection-list-item--selected & {
@@ -82,13 +84,11 @@
   }
 
   &__item-text {
+    @include textTruncate;
     flex-grow: 1;
     padding: 3px 0;
     margin-right: 6px;
-    overflow: hidden;
     text-align: left;
-    text-overflow: ellipsis;
-    white-space: nowrap;
   }
 
   &__count {
@@ -126,6 +126,4 @@
     padding-right: 3px;
     padding-left: 3px;
   }
-
-  &__icon {}
 }
--- a/client/package.json
+++ b/client/package.json
@@ -19,7 +19,7 @@
     "start-docker": "npm-run-all _build:browser-config _serve:ssr",
     "watch": "BUILD_IN=tmp/client/ npm-run-all _watch:banner --parallel _watch:ssr _watch:browser _build:ssr:rescue _serve:ssr:rescue",
     "test": "jest --verbose",
-    "test-ci": "jest --logHeapUsage",
+    "test-ci": "jest --maxWorkers=2 --logHeapUsage",
     "fix": "npm-run-all \"fix:js {*}\" \"fix:scss {*}\" --",
     "lint": "npm-run-all \"lint:js {*}\" \"lint:scss {*}\" --",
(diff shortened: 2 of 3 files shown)


ConnorChristensen:
Thank you @dananjohnson. I followed the example of the [header-reader](https://github.com/ManifoldScholar/manifold/blob/master/client/src/theme/styles/components/reader/_header-reader.scss#L80) CSS to get the ellipsis working. 

My initial instinct was just to move the `overflow`, `text-overflow`, and `white-space` rules into the containing `project-collection-list-item__link` class. Could you tell me why the Manifold project applies a `flex` rule with `overflow: hidden` on the parent and then the `overflow`, `text-overflow`, and `white-space` rules on the child instead of just putting those three rules on the parent class? I definitely feel like I'm missing something.

#### An example of my question

Manifold way
```
<a class='parent'><span class='child'>text</span></a>

.parent {
  width: 50px;
  display: flex;
  overflow: hidden;
}

.child {
  overflow: hidden;
  text-overflow: ellipsis;
  white-space: nowrap;
}
```

My initial attempt
```
<a class='parent'><span class='child'>text</span></a>

.parent {
  width: 50px;
  overflow: hidden;
  text-overflow: ellipsis;
  white-space: nowrap;
}
```

zdavis:
@ConnorChristensen I think that if you cherry-pick cd81413098661c57c92d92a324058f2a99d9c4f0 into this the client tests should pass.

client/src/theme/styles/components/backend/project-collection/_collection-list.scss
```diff
@@ -41,7 +41,7 @@
   &__inner {
     @include draggable;
     display: flex;
-    align-items: baseline;
+    align-items: center;
```
dananjohnson, line 44: What was the reason for changing this value?

client/src/theme/styles/components/backend/project-collection/_collection-list.scss
```diff
@@ -85,10 +87,8 @@
     flex-grow: 1;
     padding: 3px 0;
     margin-right: 6px;
-    overflow: hidden;
     text-align: left;
-    text-overflow: ellipsis;
-    white-space: nowrap;
+    @include textTruncate;
```
dananjohnson, line 91: CIC convention is to add the mixin before other CSS properties in a selector. This is so that you can override the mixin with other declarations if need be (ideally not, but it happens), rather than the mixin overriding your declarations.

dananjohnson:
@ConnorChristensen you raise a good question. I wish I had a good answer for you! I'm not sure I'd call the approach you identified the "Manifold way" so much as the approach we've used in certain instances in the application. But there's no hard and fast rule for how we handle text truncation. In fact, I'd say your solution is generally better, since it's simpler. In these cases, you might be encountering older geological layers of Manifold's CSS, or what I suspect are situations where we needed to make a slight tweak to the vertical alignment of the text, and decided to create an inner node for the text and adjust its position.

client/src/theme/styles/components/backend/project-collection/_collection-list.scss
ConnorChristensen: Sounds good!

client/src/theme/styles/components/backend/project-collection/_collection-list.scss
ConnorChristensen: If you set `align-items: baseline` the project collection list the project collection name looks just a little higher up than it should be:

baseline style screenshot

Before I started my work on this, all content in the project-collection-list-items were centered vertically and I wanted to maintain that, so I set the alignment to be centered. It appears that changing the `project-collection-list-item__link` to `display: flex` to limit the length of the span, bumps the text up about 2 pixels and without the centering property, it is about 2 pixels out of place.

dananjohnson: OK that's fine. You should get your fonts working locally, btw. Using the wrong fonts will make the values in your CSS inaccurate because the glyphs don't have the same shape and size.

ConnorChristensen:
I squashed my fixup commit and this PR is ready to go.

PR #2793, opened 2020-07-18
[F] Add new project collection header styles

ConnorChristensen:
I believe this satisfies the requirements laid out in the design doc for the new project collection header styles. This is opened as a draft PR as API changes need to be completed to support the UI changes here. After those changes are complete, then a more thorough round of testing can be done and approved.

I was told that there would be some additional refactoring to do regarding generalizing collections, so I did not take a long time to keep the SCSS as DRY and clean as could be. 

There are three storybook additions corresponding to these changes, a project collection header, detail and summary view. These storybook additions each come with storybook knobs that allow the viewer to see every possible interface view for a project collection that Manifold will allow. The only view that will look strange in the storybook context is the full bleed hero image, which is designed to take up the full view width of the screen, and extends outside the dashed line container for stories.

So feel free to skim over the code and check out the storybook views, but until this is tested in a Manifold environment with the supporting API changes, no need to go into too much detail, since it may change with more testing.


zdavis:
Thanks, @connor. I didn't have a chance to review this. You and @dananjohnson should work together this week to get this ready for me to accept next week. @dananjohnson, please work with Connor to make sure the code is DRY and generally in keeping with our goal to simplify styles where possible. If Connor has time and wants to move ahead this week with getting all similar FE headers to a place where they're sharing a single DRY component, I'm all for that.

client/src/frontend/components/project-collection/Constants.js
```diff
@@ -0,0 +1,9 @@
+const SQUARE = "square_inset";
+const WIDE = "wide_inset";
+const FULL = "full_bleed";
+
+export default {
+  SQUARE,
+  WIDE,
+  FULL
+};
```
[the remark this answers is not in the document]
ConnorChristensen: I also use it in the file that generates a mocked project collection over in [client/src/test/fixtures/entities/project-collection.js](https://github.com/ManifoldScholar/manifold/blob/cc/project-collection-headers/client/src/test/fixtures/entities/project-collection.js#L21). The spec I was basing my work off said the project collection has an enum value for the hero type that defaults to the square layout. I didn't want to repeat the string that denotes the square value in more than one file in case the value needs to be changed in the future, but would be happy to hard code it if that is considered overkill.

ConnorChristensen:
Thanks for the detailed review! I addressed your comments one-by-one and added a fix-up for each comment so it should be easy to track the changes and see that I noticed the feedback. The final fix-up is a general cleanup and refactor of the react component and SCSS. I will be adding one more fix-up after this message for the updated snapshots.

As far as I know, there is only one style bug left in this refactor (when I remove the description in storybook the grid-gap remains between the title and the now empty description cell, giving too much white space on the bottom). I will fix that and do a little more testing and post another message here when I am confident that it is ready for review again.

dananjohnson:
Thanks, @ConnorChristensen! Let me know when you're all ready for another review.

ConnorChristensen:
Alright, @dananjohnson I am ready for another review. My only concerns are:

* My changes to fix the style bug in 70471724852b0421407ce134956d36e9feae86bd feel like a solution, but not the best one. It feels a bit fragile. I would like some advice on how you would approach this problem if you see a better solution.
* When viewing the square hero layout with no icon on the header, there looks to be a little too much white space around the header, but I would like a second opinion on that.

client/src/theme/styles/components/frontend/project-collection/_header.scss
```diff
@@ -0,0 +1,146 @@
...
+$_hero-max-height: 340px;
+$_font-family: 'sofia-pro', 'trueno', sans-serif;
+
+.project-collection-header {
+  display: grid;
+  grid-template-areas:
+    'title'
+    'description';
+  grid-gap: 30px;
+  margin-bottom: 30px;
```
dananjohnson, line 11: This margin shouldn't be a property of this component, but rather of whatever component that wraps it. This might not be achievable until we have a layout component for all the frontend entities, but keep it in mind as you build in that direction. Eventually, the goal will be to have something like this:
```scss
.entity-wrapper {
  display: grid;
  grid-template-areas:
    'header'
    'filters'
    'body'
    'pagination';
  grid-gap: 30px;
}
```

client/src/theme/styles/components/frontend/project-collection/_header.scss
```diff
@@ -0,0 +1,146 @@
...
+  &__icon {
+    width: $_icon-size;
+    height: $_icon-size;
+    margin-right: 14px;
+
+    > .manicon-svg {
+      position: relative;
+      flex-shrink: 0;
+      width: $_icon-size;
+      height: $_icon-size;
```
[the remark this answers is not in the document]
ConnorChristensen: Good point! Thank you.

client/src/theme/styles/components/frontend/project-collection/_header.scss
ConnorChristensen: Very nice. I felt pretty stumped there, but that makes lots more sense to have a parent take care of that spacing.

ConnorChristensen:
Thanks!

> I hope that that approach to handling the layout variants was valuable to you.

It was! I didn't realize that using CSS grid was alright for Manifold. I was under the impression that CSS grid was still just a little too new to use. Even though the amount of code was about the same, it feels a little more solid and made the HTML more simple.

> mostly related to inconsistencies with the comps

Sorry about that. I got a caught up at the end with making sure that the icon view looked the same as before I started my work. I have recently been focusing on making my changes as minimally impactful as I can while still accomplishing my goal. I should have been focusing more on the comps and noticing that the comps required some changes to the current views, not just additions of new layouts.

> the square and medium-sized hero images shouldn't overflow their containers.

Ah, yes. Since the comps did not show what some of these views would look like on very small screens, I just assumed that all images should maintain a height of 340px and cover the wrapper div completely, without changing the aspect ratio of the images. After you pointed this out, I went back and played around with it some more and I think you will like what I came up with. I kept a wrapper around all images since it is a little easier on the markup to be consistent, but for both the wide and square heroes the wrapper should not be doing any cropping at any screen size, provided the images come in properly sized.

I think this should be ready to review again, but I'm going to look over this work again first thing in the morning just in case. I always seem to catch some new problem with a fresh set of eyes.

client/src/theme/styles/components/frontend/project-collection/_header.scss
```diff
@@ -78,22 +66,23 @@ $_font-family: 'sofia-pro', 'trueno', sans-serif;
   &__title {
     margin: 0;
     font-family: $_font-family;
-    font-size: 22px;
+    font-size: 26px;
     font-weight: 500;
-    hyphens: none;
     line-height: 1.32;
+    color: #363636;
```
dananjohnson, line 72: There should be a SCSS or CSS variable for this color.

dananjohnson: Please confirm for any other color values that you're hard-coding currently.

client/src/theme/styles/components/frontend/project-collection/_header.scss
```diff
@@ -105,32 +94,26 @@ $_font-family: 'sofia-pro', 'trueno', sans-serif;
...
   &__hero-image {
-    flex-shrink: 0;
-    min-width: 100%;
-    min-height: 100%;
+    width: 100%;
+
+    &--square {
+      width: auto;
+      height: 100%;
+    }
```
dananjohnson, line 107: This is still overflowing even though the image has the correct aspect ratio.

Screen Shot 2020-07-23 at 10 27 36 AM

I think the image wrapper approach is overcomplicating things here. You could omit the wrapper entirely and just set a fixed pixel size for the image and use `object-fit: cover` to make sure the image isn't stretched.

```scss
img {
  width: 340px;
  height: 340px;
  object-fit: cover;
}
```

The same approach should work for all three layouts.

ConnorChristensen:
I found two inconsistencies in my implementation and one inconsistency in the design. In the sketch file (Manifold UI Design v2.8.5) there is a 30px space between the title and description in every version except for the custom icon version, where there is a gap of 28px. If that is intentional I can adjust the code to fit that, but it would be much easier to have a consistent 30px gap across all designs.

dananjohnson:
@ConnorChristensen keep it consistent at 30px. I think there's maybe just a little whitespace around the icon.

client/src/theme/styles/components/frontend/project-collection/_header.scss
ConnorChristensen: I found two hard coded colors in the _variables.scss file and will be sure to use those color variables from here on out.

client/src/theme/styles/components/frontend/project-collection/_header.scss
ConnorChristensen: Huh. I did not see the image overflow in any of my tests after my most recent changes, but I went ahead with the changes you suggested and they work very well.

ConnorChristensen:
`object-position` added in and commits have been squashed.

dananjohnson:
I just noticed one more thing 🙃: the hover style on the collection header isn't showing up anymore.

ConnorChristensen:
I pushed a fix, but I'm not particularly satisfied with it. I'm confused about how Manifold has both SCSS and CSS variables mixed in, and would like some help understanding how that was intended to work. It seems important to use the `@include defaultHoverStyle;` but that gave the title two different colors on hover depending on its context and that seemed undesirable.

dananjohnson:
Good question, @ConnorChristensen. The CSS variables came out of our work to address color contrast issues in Manifold. Many colors on the frontend in particular had to be adjusted. As part of that work, we went through the application and removed as many individual color declarations as possible, replacing them with just a handful of CSS variables that get [redefined contextually](https://github.com/ManifoldScholar/manifold/blob/master/client/src/theme/styles/base/_appearance.scss#L54-L89) (frontend vs backend, light vs dark reading modes, etc). We also removed hover and focus styles from as many places as we could in favor of having default styles just [applied to HTML element selectors](https://github.com/ManifoldScholar/manifold/blob/master/client/src/theme/styles/base/_resets.scss#L55).

The general rule is as much as possible to avoid declaring colors and interaction styles for single selectors, and lean on the defaults instead. But there are cases when markup and/or styling are a bit too complex to, for instance, just change the color of an `a` tag on hover. The PC header is an example of that, since what needs to change is the color of an `h2` nested inside the anchor. So in this case, I would apply the `defaultHoverStyle` mixin to the selector `.project-collection-header__link:hover .project-collection-header__title`. Does that help?

zdavis:
@ConnorChristensen @dananjohnson Would like to start my review of this at some point today / early tomorrow if possible. @ConnorChristensen, think you can wrap this up today?

ConnorChristensen:
Yeah, that makes sense! @dananjohnson Thank you for the explanation. @zdavis I will get this wrapped up shortly and will post a message when ready.

ConnorChristensen:
@zdavis @dananjohnson This is ready for review. I pushed a fixup that cleaned up the SCSS some more and fixed a bug where long titles overflowed the title grid area on square hero layouts when close to the break-point. 

A new thing to note is that the hover color for the title depends on the context in which it is used. This only affects the storybook stories. The summary story will have a darker green color and the header story will show a brighter green color since it is using the fallback color value.

If you would like me to squish the commits before review I am happy to do so.

zdavis:
Thanks, Connor. Please do squash it into one commit.

client/src/theme/styles/components/frontend/project-collection/_header.scss
```diff
@@ -0,0 +1,141 @@
...
+
+    &--link {
+      text-decoration: none;
+
+      &:visited {
+        color: $neutral90;
+      }
+
+      &:hover {
+        @include defaultHoverStyle;
```
dananjohnson, line 75: @ConnorChristensen it looks like this declaration is redundant with the default hover color of `a` tags. As I mentioned, try to use the defaults whenever possible to keep the code DRY.

ConnorChristensen: Good catch. Thank you.

ConnorChristensen:
Commits have been squashed and the redundant declaration has been removed.

PR #2828, opened 2020-08-21
[WIP] Support keyboard annotation

### LINKED ISSUE #2326: Reader highlighting / annotating is keyboard inaccessible
This is a complex design issue with no clear precedent for how the keyboard interaction should behave, however we think the highlighting / annotating is a great feature and everyone should be able to use it!

## OBSERVED BEHAVIOR:
Manifold’s annotation feature is completely inaccessible by keyboard. There is no way to highlight the text, or bring up the options for annotation. Even if the options menu has been opened by a mouse, attempting to reach the menu using a keyboard dismisses it. As for previously made annotations, there is also no way to interact with those annotations using a keyboard from the text. In fact, they are not even announced as annotations, leading a non-visual user to perceive annotated text as simple plain text when reading.

## EXPECTED BEHAVIOR:
There is no true precedent for how this kind of annotation feature should interact. However, we have discussed ideas for the issue with many accessibility experts and developers, and have gathered a few ideas.

First, the text could be placed in a sort-of read-only form, content editable div, or text editing field. This would allow for a cursor, which would make highlight the text possible for keyboard users.

With this highlighting behavior, differing ideas were proposed for the pop-up menu. Some developers proposed that the menu would appear once shift was released, similar to how the menu would pop up when the mouse was released. The menu could then be dismissed completely using escape, or using shift to continue highlighting more text. 

There are definitely multiple ways to implement this in a accessible way.

## STEPS TO REPRODUCE:
1. Navigate to a published project (eg https://edge.manifoldapp.org/projects/the-black-tulip)
2. Navigate to the project’s main text, and move to a section with previous annotations (eg, the “A Grateful People” section of “The Black Tulip”)

### Mouse Functionality:
3. Move the mouse cursor over a portion of the text, and click and drag. Notice how the text is highlighted.
4. Release the mouse cursor, and notice how a context menu pops up for options (share, resource, annotate, highlight) for that selected text.
5. Attempt to reach this context menu with tab, and note how the menu is dismissed, making it unreachable.
5. Click on a previously annotated line of text, and note how there is a pop up window for interacting with that previous annotation. Note how there is no announcement for this popup window opening.

### Keyboard Functionality:
6. Read the text on the page with a screen reader by using arrow keys to navigate through the text. Note how text that has been highlighted or underlined as an annotation is not announced as an annotation, responding just like regular text.
7. Note how, if there are links in the text, tab can be used to navigate to those links. Note however, that tab cannot be used to navigate to the underlined text for previous annotations, despite the fact that they can be interacted with.
8. Attempt to highlight the text using keyboard keys: shift + arrow keys for individual characters, and shift + ctrl + arrow keys for full words. Note how, with no cursor, there is no indication for the user’s position.
9. Finally, note how there is no way to bring up the annotations options menu (share, resource, etc) by keyboard

## IMPLEMENTATION POC
```html
<div contenteditable oncut="return false"
  onpaste="return false"
  onkeydown="var key = event.keyCode; if(key == 37 || key == 39 || key == 9) return true; return false;">
  this is the text of the book yes this is the text of the book yes this is the text of the book yes this is the text of the book yes this is the text of the book yes this is the text of the book yes
</div>
```
* Until focused, the text looks like regular static text
* Allows keyboard users to navigate through the text easily with keyboard controls
* Allows keyboard users to highlight
* Just a proof of concept, still needs to allow for page up, page down, home, end, up arrow, down arrow and all other navigation keys, still needs popup functionality
* Not sure how this will work with rich text features like links, previous comments and annotations, resources. 
* Works in JsFiddle, not sure how browsers might handle this differently.

zdavis:
Will resolve #2326


zdavis:
Accidentally closed by PR

dananjohnson:
Closing this and reopening it against my a11y branch.

PR #2840, opened 2020-09-11
[B] Fix missing buttons for annotation and comments

### LINKED ISSUE #2819: Make Edit and Delete buttons more prominent in annotation interface
## Feature Description

Currently, to edit or delete a comment, a user clicks on "flag" in order to reveal the edit and delete buttons; it would be helpful if the edit and delete buttons were displayed directly in the interface, as users may not think of editing and deleting as a function of "flag" so they won't know to click there to find them.

Some users will want to edit or delete annotations they have created, and should be able to quickly and easily find the buttons to do so. 

## User Stories

I am an instructor who is creating an annotated version of a course text to help my students understand a difficult passage. I write a long annotation, but then on re-reading it, want to clarify something I have written. As an instructor, I need easy access to edit and delete my annotations to make sure they are as clear as possible for my students.

I am a student who is being graded on the quality of my annotations. I clicked "save" prematurely and now want to delete my annotation. As a student, it is important that I am able to easily edit and delete my annotations, as I am being graded on them.

ConnorChristensen:
Resolves #2819.

Includes the full serialized attributes for the annotation and comment so the UI knows that the user has permission to update and delete a comment or annotation they created. This fixes a regression introduced in cce22253eb1559f3bec496960bc3bab9af75d0f2 when abilities were simplified for performance reasons.

### DIFF
--- a/api/app/serializers/v1/annotation_serializer.rb
+++ b/api/app/serializers/v1/annotation_serializer.rb
@@ -59,6 +59,10 @@ module V1
 
     class << self
 
+      def include_abilities?(_object, _params)
+        true
+      end
+
       def moderator?(object, params)
         params[:current_user]&.can_update? object.reading_group
       end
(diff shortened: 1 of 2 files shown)


zdavis:
Cherry picking commit into a new branch off of v5.x. Closing this PR.

PR #2898, opened 2020-12-07
[F] Configure Resource Block title and description

### LINKED ISSUE #2882: Changes to resource block
## Problem: At present there is only one kind of button label for the resource block. This is often not sufficient to guide the user to explore what the specific type of resource is available.

scryptmouse:
Resolves #2882


scryptmouse:
I'm not sure if a FE pass will be needed from @dananjohnson, as once I got it configured, it seems to be rendering properly.

crowesn:
Would it be possible to see a screen capture? @konecnymark

zdavis:
image

PR #2906, opened 2020-12-10
[B] Use correct form controls for FE/reader search

### LINKED ISSUE #2759: Suggestions for Search Results Filtering (Checkboxes Are Invisible And Imply Multiple Filters, and Filters Are Applied Instantly)
Manifold's Reader Interface Search Results have a section for "Search Within:" and "Show Results For:". This is reminiscent of search filters. There are a few implementation ideas that would improve both the usability and accessibility of this feature.

Notably, the implementation of switching filters is a bit odd. Visually, there is little indication that a user can interact with these filters. Without a box to contain checkboxes, it appears like these filters are static text, and not interactable elements. An empty box does appear when a keyboard user focuses these filters, but might not realize they could focus the element in the first place, and a visual mouse user never sees this, so they might not realize they can interact with this at all.

Further, while this is implemented as a group of checkboxes, it does not truly behave as such. Checkboxes typically allow users to select multiple options at once, but this instance only allows one item to be selected. As such, this may be better implemented as a group of radio-buttons, which imply that a user can only select one option. Standard checkbox implementation may be found here: [https://www.w3.org/TR/wai-aria-practices/examples/checkbox/checkbox-1/checkbox-1.html](https://www.w3.org/TR/wai-aria-practices/examples/checkbox/checkbox-1/checkbox-1.html)

Finally, this implementation of filters applies the filters as soon as a filter is selected. As such, a user may select a filter without meaning to do so, and lose track of their search results. It would be better for a user to be allowed to choose their filters, and apply those filters via an "Apply Filters" button, so that their filters are applied when they mean to apply them. An example for why this would make sense can be seen through the fact that there are two different filters. A user may want to change both filters, but the first filter they change would be applied as soon as they change it, so the search would have to run twice before the user reaches their desired results.

## Expected Behavior
Checkboxes should allow users to select multiple items. If a user can only select one item, a group of radio buttons should be used instead. The boxes that contains checks should be visible to imply to users that they can interact with this widget. A button should be provided to allow users to apply their filters, and filters should not be applied on changing a filter.

## Current Behavior
Filters are implemented as checkboxes, but the boxes are invisible so a user does not know they can interact with them. Further, a user can only select one option from these checkboxes, though checkboxes imply a user can select multiple items. Finally, filters are applied as soon as a user selects a different filter, which could cause multiple unintended search submissions that could confuse or slow down users.

The following image shows a visible box when a user places keyboard focus on the box. Without that box, it appears like the user can't interact with this widget:

image

## Steps to Reproduce
1. Navigate to the Reader Interface
2. Navigate to a project with readable and searchable text
3. Navigate into a text contained in that project
4. Navigate to and activate the search button in the menu
5. Enter any search criteria into the edit field and submit the search
6. Navigate to the "Search Within" and "Show Results For" sections. Notice that these are implemented as checkboxes, but only one item can be selected, it does not visually appear interactable, and filters are applied immediately.

## Testing Environment
NVDA 2020.1
ESR Firefox 68.8.0 esr

dananjohnson:
This commit addresses a couple issues with the controls used to define search scope and facets in the context of the frontend and reader:

First, it replaces the checkboxes for search scope with radio controls, since only one scope can be selected at a time.

Second, it creates a new mixed-state checkbox to replace the existing checkboxes that control search facets. Now the "Everything" box has been replaced with a select all/none box that controls the other boxes, following the recommended WAI-ARIA design pattern here: https://www.w3.org/TR/wai-aria-practices-1.1/examples/checkbox/checkbox-2/checkbox-2.html

Resolves #2759

Screen Shot 2020-12-09 at 4 35 40 PM
Screen Shot 2020-12-09 at 4 35 31 PM


zdavis:
Why are we stacking these checkboxes vertically and adding the nesting? To my mind, that adds visual complexity that actually isn't necessary. This will also significantly increase the height of this component. Let's discuss.

zdavis:
I'd also like to see how this affects the full search, which has a few more facets. Can we follow the pattern from a markup perspective but keep a more horizontal layout?

dananjohnson:
@zdavis my PR's been updated. New design (same as the original design):
Screen Shot 2020-12-10 at 10 42 07 AM

zdavis:
Thanks, Dana!

Issue #2918, opened 2020-12-29
Formatted attributes should be backed by the database


scryptmouse:
Presently, they are backed by `Rails.cache` / Redis, which creates N+1 issues when fetching and serializing a large number of models.

PR #2930, opened 2021-01-27
[WIP] OER Improvements

### LINKED ISSUE #2929: Build APIs to support OER work
Current plan for this OER push, more or less in the order it will get tackled:

- [x] ReadingGroupCategory: an optional organizational item that can be reordered and has a number of associations to join against entries (manifold types like text, project, resource, etc).
- [x] ReadingGroupEntry: a concern that is implemented by said manifold types, always belongs to a reading group, can be set up to belong to a category
- [x] Join models, e.g. ReadingGroupText, ReadingGroupProject, ReadingGroupResource, ReadingGroupResourceCollection: which implement the Entry concern and join against the associated type. Optionally belongs to a category.
- [x] /api/v1/reading_groups/:reading_group_id/relationships/categories to support creating a category for a reading group
- [ ] /api/v1/reading_groups/:reading_group_id/relationships/entries/:id endpoint (PUT/DELETE only) for an abstract entry type that accepts a model type, id, category_id, and position to add, reorder, and delete a specific entry from the reading group's collection. Designed to be used with DND strategies and keep the client oblivious as to the underlying structure. The idea is that the client tells the server where each entry is located (categorized or not) and should be, and the server makes sure it gets there.
- [x] The service to support the above endpoint's abstraction and know where to put entries, reorder if necessary, etc.
- [x] (Maybe) /api/v1/reading_groups/:reading_group_id/relationships/collection endpoint for an abstract collection type that accepts an array of the above entry types in an additive fashion: allowing for batch adding/updating of the reading group's collection of items.
  - It would be PATCH style as we don't want to overwrite resources, just add them if they are not already there, or update them if the provided tuple differs from the current status
  - I'm thinking specifically of the modal collection interface, being able to push a list to the specified endpoint and have it be updated in one request. In JSONAPI terms it would be a proxy type that doesn't exist in the database but does exist for serialization purposes to ease things on the client.
  - Another benefit of this approach is we also gain the ability to set the category in the modal UI, if we want to give that option at some point.
- [x] API endpoints for the actual entries, via has_many :through, e.g. /api/v1/reading_groups/:reading_group_id/relationships/texts—ordering and such happens on the server because it knows how to order those types for that reading group
- [x] Build migration to modify RG memberships to support being archived (and de-archived for whoopsies): current thought is an archived_at timestamp, updating current has_many :reading_group_memberships to exclude anything archived, as well as exclude annotations from archived reading groups, add new archived_reading_groups association. keeps things simple and easy to undo when (not if) a user accidentally archives themselves from the wrong reading group.
- [x] Build services to archive a reading group membership, duplicate reading groups, and clone RG annotations
- [x] Update authorizers with logic for all of the above
- [x] Enum for annotation style
- [x] Migration to allow labels and notes in reading group memberships, update serializers
- [ ] Add ReadingGroupType model (TODO: Find out sane defaults from manipeeps?)
- [ ] Add course? boolean and course_start_date, course_end_date to ReadingGroup
- [ ] Add controller endpoint for reading/modifying/creating RG types, authorizer logic
- [x] Tighten up Favorite model
  - add NOT NULL to columns where it should be
  - maybe add actual inferred project relationship based on favoritable rather than introspecting within the model at runtime–this could allow something like User#favorite_projects to refer to any project a user had collected something in rather than just directly collected projects
  - alt Q: should we refactor favorites to remove polymorphism so it has parity with collection types (minus categorization)? We could recreate the current polymorphic favorites table with a view but otherwise allow us to use similar logic / structure for collections.
- [ ] Update filtering for rg/relationships/annotations controller to support comps
- [ ] Update API to support listing public reading groups—possibly separate /api/v1/public_reading_groups endpoint rather than overloading the current RG controller? Also makes the "join" action for a public group a very simple affair: POST /api/v1/public_reading_groups/:reading_group_id/relationships/memberships inferring the current user and joining them right there rather than having to use the invite logic within the API itself

scryptmouse:
Resolves #2929

Requires #2919 (issue #2918)


zdavis:
Closing, because this work is all in `feature/oer-improvements` now.

PR #2997, opened 2021-05-03
[F] Add reading group filter to notes drawer

jen-castiron:
Resolves PROJ-133


jen-castiron:
Looks like my branch got out of sync with Dana's - closing to refactor

jen-castiron:
Hey @dananjohnson & @zdavis ! I haven't hooked this up to the API yet, but it's at the point where I think you should take a look. Let me know if there's anything I can do to make it more "manifold" - even if it's nit picky.

client/src/reader/containers/ReaderNotes/index.js
```diff
@@ -131,6 +131,15 @@ export class ReaderNotesContainer extends Component {
     );
   };
 
+  handleReadingGroupChange = rgId => {
```
dananjohnson, line 134: You could probably just use the `handleFilterChange` method for this to simplify things. Would just take a slight change to the shape of the data passed back from `GroupFilter`.

dananjohnson (changes requested):
Don't kill me for saying this, but my feeling is that this should be a simple `<select>`. (Sorry, I should have taken a closer look at the mockup you shared earlier in your work.) This would be more consistent with how filtering is done elsewhere in Manifold. And a link to manage reading groups is not far away already if we lose the button in the popup. What do you think, @zdavis? I do see that you've addressed a11y with the hidden select, so that's not really a concern.

Screen Shot 2021-05-03 at 10 30 09 AM

A couple other changes:
- [ ] The responsive layout needs an update since there's more there in the filter bar now.
Screen Shot 2021-05-03 at 10 14 30 AM
- [ ] can you update the See All button styling too? I don't think we have an existing utility style for this type of button, and I don't think it exists elsewhere, so I'd probably drop the `.button-primary` and just make the changes on `.notes-filtered-list__see-all`.
- [ ] 
Screen Shot 2021-05-03 at 10 34 31 AM

PR #3007, opened 2021-05-12
[F] Add reading group to notes drawer

jen-castiron:
- Resolves PROJ-133

TODO:
- [x] add RG filter
- [x] show all of group's notes for current text
- [x] show member name & comment count next to note
- [x] fix drawer filters responsive styling


client/src/reader/components/notes/partial/GroupFilter.js
```diff
@@ -0,0 +1,71 @@
+import React, { PureComponent } from "react";
```
dananjohnson, line 1: I noticed I was getting an empty array for `props.readingGroups` at first, but changing this from `PureComponent` to `Component` seemed to fix it. Presumably `PureComponent` was causing it not to update when `withReadingGroups` updated that empty array with group objects.

dananjohnson (changes requested):
Looks good to me other than the one change!

client/src/theme/styles/components/global/annotation/_tag.scss
```diff
@@ -0,0 +1,45 @@
+.annotation-tag {
+  display: flex;
+  align-items: center;
+  overflow: hidden;
+  background-color: $neutral05;
```
dananjohnson, line 5: use `var(--box-bg-color)` here

client/src/theme/styles/components/global/annotation/_tag.scss
```diff
@@ -0,0 +1,45 @@
...
+    & + & {
+      padding-left: 8px;
+    }
+
+    > * + * {
+      margin-left: 4px;
+    }
+
+    &--dark {
+      background-color: $neutral10;
```
dananjohnson, line 28: Use `var(--box-medium-bg-color)` here

client/src/theme/styles/components/global/annotation/_tag.scss
```diff
@@ -0,0 +1,45 @@
...
+    }
+  }
+
+  &__text {
+    overflow: hidden;
+    font-size: 14px;
+    font-weight: medium;
+    color: $neutral75;
```
dananjohnson, line 36: Create a var for this in `_appearance.scss` to make this usable on the backend potentially.

client/src/theme/styles/components/global/annotation/_tag.scss
```diff
@@ -0,0 +1,45 @@
...
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
+
+  &__icon {
+    flex-shrink: 0;
+    color: $neutral75;
```
dananjohnson, line 43: same color here, or just use `currentColor`.

dananjohnson (changes requested):
Looking good! A few comments below, plus two other things I noticed:
* the vertical alignment looks a bit off
Screen Shot 2021-05-12 at 2 15 05 PM

- the text should truncate so that the tag doesn't get too wide. There's a `@mixin textTruncate` you can use for this.
Screen Shot 2021-05-12 at 2 15 10 PM

PR #3054, opened 2021-08-12
[B] Allow creators to be optional

zdavis:


zdavis:
This buster need some testing before we can call it good to go, but I think it's close.

zdavis:
Closing this, as these changes are included in #3688

Issue #2969, opened 2021-04-09
Improve tab stops/focus in reader and global search results.


tsangaaron09:
## Current Behavior ##
When navigating throughout Manifold, for example, when viewing search results, with keyboard navigation, there are seemingly empty tab stops during which you lose focus. From using a screenreader, we realised that the empty tab stop happens because the focus indicator will focus on the icon beside the title of the result before returning to the next navigable element. We have also found similar behaviour in one project. 

Project Link: https://uw.manifoldapp.org/projects/the-mill-on-the-floss
![Screenshot 2021-04-08 at 14 46 06](https://user-images.githubusercontent.com/29372531/114100464-2b4f1580-9879-11eb-9c4d-bbefe8308d0b.png)

## Expected Behavior ##
When navigating through elements like the search results, the icon and the title of the search result will activate the same action. Because of this, there should only be one tab stop. 

## Steps to Reproduce ##
1. Go to the search bar, either the global search or within a project
2. Type in a keyword that yields multiple results
3. Traverse the search results with Tab/Shift+Tab

If you have any questions feel free to ask.

zdavis:
I'm going to rewrite the issue title so that it's specific and actionable. We need issues to describe specific problems that we can address in specific spots. Thanks!

zdavis:
Resolved by #3080

PR #3080, opened 2021-09-09
[B] Add tabIndex param to wrapWithLink()

1aurend:
Added `tabIndex` prop to `Link` components created with `wrapWithLink` and set the tab index to -1 on figures to solve improve tabs in search results.

Resolves: #2969


dananjohnson (approved):
Looks great, @1aurend!

Issue #2757, opened 2020-05-27
Reader Interface Search Results Page Has Broken Heading Hierarchy


Santoz2:
Manifold's Reader Interface Search Results present a broken heading hierarchy. "Search Results" should be the main heading of the page, and other headings should be a sub-heading of that heading.

## Expected Behavior
A page's heading structure should have a hierarchy; headings should start at heading level 1, and subcontent should have consecutively higher headings, with categories of the same level of prominence having the same level heading. 

## Current Behavior
The Search Results page starts with "Search Results" as a level 3 heading. This was likely done as level 1 and level 2 headings may be found in the text underneath the search overlay. However, as Issue #2754 mentions, the content of the text underneath the search overlay should be hidden. The level 2 heading is also not always present, so there are instances where heading level 2 could be directly skipped if the content was included. Notably, skipping heading levels could imply to a non-visual user that they missed content somewhere on the page.

Further, the actual search results are also level 3 headings, but they are content that belong under the "Search Results" heading, so they should be lower in the hierarchy.

Heading hierarchy can often be determined by logic tests. In this instance, the actual search results should be under the "Search Results" headings. For the more complex issue of the main page heading, consider what content is being presented. The original heading level 1 is the section that the reader was in. However, the search results can include content that belongs to the rest of the chapter, or the rest of the project, so the search content does not necessarily belong in the original section, and therefore should have a higher level heading because it is not a subset of that content. 

The following image is a visual representation of the search page hierarchy, made visible by the Accessibility Bookmarklets tool found here: [https://accessibility-bookmarklets.org/]( https://accessibility-bookmarklets.org/)

![image](https://user-images.githubusercontent.com/43553368/83061627-2ba04280-a012-11ea-85a3-91eedf43d345.png)

## Steps to Reproduce
1. Navigate to the Reader Interface
2. Navigate to a project with readable and searchable text
3. Navigate into a text contained in that project
4. Navigate to and activate the search button in the menu
5. Enter any search criteria into the edit field and submit the search
6. By using a tool such as a Screen Reader's Rotor, or accessibility bookmarklets, inspect the Headings on the page. Note that the heading from the underlying page is still present, and the highest level heading on the actually visible page is a heading level 3.

## Testing Environment
NVDA 2020.1
ESR Firefox 68.8.0 esr

zdavis:
Resolved in #3081

PR #3081, opened 2021-09-09
[B] Fix heading level in Overlay

1aurend:
Change global overlay title from `h3` to `h2`.

Resolves: #2757


1aurend:
@dananjohnson I think this is ready now.

dananjohnson (approved):
approved!

PR #3084, opened 2021-09-13
Update Collapse animation and move styles to css

1aurend:


dananjohnson (changes requested):
Looks really good, @1aurend! The one thing I might change is to make the custom property names a bit more specific by prepending them with `collapse` (e.g. `--collapse-height`). I think that will make it just a bit easier for other devs to see what these vars are for, and will be especially useful if at some point we want to redefine the var values in the scope of another component!

1aurend:
@dananjohnson Let me know what you think of these names and if it's good to merge and start implementing. Thanks!

dananjohnson (approved):
Names look good! Go ahead and merge.

PR #3086, opened 2021-09-14
 [F] Support keyboard annotation

dananjohnson:
This PR adds support for annotating texts via keyboard in the Manifold reader. There were two aspects to this work:

The first was adding the appropriate attributes to previously highlighted or annotated text nodes. Both now have a `role` of `button`, `tabindex` of `0`, and an `aria-label`. They also received values of `aria-haspopup` appropriate to the expected interaction, with annotations getting `dialog` and highlights getting `menu`.

The second aspect was adding support for creating new annotations via keyboard. This work is scaffolded upon existing browser and OS functionality for keyboard text selection; @zdavis laid out on the relevant [GH issue](https://github.com/ManifoldScholar/manifold/issues/2326#issuecomment-523208990) why this is the responsibility of the browser and not Manifold. Our job was to ensure that text selection coming via keyboard triggered the annotation popup menu, which previously only responded to mouse and touch events. In addition, we needed to refactor the popup menu to be useable via keyboard and screen reader software. We've used the [WAI-ARIA menu design pattern](https://www.w3.org/TR/wai-aria-practices/#menu), with appropriate roles, focus management, and keyboard navigation via arrow keys.

To use this feature, users need to first enable caret browsing by pushing F7 in Firefox, Chrome, or Edge, or by [enabling Mouse Keys](https://support.apple.com/guide/mac-help/control-the-pointer-using-mouse-keys-mh27469/mac) in MacOS system accessibility preferences for use in Safari.


zdavis:
@dananjohnson — could you rebase this on master when you get a chance? This PR needs to get into v7, and it's starting to drift pretty far from master. Thanks!

zdavis:
Rebased on master

zdavis:
Posting Dana's note from Slack in here:

> So here’s where things are at with keyboard annotation, and where I need help:
> 
> There have been two aspects to this work. This first was to make existing highlights and annotations reachable by tabbing. That was pretty simple; just a matter of adding the right attributes to highlighted and annotated text nodes. The second was to trigger the annotation popup when text is selected via caret browsing. (This led to also needing to refactor the popup so it too could be interacted with via keyboard, and have meaningful markup and ARIA attributes, etc.)
> Once the popup refactor is done, we should have full support for keyboard annotation in Chromium-based browsers. But there’s one limitation with Safari and Firefox that I’m trying to solve. The issue is that unlike Chrome, Safari and Firefox treat focus and text selection interdependently. So as focus moves inside the popup when it opens, as is required, the selection is lost. Note that I’m talking only about what the browser considers to be the active selection. This is (fortunately!) different than what the Annotatable component has stored as the active selection. In other words, the issue is that Manifold knows what the active selection is, but the user can’t see it anymore because the yellow highlighting disappears.
> 
> The best solution I could think of was to do something similar to when a selection is locked (and the annotation drawer opens). In that case, the selection is wrapped in a <span>, and some CSS applies the same yellow color as ::selection. (It looks a little different than ::selection , but that’s a separate issue, and solvable I think.)
> As a POC, I changed the logic for locking a selection to also do it when a selection is complete and pending action (i.e. when the popup appears). This has the effect we want, but it causes some issues with the wrong text being selected and the popup being wrongly positioned. So I think we’ll need to build separate logic to add HTML to pending selections vs locked ones. That’s where I’m hoping you can come in. There’s a lot happening in this functionality and I’m not sure I can address this confidently. But if we can solve it, I think we’ll be at full support for keyboard annotation in all modern browsers!

PR #3099, opened 2021-10-06
New components for Journals

1aurend:
This PR creates new versions of some existing components, which, while intended for use with journals, are generic and can be implemented for all entities eventually. New components include:
- EntityThumbnail
- EntitiesBox
- ThumbnailGrid

Resolves:
- PROJ-736


1aurend:
Also, should I be adding tests for these new components?

PR #3116, opened 2021-11-08
[C] Migrate new journal components to Linaria

1aurend:


client/src/frontend/containers/Test/index.js
```diff
@@ -0,0 +1,216 @@
+import React, { Component } from "react";
```
1aurend, line 1: Will delete this file/route before merging!

PR #3117, opened 2021-11-10
[F] Build EntityGroup component

1aurend:
Creates `EntityGroup` component for Journals. Also, includes Linaria migration and styling tweaks to `EntityGrid`, `EntityThumbnail`, and `EntityBox` for use in the new group component.

New components are on `/test` route.


client/src/global/components/atomic/EntityBox/styles.js
```diff
@@ -0,0 +1,13 @@
+import { styled } from "@linaria/react";
+import { panelRounded } from "theme/styles/mixins";
+
+export const Box = styled.div`
+  --entity-box-padding-vertical: 20px;
+  --entity-box-padding-left: 45px;
+  --entity-box-padding-right: 30px;
```
1aurend, line 7: Made these into vars for styling the `EntityGroup` header.

client/src/global/components/composed/ThumbnailGrid/styles.js
```diff
@@ -26,20 +26,28 @@ export const Grid = styled.div`
...
+  ${({ $grid, $empty }) => {
+    /* eslint-disable no-nested-ternary */
+    const finalStyles =
+      $grid && $empty
+        ? gridStyles + `margin: auto;`
+        : $grid
+        ? gridStyles
+        : $empty
+        ? emptyStyles
+        : null;
```
dananjohnson, line 47: I find this kind of thing really hard to read. (eslint doesn't like it apparently either. 🙂) Is there another way you could write it? (Not saying the way I did is right.  The logic seems more if/elseif than mine.)

dananjohnson (changes requested):
Looking good, @1aurend. Let me know if you have any questions about my feedback. One more thing to note is that it looks like EntityBox has more inline padding in the comps.

Here's design:
Screen Shot 2021-11-12 at 11 13 28 AM

And here's the implementation:
Screen Shot 2021-11-12 at 11 12 22 AM

client/src/global/components/composed/EntityGroup/index.js
```diff
@@ -3,35 +3,41 @@ import PropTypes from "prop-types";
...
-        </ThumbnailGrid>
-      </EntityBox>
-    </Styled.Wrapper>
+    <EntityBox
+      style={{
+        "--EntityBox-Container-padding-block": "30px",
+        "--EntityBox-Background-padding-block": "20px",
+        "--EntityBox-Background-padding-inline": "60px"
+      }}
+    >
```
dananjohnson, line 17: Using `style` should be a last resort, meaning we shouldn't build it into `EntityGroup`. The reason is that styles inlined in HTML have highest specificity, so couldn't be overridden by any other styled components (except by using `!important`). If you want to pass CSS var values to EntityBox, I think you should expose `className` as a prop on EntityBox and use it as a styled component in EntityGroup.

client/src/global/components/composed/EntityGroup/index.js
```diff
@@ -3,35 +3,41 @@ import PropTypes from "prop-types";
...
-            ))
-          }
-        </ThumbnailGrid>
-      </EntityBox>
-    </Styled.Wrapper>
+    <EntityBox
+      style={{
+        "--EntityBox-Container-padding-block": "30px",
+        "--EntityBox-Background-padding-block": "20px",
+        "--EntityBox-Background-padding-inline": "60px"
```
dananjohnson, line 15: Are the last two needed? They have the same values as EntityBox.

client/src/global/components/atomic/EntityBox/styles.js
```diff
@@ -1,9 +1,25 @@
...
+  padding-block: var(--EntityBox-Container-padding-block, 0)}
+`;
+
+export const Background = styled.div`
   ${panelRounded}
-  padding-block: var(--entity-box-padding-vertical, 20px);
-  padding-inline-start: var(--entity-box-padding-left, 30px);
-  padding-inline-end: var(--entity-box-padding-right, 30px);
+  padding-block: var(--EntityBox-Background-padding-block, 20px);
+  padding-inline: var(--EntityBox-Background-padding-inline, 60px);
```
dananjohnson, line 24: A couple changes with this:
* I measured this value in the comps as 72px, not 60px.
* On mobile, it's too wide. Have it match the inline padding of the project grid above it, scaling between desktop and mobile with `fluidScale`.
Screen Shot 2021-11-15 at 10 12 54 AM

client/src/global/components/composed/EntityGroup/styles.js
```diff
@@ -57,9 +48,10 @@ export const GroupHeader = styled.div`
     `}
 `;
 
-export const HeaderText = styled.span`
+export const HeaderText = styled.h2`
   ${headingPrimary}
+  margin-inline-start: -15px;
   margin-bottom: 0;
-  color: var(--color-neutral-text-extra-dark);
+  color: var(--strong-color);
```
dananjohnson, line 55: Not essential, but if you define this on `GroupHeader` you only have to do it once. The icon and HeaderText will just inherit it.

client/src/frontend/containers/Test/index.js
```diff
@@ -146,68 +146,60 @@ export class ProjectsContainer extends Component {
...
+        <div>
+          <EntityGroup
+            title="Volume 2"
+            entities={this.props.projects.slice(2, 6)}
+          />
+          <EntityGroup
+            title="Volume 3"
+            entities={this.props.projects.slice(6, 10)}
+            to="/home"
+          />
```
dananjohnson, line 200: This component has a really nice, simple set of props!

client/src/global/components/composed/EntityGroup/styles.js
```diff
@@ -26,14 +18,11 @@ export const Icon = styled(IconComposer)`
...
-  padding-inline-start: var(--entity-box-padding-right);
-  padding-inline-end: var(--entity-box-padding-left);
+  padding-inline: var(--EntityBox-Background-padding-inline);
 
   /* Ensure background hover color goes to edge of EntityBox. */
-  margin-block: calc(-1 * var(--entity-box-padding-vertical));
-  margin-inline-start: calc(-1 * var(--entity-box-padding-left));
+  margin-block: calc(-1 * var(--EntityBox-Background-padding-block));
+  margin-inline-start: calc(-1 * var(--EntityBox-Background-padding-inline));
   box-sizing: content-box;
```
dananjohnson: Nice solution here!

dananjohnson (changes requested):
Great work, @1aurend! I think the responsive styles just need a bit of attention. See my comments below.

client/src/global/components/atomic/EntityBox/styles.js
```diff
@@ -15,11 +15,14 @@ export const Container = styled.section`
     breakpoints[120],
     breakpoints[120]
   )};
-  padding-block: var(--EntityBox-Container-padding-block, 0)}
+  padding-block: var(--EntityBox-Container-padding-block, 0);
 `;
 
 export const Background = styled.div`
+  --EntityBox-Background-padding-block: 20px;
+  --EntityBox-Background-padding-inline: ${fluidScale("72px", "20px")};
```
dananjohnson, line 23: Prob simpler to add these values as defaults when you use the vars below rather than define them here.

dananjohnson (approved):
Looks good! 👏🏻  Go ahead and merge.

PR #3118, opened 2021-11-10
[C] Switch iconClass to className for use with styled

1aurend:


dananjohnson:
Looks good! I'm gonna go ahead and merge this so I can make use of it too!

PR #3134, opened 2021-12-17
[F] Integrate Breadcrumbs

1aurend:
Incorporates prior `BackLink` components into a new `Breadcrumbs` component that can handle one or multiple links. Also, removes secondary styling and incorporates backend styling.


client/src/global/components/atomic/Breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,51 @@
...
+  breadcrumbs,
+  backend = false,
+  hideOnDesktop = false
+}) {
+  const Wrapper = backend ? Styled.BackendOuter : Styled.Outer;
+  const Inner = backend ? Styled.BackendInner : Styled.Inner;
+  const Icon = backend ? Styled.Icon : IconComposer;
+
+  return (
+    <Wrapper aria-label="Breadcrumb" $hideOnDesktop={hideOnDesktop}>
```
dananjohnson, line 16: Set the `aria-label` to `Breadcrumbs`.

client/src/frontend/containers/ProjectCollections/index.js
```diff
@@ -145,9 +145,10 @@ export class ProjectsCollectionsContainer extends Component {
     return (
       <>
         <CheckFrontendMode debugLabel="ProjectCollections" isProjectSubpage />
-        <Utility.BackLinkPrimary
-          link={lh.link("frontendProjectsAll")}
-          backText={"Back to projects"}
+        <Breadcrumbs
+          breadcrumbs={[
+            { to: lh.link("frontendProjectsAll"), label: "Back to projects" }
+          ]}
```
dananjohnson, line 151: Is there a reason why the component is being rendered directly rather than using the breadcrumb context?

client/src/global/components/atomic/Breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,51 @@
...
+          breadcrumbs.map(crumb => (
+            <Styled.Breadcrumb to={crumb.to} key={crumb.to}>
+              <Icon icon="arrowLeft16" size={backend ? "default" : 24} />
+              {backend && <Icon icon="arrowLeft32" size="default" $desktop />}
+              <Styled.Label>{crumb.label}</Styled.Label>
+            </Styled.Breadcrumb>
+          ))}
+        {breadcrumbs.length > 1 && (
+          <>
+            <Icon icon="arrowLeft16" size={24} />
```
dananjohnson, line 28: I think people are going to be confused if this icon is unclickable. Let's make it part of the first link.

dananjohnson (changes requested):
Really good stuff, @1aurend! This is basically ready to be merged. I'm mainly wondering about why the context is used in, for instance, the resource collection container, but in other places the Breadcrumbs component is used directly. If possible, I think it'd be preferable to use the context in all cases. But maybe there's a reason we can't?

PR #3144, opened 2022-01-11
Initial l10n implementation

dananjohnson:


dananjohnson:
@zdavis this should be ready to merge. I've pulled out any actual translations and just left the framework in place, along with a couple other commits that should get into master.

PR #3145, opened 2022-01-11
[F] Add common localization namespace

dananjohnson:


dananjohnson:
@LeonorPDX the initial localization work has been merged into `master`, and I've moved your work-in-progress to this branch, `feature/l10n`. You should push the rest of your work here instead of `am/build-l10n`.

PR #3149, opened 2022-01-13
[F] Add Issue routes and containers

1aurend:


dananjohnson:
@1aurend want to take a quick peek and make sure it works for you? There's not much to see on the issue detail page yet; I've composed it just like a project detail page and am assuming it'll Just Work™️ if issues are basically projects.

1aurend:
@dananjohnson what I can see looks good, and the routes are working for me. There's some weirdness in where the breadcrumbs direct you, but I think that's an artifact of our mock data. 

One small thing I noticed is that we haven't been totally consistent about whether the See All button renders at the bottom of these detail containers, but we can probably tackle that when we have the api.

dananjohnson:
Thanks for looking it over, @1aurend! Yeah, we should take a pass at all the frontend containers when we're done and make sure they're all rendering breadcrumbs and "See all" links in a consistent manner.

PR #3150, opened 2022-01-15
[F] Build out journals UI

dananjohnson:
This adds UI for the JournalsList, JournalDetail, and VolumeDetail containers. These will need to be updated with actual data when the API exists, but the UI is in place.


dananjohnson:
@1aurend requesting a quick review to make sure I didn't do anything stupid and that everything matches the comps. Thanks!

client/src/frontend/components/composed/EntityCollection/patterns/JournalIssues.js
```diff
@@ -0,0 +1,85 @@
...
+        )
+      }
+      FooterComponent={() =>
+        totalIssueCount > limit && (
+          <FooterLink
+            to={lh.link("frontendJournals", slug)}
+            label="See all issues"
+          />
+        )
+      }
```
1aurend, line 69: In design, this sometimes has a "Visit the journal page" link rather than "See all issues". Maybe we could render the visit link if `totalIssueCount` is less than `limit`?

Screen Shot 2022-01-18 at 11 10 39 AM

client/src/frontend/components/journal/index.js
```diff
@@ -0,0 +1,9 @@
+import IssueList from "./IssueList";
```
1aurend, line 1: Just a question about how we're organizing components: do these not live under `composed` because they're at the level of a page layout?

client/src/frontend/components/journal/Metadata/index.js
```diff
@@ -0,0 +1,27 @@
...
+          renderable: true,
+          type: "Content::MetadataBlock",
+          title: "Journal info",
+          icon: "journals64"
+        }
+      }}
+      entity={journal}
+    />
+  );
+}
```
1aurend, line 19: I think the white space between this and volume groups is taller in design?

Screen Shot 2022-01-18 at 11 16 17 AM

1aurend: Same below the metadata block. It looks to me like there's more white space in design.

client/src/frontend/components/journal/VolumeDetail/index.js
```diff
@@ -0,0 +1,95 @@
...
+    <Styled.Wrapper>
+      <EntityGroup title={`${journalTitle}: ${volumeTitle}`}>
+        {issues.map(issue => (
+          <Styled.IssueWrapper>
+            <ThumbnailGrid key={issue.id} minColumns={4} minItemWidth="210px">
+              {({ stack }) => (
+                <EntityThumbnail entity={issue} stack={stack} key={issue.id} />
+              )}
+            </ThumbnailGrid>
+            <ContentBlockList entity={issue} nested hideHeader />
```
1aurend, line 80: Should I be able to see these edit buttons when not logged in? They don't appear at all in design. The icons also aren't in design for volume detail, just issue, but I can see if this was a choice for consistency.

Screen Shot 2022-01-18 at 11 28 14 AM

client/src/frontend/components/journal/IssueList/CountTemplate.js
```diff
@@ -0,0 +1,29 @@
...
+import PropTypes from "prop-types";
+import pluralize from "pluralize";
+
+function CountTemplate({ count, unit, categoryCount }) {
+  if (categoryCount)
+    return (
+      <>
+        There are <strong>{count}</strong> {unit}
+        {" in "}
+        <strong>{categoryCount}</strong> {pluralize("volume", categoryCount)}
```
1aurend, line 11: Just making a note that we should handle `pluralize` when we do localization.

client/src/frontend/components/journal/VolumeDetail/styles.js
```diff
@@ -0,0 +1,25 @@
...
+
+export const Wrapper = styled.div`
+  padding-block-start: var(--container-padding-block-start);
+  padding-block-end: var(--container-padding-block-end);
+`;
+
+export const IssueWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 5px;
```
1aurend, line 12: Fwiw, we ran into issues using `gap` with `flex` on NGLP in Safari <= 14.0, and ended up deciding not to use it for now.

client/src/hooks/journals/use-select-volume/index.js
```diff
@@ -2,11 +2,11 @@ import { useSelector } from "react";
...
-        item => item.attributes.slug === slug
-      )[0]
-    : null;
-  return { volume };
+export default function useSelectVolume(match, journal) {
+  if (!journal) return { volume: null, volumeResponse: null };
+
+  const volume = journal.relationships.volumes.find(
+    v => v.id === match.params.volumeSlug
+  );
```
1aurend, line 10: Thank you for this! How did I not know about `.find()`??

1aurend (approved):
@dananjohnson these look great! Like I said, it's neat to see it all come together. I added a few small notes and questions. And, of course, there are elements I can't see because we don't have real data. It seems like we're just missing `IssueDetail` UI now?

client/src/frontend/components/composed/EntityCollection/patterns/JournalIssues.js
dananjohnson: Good eye! Yeah, that must be the logic for when those different labels appear.

client/src/frontend/components/journal/index.js
dananjohnson: Since these are specific to journals, I think it makes sense to keep them in a `/journals` dir. I think `/composed` should be limited to model-agnostic "Entity" components so we don't end up just putting every component we build there. How's that sound?

client/src/frontend/components/journal/VolumeDetail/index.js
dananjohnson: These aren't edit buttons; they're supposed to show highlight and annotation counts, but the test fixture I used appears to have incomplete attributes. In any case, issues will have the same content blocks as projects, so I'm just rendering those, without making any changes to design.

client/src/hooks/journals/use-select-volume/index.js
dananjohnson: 🙂 

It's hard to keep track of all the Array methods in JS. I have to go looking for them all the time.

client/src/frontend/components/journal/VolumeDetail/styles.js
dananjohnson: Boo Safari 👎🏻. Since it's fixed in 14.5, I'm inclined to keep it rather than revert to the old negative margin hack. That's two major iOS versions we'd be supporting. Happy to discuss with the larger group though. It raises a larger question of how the team should handle Safari (and all iOS browsers), since it's very frustratingly tethered to the OS version. If we want to support older releases, we'll want to take a closer look at what we can/can't use in CSS and JS. I know, for instance, that logical properties have only partial support in Safari 13.

client/src/frontend/components/journal/Metadata/index.js
dananjohnson: Yeah true. In these cases I'm just using the block padding defined on `.container`. Even though it departs from design a bit, I'm inclined to use global values like this rather than to start defining specific spacing for a particular detail view. Maybe where we should be heading is creating something like a `EntityDetail` component that lays out page elements with consistent spacing. Abstract it at that level, rather than to try to define it consistently on each route.

PR #3154, opened 2022-01-19
[B] OER bug fixes

1aurend:
Resolves:
- PROJ-745
- PROJ-746
- PROJ-750
- PROJ-698


client/src/frontend/components/collecting/reading-group/edit/CollectionEditor/SortableCollectables/Collectable.js
```diff
@@ -44,24 +43,46 @@ function Collectable({ responses, type, id, index, onRemove, onMove }) {
...
-                  onBlur={() => setKeyboardActionFocused(false)}
-                  direction="up"
-                />
-                <Move
-                  onClick={() => onMove({ id, type, direction: "down" })}
-                  onFocus={() => setKeyboardActionFocused(true)}
-                  onBlur={() => setKeyboardActionFocused(false)}
-                  direction="down"
+              <div
+                role="none"
```
dananjohnson, line 35: This isn't really necessary, since divs have no inherent role. Unless I'm missing something with the intent here?

dananjohnson (commented):
This looks good, @1aurend! I tried to dig into the RG tabbing stuff a bit more, and what I worked out was that with my earlier solution, the focus goes to the `body` element after you tab away from the Drag component. No idea what's causing this. But your solution works, so let's go with that. I noted a couple things here, and I noticed one unrelated issue that I'm gonna push a fix for, but otherwise this is ready to go. Thanks for your work on it!

dananjohnson:
To clarify, once you've made those changes, please mark this as ready for review and flag Zach as a reviewer.

PR #3160, opened 2022-01-28
[B] Regression in EntityHero

1aurend:
Resolves:

- PROJ-804
- PROJ-805


1aurend:
@dananjohnson I think I got these two, but would you mind taking a look? I had to add `transientOptions` to `Styled.LinkCallout` to handle the `as` prop.

zdavis:
Rebased on master

PR #3168, opened 2022-02-03
[F] Refactor journals containers and layouts

1aurend:
Changed to draft to save my comments for reference. Most of these commits were merged with `feature/journals`.


client/src/frontend/components/atomic/EntityHeadContent/index.js
```diff
@@ -25,10 +30,14 @@ function EntityHeadContent({ entity, parentEntity = { attributes: {} } }) {
...
-      : titlePlaintext;
-    return `\u201c${title}\u201d on ${installationName}`;
+    const titleOrNum = (() => {
+      if (parentTitle) {
+        if (number) return `${parentTitle}: ${type} ${number}`;
+        return `${parentTitle}: ${titlePlaintext}`;
+      }
+      return titlePlaintext;
+    })();
+    return `\u201c${titleOrNum}\u201d on ${installationName}`;
```
1aurend, line 40: Since volumes and issues have titles not numbers, I've ended up doing a version of this in several places. Is there a better solution?

client/src/frontend/components/atomic/EntityHeadContent/index.js
```diff
@@ -57,6 +66,7 @@ function EntityHeadContent({ entity, parentEntity = { attributes: {} } }) {
 
 EntityHeadContent.propTypes = {
   entity: PropTypes.object.isRequired,
+  type: PropTypes.string,
```
1aurend, line 69: We could use `entity.type` but we'd have to format the strings, so I decided to hard code the prop. Could be a bad decision.

client/src/frontend/components/composed/EntityHero/parts/CalloutList/Callout/params.js
```diff
@@ -6,7 +6,21 @@ export const getCalloutParams = (data, type, isLink) => {
...
+      case "Website":
+        return "WEBSITE";
+      case "EMAIL":
+        return "EMAIL";
+      case "SHARE":
+        return "SHARE";
+      default:
+        return "LINK";
+    }
+  })();
```
1aurend, line 21: Not sure what the best solution is here. All of these are `type: LINK`, but they have unique icons in design. These strings certainly aren't going to catch all cases.

client/src/frontend/components/composed/EntityMasthead/index.js
```diff
@@ -44,7 +44,7 @@ export default function Masthead({ entity }) {
           alt=""
         />
       )}
-      {logo && (
+      {!image && logo && (
```
1aurend, line 47: Larger open question here: how do we want to handle the masthead? Do we need to add a color selector and place for logo in the backend?

client/src/frontend/components/journal/IssueList/index.js
```diff
@@ -3,64 +3,45 @@ import PropTypes from "prop-types";
...
         <Utility.EntityCount
-          count={8}
+          count={journal.attributes.journalIssuesCount}
           unit="issue"
           customTemplate={(count, unit) => (
             <CountTemplate
               count={count}
               unit={unit}
-              categoryCount={categories.length}
+              categoryCount={journal.attributes.journalVolumesCount}
```
1aurend, line 30: For discussion: what's the logic here if there are issues in volumes and not. Are there two count labels, one for each section? Or an option to say something like "There are X issues _and_ Y volumes"

client/src/frontend/components/journal/VolumeDetail/index.js
```diff
@@ -4,80 +4,28 @@ import EntityGroup from "global/components/composed/EntityGroup";
...
-    };
-  });
-}
-
 function VolumeDetail({ journal, volume }) {
   const journalTitle = journal.attributes.title;
-  const volumeTitle = volume.attributes.title;
-  const issues = getIssues(journal) || [];
+  const volumeTitle = `Volume ${volume.attributes.number}`;
+  const issues = volume.relationships.journalIssues;
```
1aurend, line 12: We're still missing a couple of minor things from issues when grabbing them from `relationships`, including `recentlyUpdated` and `draft`. We could dig into the associated project to get them, but do we want those tags to show up on journal/volume detail pages?

client/src/frontend/components/journal/VolumeDetail/index.js
```diff
@@ -4,80 +4,28 @@ import EntityGroup from "global/components/composed/EntityGroup";
...
-            <ThumbnailGrid key={issue.id} minColumns={4} minItemWidth="210px">
+          <Styled.IssueWrapper key={issue.id}>
+            <ThumbnailGrid minColumns={4} minItemWidth="210px">
               {({ stack }) => (
                 <EntityThumbnail entity={issue} stack={stack} key={issue.id} />
               )}
             </ThumbnailGrid>
-            <ContentBlockList entity={issue} nested hideHeader />
+            <ContentBlockList
+              entity={issue.relationships.project}
```
1aurend, line 25: We do have the whole project available; I'm just not passing it to `EntityThumbnail`. Maybe we should?

client/src/frontend/components/atomic/EntityHeadContent/index.js
zdavis: I'm confused by this comment. Volumes and issues have numbers, not titles, right?

client/src/frontend/components/composed/EntityHero/parts/CalloutList/Callout/params.js
zdavis: Don't include icons for now. Have them behave as they do on existing project pages.

client/src/frontend/components/atomic/EntityHeadContent/index.js
1aurend: Oh oops. Yep! I mean what you said. To elaborate, I am wondering if checking for `number` in these entity components is the right way to determine whether to render number or title for a given entity.

PR #3184, opened 2022-02-09
[F] Show selected journals on homepage

1aurend:


client/src/frontend/containers/Home/Collections.js
```diff
@@ -12,17 +12,47 @@ export default function HomeCollectionsContainer() {
...
+        collections.map((projectCollection, i) => (
+          <EntityCollection.ProjectCollectionSummary
+            key={projectCollection.id}
+            projectCollection={projectCollection}
+            limit={projectCollection.attributes.homepageCount}
+            bgColor={(journalCount + i) % 2 === 0 ? "neutral05" : "white"}
+          />
+        ))}
+    </>
+  );
```
dananjohnson, line 57: I see this comment in the design:
> For now, publishers can choose to have their featured Journals appear at the top of the home page, or at he bottom (i.e. below the Projects)

@zdavis sounds like this needs to be added to the API somewhere. Or do we want to just fix the order for now?

client/src/hooks/api/useFetch/index.js
```diff
@@ -105,6 +110,28 @@ export default function useFetch({ request, afterFetch, options = {} }) {
...
+      triggerFetchData().then(
+        () => {
+          if (isFunction(afterFetch)) afterFetch();
+        },
+        () => {
+          // do nothing
+        }
+      );
+    }
+  }, [authentication]);
```
dananjohnson, line 132: I don't know that `useEffect` can detect changes within the `authentication` object. So you may need to use `JSON.stringify(authentication)` here. Or is the effect just supposed to run when `authentication` changes to and from `null`?

1aurend: That's a good question. I don't know exactly what it detects, but it did refetch as expected when I logged in and out a few times.

zdavis: Our redux store should never be mutating an object. I'd expect useEffect (or any memoized react hook) to do a simple equality comparison on authentication, rather than a deep comparison. Because our store never mutates an object, you can expect authentication to not be equal if any property has changed in any way. In short, what Lauren is doing here should be reliable.

zdavis: That said, I'm not totally on board with this change to useEffect. It introduces redundancy, and it seems like we should be able to piggy back on useEffect dependencies to get to the same end result.

If I were doing this, I think I'd take a different approach. First, I think I'd introduce a `dependencies` array argument for the hook (defaulting to []). I'd include the values in that array in the useEffect dependencies, which would allow the useFetch caller to make anything trigger a refetch. Then, I might include a convenience boolean argument `withAuthDependency` that includes authentication as a dependency for the useEffect call if it's true. Refetching based when auth state changes seems really common, so I'd be ok with a convenience property. Also, even though we don't mutate the auth state, I'd probably make the auth dependency more focused. For example, I think there's a boolean authentication.authenticated you could use. That combined with the authenticated user's ID would probably be better dependencies. Otherwise, anytime the user is updated, which can happen when, for example, they change their reader preferences, everything is going to refetch.

The main thing I don't like about the change in this PR is that lines 97-104 and 123-131 are duplicate code, and it seems like we'll have to add boolean props anytime we need to make the effect dependent on something, which feels unsustainable.

client/src/frontend/containers/Home/Collections.js
zdavis: For now, let's just put them at the top.

PR #3186, opened 2022-02-09
[F] Build out Volumes List route and view

1aurend:


client/src/frontend/components/composed/EntityCollection/patterns/JournalVolumes.js
```diff
@@ -0,0 +1,61 @@
...
+  if (!journal || !volumes) return null;
+
+  const { title } = journal.attributes ?? {};
+
+  return (
+    <EntityCollection
+      title={`${title}: Volumes`}
+      IconComponent={props => (
+        <ProjectCollectionIcon {...props} collection={journal} />
+      )}
```
1aurend, line 23: I didn't put a title link here because I couldn't think of anything that would make sense.

client/src/frontend/components/journal/IssueList/index.js
```diff
@@ -46,6 +51,15 @@ function JournalIssueList({ journal }) {
...
         )}
+        {/* Replace with showVolumesLink before merging */}
+        {true && (
+          <Styled.LinkWrapper>
+            <FooterLink
+              to={lh.link("frontendJournalAllVolumes", journal.id)}
+              label="See all volumes"
+            />
+          </Styled.LinkWrapper>
+        )}
```
1aurend, line 62: Just making it easy to get to the new route for review...

client/src/frontend/components/journal/IssueList/styles.js
```diff
@@ -13,3 +14,10 @@ export const List = styled.div`
   flex-direction: column;
   gap: 60px;
 `;
+
+export const LinkWrapper = styled(Box)`
+  > div {
+    background-color: var(--color-base-neutral-white);
+    padding: 0;
+  }
+`;
```
1aurend, line 23: This is bad but I couldn't come up with another way of getting the link in the right place, aside from refactoring this whole view into an `EntityCollection` pattern. Should we do that?

client/src/frontend/components/journal/VolumeList/index.js
```diff
@@ -0,0 +1,32 @@
...
+            title={`Volume ${volume.attributes.number}`}
+            to={lh.link("frontendVolumeDetail", journal.id, volume.id)}
+            entities={volume.relationships.journalIssues}
+            parentView
+          />
+        ))}
+      </Styled.List>
+    </Styled.Wrapper>
+  );
+}
```
1aurend, line 23: We can do some DRYing out of components by combining this with `Journal.IssuesList` if desired.

PR #3188, opened 2022-02-10
[B] Fix thumbnail presentation

1aurend:


zdavis:
This still isn't quite correct:

image

The style of the volume, issue line is incorrect. See design:

image

PR #3193, opened 2022-02-10
[B] Update button styles in project texts

1aurend:


1aurend:
@dananjohnson do you mind taking a quick look at this before I merge? I didn't want to refactor into styled here because this should be part of the larger button refactor we've talked about.

dananjohnson:
@1aurend changes looks good, expect that it looks like this iteration of `button-lozenge-secondary` needs to use `var(--highlight-color)`. Fix that and this should be good to go.

PR #3240, opened 2022-02-23
[C] Implement useFetch in RG + annotation containers

1aurend:
Also resolves:
- MNFLD-214
- MNFLD-280


client/src/frontend/containers/ReadingGroup/index.js
```diff
@@ -1,134 +1,70 @@
...
 import HeadContent from "global/components/HeadContent";
 import Drawer from "global/containers/drawer";
 import { RegisterBreadcrumbs } from "global/components/atomic/Breadcrumbs";
 import { GroupHeading } from "frontend/components/reading-group/headings";
 import Settings from "frontend/components/reading-group/Settings";
 import SearchDialog from "frontend/components/collecting/SearchDialog";
+import { useFetch, useFromStore } from "hooks";
 
-const { request } = entityStoreActions;
+export default function ReadingGroup({ route }) {
```
1aurend, line 15: There's definitely more refactoring that can be done to the child containers of this one. I tried to only do what was needed to replace the hooks, but I'd appreciate input on whether to keep going.

PR #3248, opened 2022-02-25
[C][F] Remove namespaces + Localize frontend/components/composed

1aurend:


1aurend:
@LeonorPDX I started with a set of newer components that I know well, so these all use the translation hook. Let me know if you have suggestions or questions!

Also, I forgot to include some exports of the `common.json` files in my previous PR, so we should merge this before you start a new l10n branch.

PR #3255, opened 2022-03-01
Begin reader localization

LeonorPDX:


LeonorPDX:
@1aurend I'm not familiar with the tests on Manifold, any tips to troubleshoot the failed test?

1aurend:
> @1aurend I'm not familiar with the tests on Manifold, any tips to troubleshoot the failed test?

The snapshots just need updating. If you run `yarn test -u` in the `client` directory and commit the new snapshots, tests should pass.

client/src/config/app/locale/en/en.json
```diff
@@ -163,7 +165,52 @@
...
+      "collection_other": "collections",
+      "contributor_one": "contributor",
+      "contributor_other": "contributors",
+      "item_one": "item",
+      "item_other": "items",
+      "progress_bar": "progress bar",
+      "file_one": "file",
+      "file_other": "files",
+      "table_of_contents": "table of contents",
+      "selection": "selection"
```
1aurend, line 207: Even if you're just using one, I think it makes sense to add both the singular and plural to glossary with common entities. I imagine "selections" is used somewhere in Manifold.

client/src/config/app/locale/en/en.json
```diff
@@ -105,6 +105,8 @@
       "cancel_delete": "$t(actions.cancel) $t(actions.delete)",
       "share": "Share",
       "toggle_to_state": "Toggle {{label}} to {{state}}",
+      "select_item": "Select {{ item }}",
+      "back": "Back",
```
1aurend, line 109: Nit: Back strikes me as navigation rather than an action, but that's just me.

client/src/config/app/locale/en/en.json
```diff
@@ -207,6 +254,58 @@
...
     },
+    "sections": {
+      "journal_info": "$t(glossary.journal_one) info",
+      "full_description": "Full description",
+      "tags_list": "Tags list",
+      "variants": "Downloadable Variants",
+      "resource_slideshow": "Resource slideshow",
+      "resource_list": "Resource list"
+    },
+    "reader": {
```
1aurend, line 265: I think, even at this stage, we should introduce some structure here and not just put everything under a reader block. Just off the top of my head, I could imagine something like:
```
"reader": {
   "actions": {
     ...keys here
   },
    "menus": {
       "settings":  {
         ...keys here
       },
       "appearance": {
         ...and so on
       }
    } 
}
```
Then we can decide later if we want to keep things under a top-level "reader" key or move some things into other categories. I'm surprised that "menus" hasn't come up yet elsewhere, but I can imagine we might later want a structure where "menus" is the top key and then "reader", "global", etc. are under it.

I'll just put [Jen's great notes](https://www.notion.so/castiron/Translations-best-practices-f9f5da36adeb4b948d6f6e3cb5c0e3e8#d0e9e065de9140a6b1007e5c1670e5df) on best practices here because I find them so helpful.

client/src/reader/components/annotation/Share/Citation.js
```diff
@@ -59,7 +61,9 @@ export default class AnnotationShareEditor extends PureComponent {
 
     return (
       <fieldset className="citation__radios">
-        <legend className="citation__legend">Citation style:</legend>
+        <legend className="citation__legend">
+          {this.props.t("reader.citation_style") + ":"}
```
1aurend, line 65: Consider adding punctuation like this to the translation. I'm not sure all languages use ":" in the same way.

client/src/reader/components/annotation/Share/Citation.js
```diff
@@ -85,7 +89,9 @@ export default class AnnotationShareEditor extends PureComponent {
   }
 
   render() {
-    const copiedText = this.state.copied ? "Copied!" : null;
+    const copiedText = this.state.copied
+      ? this.props.t("reader.copied") + "!"
```
1aurend, line 93: Likewise, re: punctuation.

client/src/reader/components/annotation/popup/menus/Main/Items/CurrentReadingGroup.js
```diff
@@ -37,9 +41,14 @@ function CurrentReadingGroup({
         data-name="readingGroup"
         className="annotation-popup__button annotation-popup__button--stacked annotation-popup__button--secondary-dark"
       >
-        <span className="screen-reader-text">Select reading group</span>
+        <span className="screen-reader-text">
+          {t("actions.select_item", { item: t("glossary.reading_group_one") })}
```
1aurend, line 45: Interpolation of keys should only happen inside the translation file. So, you might need to add a "select_reading_group" key in addition to "select_item":
```
"select_reading_group": "Select $t(glossary.reading_group_one)"
```

client/src/reader/components/control-menu/AppearanceMenuBody.js
```diff
@@ -68,37 +73,48 @@ class AppearanceMenuBody extends Component {
...
   }
 
   get decrementFontMessage() {
-    const baseMessage = "Reader font size decreased";
+    const baseMessage = this.props.t("reader.font_decreased");
     const appendedMessage =
-      this.fontSize.current > this.fontSize.min ? "." : " to minimum size.";
+      this.fontSize.current > this.fontSize.min
+        ? "."
+        : " " + this.props.t("reader.to_min");
```
1aurend, line 112: I know it seems like adding duplicates to the translation file, but we shouldn't use partial sentence translations with conditionals. I ran into some like this in `global` that I talked through with Dana the other day. These can be tricky!

I'd probably suggest having two translation keys for each message, each of which returns a full sentence, and then the conditional decides between keys rather than adds a phrase/punctuation. Maybe something like...

```
"font_size_decreased": "Reader font size decreased."
"font_size_decreased_max": "Reader font size decreased to max."

get decrementFontMessage() {
    return this.fontSize.current > this.fontSize.min ? this.props.t("reader.font_size_decreased") : this.props.t("reader.font_size_decreased_max");
}
```

1aurend (changes requested):
This looking great @LeonorPDX! It looks like you were able to get the hook and hoc working with no problem.

I've added comments about structuring some of the translations. Let me know if you have questions or disagree with anything. I'm pretty new to reviewing, so if anything isn't helpful, that's definitely useful info for me.

The tests should be an easy fix. See my response to your comment.

client/src/reader/components/annotation/Share/Citation.js
1aurend: Wanted to add, I haven't been consistent with this either, so it's on my list of things to go back and check for if we have time.

LeonorPDX:
@1aurend I made most of the changes you requested, I'll make further structure improvements to the reader translation keys as I finish up. I wanted to get this PR merged in as a starting point so I can make smaller PRs and merge them quickly going forward.

Linting and testing passes locally. How should I handle the conflicting snapshots? I haven't had to deal much with merge conflicts in the past.

Thanks again for all your help!

PR #3286, opened 2022-03-11
[POC] Localize FormattedDate

1aurend:


client/src/global/components/LanguageSelect/index.js
```diff
@@ -36,7 +36,8 @@ function LanguageSelect({ authentication, language }) {
   useEffect(() => {
     setLang(initialLanguage);
     updateI18n(initialLanguage);
-  }, [initialLanguage]);
+    updateLanguage(initialLanguage);
+  }, [initialLanguage, updateLanguage]);
 
   return (
     <Select
```
1aurend, line 43: @zdavis `state.ui.persistent.locale.language` was resetting to "en" on every route change. I assumed this wasn't the expected behavior and updated this `useEffect` to fix it. Was that the right thing to do?

client/src/global/components/FormattedDate/index.js
```diff
@@ -1,56 +1,66 @@
...
-      if (this.props.format === "distanceInWords")
-        return formatDistance(this.date, Date.now());
-      const outFormat = this.props.format ? this.props.format : "MMMM dd, yyyy";
-      return format(this.date, outFormat);
+      if (format === "distanceInWords")
+        return formatDistance(parsedDate, Date.now(), {
+          addSuffix: suffix,
+          locale
+        });
+      return formatDate(parsedDate, format);
```
1aurend, line 46: Add `locale` here.

client/src/global/components/LanguageSelect/index.js
dananjohnson: That doesn't seem right. I think this component may need another pass to get the store, `i18next`, and the internal state working in sync.

client/src/global/components/FormattedDate/index.js
```diff
@@ -1,56 +1,66 @@
...
+}) {
+  const lng = useSelector(state => state.ui.persistent.locale.language);
+  const locale = useMemo(() => {
+    switch (lng) {
+      case "es":
+        return es;
+      case "nl":
+        return nl;
+      default:
+        return null;
```
dananjohnson, line 23: Why does `en` become `null`? More generally, why is this necessary? I feel like I'm missing something.

dananjohnson: We should also all have a conversation about how available locales are defined in the client. I'm not sure that English, Dutch, and Spanish were ever determined as the options; I think they were just specified for testing. Ideally someone could define locales for the instance somewhere, but I don't think we'll have that in this first pass. In any case, we should be careful about hard-coding specific options anywhere, as presumably we'll need to remove those before v7 is released.

dananjohnson:
@1aurend oh that's a neat idea! I think I like it. Is there a reason in FormattedDate that you're not just doing this?
```js
const { t } = useTranslation();
...
formatDate(parsedDate, "P", { t.date_fns })
```
This looks overly complex, but I'm probably missing something:
```js
const lng = useSelector(state => state.ui.persistent.locale.language);
const { i18n } = useTranslation();
const locale = useMemo(() => i18n.store.data[lng].translation.date_fns, [
  lng,
  i18n.store.data
]);
```

1aurend:
> @1aurend oh that's a neat idea! I think I like it. Is there a reason in FormattedDate that you're not just doing this?
> 
> ```js
> const { t } = useTranslation();
> ...
> formatDate(parsedDate, "P", { t.date_fns })
> ```
> 
> This looks overly complex, but I'm probably missing something:
> 
> ```js
> const lng = useSelector(state => state.ui.persistent.locale.language);
> const { i18n } = useTranslation();
> const locale = useMemo(() => i18n.store.data[lng].translation.date_fns, [
>   lng,
>   i18n.store.data
> ]);
> ```

Sadly, that doesn't work. It was my first thought too. If you tell `t` to return an object, it returns a message that says something to the effect of "This key points to an object and not a string." rather than the object itself. We could get `lng` out of `i18n` rather than pulling it from the store, though?

dananjohnson:
@1aurend I did mean `t("date_fns")`. Too bad that doesn't work. I agree though about getting `lng` from `i18n`. Since we're relying on `i18next` to be the source of truth everywhere else. (On that note, we probably need to finalize how `i18next` and the redux store interact generally. It makes sense that we'd rely on `i18next` as the source of truth, but can set that value from the store initially.)

dananjohnson:
Oh actually, does this work? https://www.i18next.com/translation-function/objects-and-arrays#objects

1aurend:
> Oh actually, does this work? https://www.i18next.com/translation-function/objects-and-arrays#objects

Yes, yes it does. 🎊  I think we have a solution.

1aurend:
> @1aurend I did mean `t("date_fns")`. Too bad that doesn't work. I agree though about getting `lng` from `i18n`. Since we're relying on `i18next` to be the source of truth everywhere else. (On that note, we probably need to finalize how `i18next` and the redux store interact generally. It makes sense that we'd rely on `i18next` as the source of truth, but can set that value from the store initially.)

I made an issue last week for revisiting `LocationSelect`, but yes, definitely we should settle the question of `i18n` and `store` interaction generally. I don't know whether it's important for v7 or could be something we do down the line.

PR #3287, opened 2022-03-12
[F] Localize reader notes and return menu

LeonorPDX:


dananjohnson (changes requested):
Looking good! See my note about redundant keys.

dananjohnson (approved):

PR #3288, opened 2022-03-14
[C] Improve implementation of Trans component

1aurend:


1aurend:
@dananjohnson there's nothing new here, so this one can probably just be a quick click through. The counts are much improved, but would still be good to revisit to see how we can make them consistent across Manifold.

dananjohnson (approved):

PR #3290, opened 2022-03-15
[F] Localize Reader containers

LeonorPDX:


LeonorPDX:
@dananjohnson or @1aurend, can I get a quick peep on this PR when you have a moment?

client/src/config/app/locale/en/json/glossary.json
```diff
@@ -95,6 +95,7 @@
     "selection_one": "selection",
     "selection_other": "selections",
     "notes_title_case": "Notes",
+    "notes": "notes",
```
1aurend, line 98: Should we add plural/singular for these, just in case we need them for a count somewhere?

client/src/config/app/locale/en/json/reader.json
```diff
@@ -81,6 +82,8 @@
       "reader_appearance": "Reader Appearance"
     },
     "my_private_annotations": "My Private $t(glossary.annotation_title_case_other)",
-    "my_public_annotations": "My Public $t(glossary.annotation_title_case_other)"
+    "my_public_annotations": "My Public $t(glossary.annotation_title_case_other)",
+    "full_text": "Full Text",
+    "section_description": "Start reading this text on {{append}}"
```
1aurend, line 87: I might use a variable name here that would be more descriptive for the translator? `appName` or something? This isn't you; this is case where the old component uses an uninformative variable name.

client/src/reader/containers/ReaderFullNotes/index.js
```diff
@@ -99,10 +100,14 @@ function ReaderFullNotesContainer({
...
+  const { t } = useTranslation();
+  const myNotes = t("reader.menus.notes.my_notes"); // Set translations for getOverlayPropsForGroup(), cannot use hook inside a conditional statement
+  const allNotes = t("reader.menus.notes.all_notes");
+
   function getOverlayPropsForGroup() {
     return {
-      title: readingGroup === "me" ? "My Notes" : readingGroup.attributes.name,
-      subtitle: readingGroup === "me" ? null : "All notes for group:",
+      title: readingGroup === "me" ? myNotes : readingGroup.attributes.name,
+      subtitle: readingGroup === "me" ? null : allNotes,
```
1aurend, line 110: Right, so you can't call the hook conditionally, as in line 103 there. But you can use `t` conditionally once it's defined. I would just move lines 104-105 inside `getOverlayPropsForGroup`.

client/src/reader/containers/notation/Picker.js
```diff
@@ -143,38 +144,46 @@ export class NotationPickerContainerImplementation extends PureComponent {
...
+                label="options"
+                optionOne={{
+                  translatedLabel: t("glossary.resource_other"),
+                  label: "resources",
+                  icon: "resource24"
+                }}
+                optionTwo={{
+                  translatedLabel: t("glossary.collection_other"),
+                  label: "collections",
+                  icon: "resourceCollection64"
```
1aurend, line 163: See comment above. I think you can just handle the localization here.

1aurend:
> @dananjohnson or @1aurend, can I get a quick peep on this PR when you have a moment?

@LeonorPDX I did a quick click through, and everything looks good to me. Just left you  a few minor questions/comments above.

client/src/frontend/components/utility/Toggle.js
```diff
@@ -68,7 +68,9 @@ class Toggle extends Component {
             className="button-switch-primary__icon"
           />
         )}
-        <span className="button-switch-primary__label">{option.label}</span>
+        <span className="button-switch-primary__label">
+          {option.translatedLabel ? option.translatedLabel : option.label}
```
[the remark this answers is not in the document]
LeonorPDX: The toggle is used by other components that haven't been localized yet, I so left that check there so they wouldn't break the toggle if they don't have a translated label yet. I could just go through and localize those components as well now, and then we wouldn't need this check.

client/src/reader/containers/notation/Picker.js
LeonorPDX: The label is used in class names and managing state, so we need that value without translation. Only the displayed label should be translated.

client/src/reader/containers/notation/Picker.js
```diff
@@ -143,38 +144,46 @@ export class NotationPickerContainerImplementation extends PureComponent {
...
-              onRowClick: this.props.selectionHandler
-            }}
-            search={
-              <Search {...this.props.entitiesListSearchProps("notations")} />
-            }
-          />
-        </div>
-      </section>
+      <Translation>
+        {t => (
```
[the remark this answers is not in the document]
LeonorPDX: Yes, I tried wrapping the export with `withTranslation` but I get an error saying t is not defined. I'm probably missing something in the props/state construction on this component, any tips?
Here's the props at the top of the component, I added `t: PropTypes.func` in static PropTypes:
```
export class NotationPickerContainerImplementation extends PureComponent {
  static mapStateToProps = (state, ownProps) => {
    const newState = {
      resources: select(requests.beResources, state.entityStore),
      resourcesMeta: meta(requests.beResources, state.entityStore),
      resourceCollections: select(
        requests.beResourceCollections,
        state.entityStore
      ),
      resourceCollectionsMeta: meta(
        requests.beResourceCollections,
        state.entityStore
      )
    };
    return { ...newState, ...ownProps };
  };

  static displayName = "ReaderContainer.Notation.Picker";

  static propTypes = {
    projectId: PropTypes.string,
    resources: PropTypes.array,
    resourcesMeta: PropTypes.object,
    resourceCollections: PropTypes.array,
    resourceCollectionsMeta: PropTypes.object,
    dispatch: PropTypes.func,
    selectionHandler: PropTypes.func,
    t: PropTypes.func
  };

  constructor(props) {
    super(props);
    this.state = {
      context: "resources"
    };
  }
```
And here's the export at the end of the file. I tried wrapping `withTranslation` in a couple places and got errors every time.
```
export const NotationPickerContainer = withFilteredLists(
  NotationPickerContainerImplementation,
  {
    notations: keywordFilter()
  }
);

export default connect(NotationPickerContainer.mapStateToProps)(
  NotationPickerContainer
);
```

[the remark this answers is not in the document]
LeonorPDX: Just using the `<Translation>` render prop was easier than sorting out what was going on, but I'm happy to switch to `withTranslation` if you can help me figure it out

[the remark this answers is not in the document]
1aurend: Oh, I think that's fine. I was really just curious! I haven't run into a case where I haven't been able to use the hoc relatively easily, so it was interesting to see one. I have found that with `connect()`, `withTranslaton()` has to go on the outside, fwiw.

client/src/reader/containers/notation/Picker.js
1aurend: Gotcha. Ideally, we'd solve this by doing something different with that `selected` prop, so we don't have string match to, e.g. "collections", especially since the toggle is only used in a couple of places. But you're right. Keep what you have.

PR #3298, opened 2022-03-25
[F] Localize global components

1aurend:


1aurend:
@LeonorPDX @dananjohnson the rest of the global components are ready for a second set of eyes when you get a chance. The only things in global that are outstanding now are components/forms and components/sign-in-up, and components/utility/EntityCount which I'm leaving until the end so we can make sure that plurals are handled by the localized props coming in before we remove `pluralize` from there. I've made issues in linear for all of these.

LeonorPDX (approved):

PR #3301, opened 2022-03-30
[F] Localize global forms + sign-in overlay

1aurend:


1aurend:
@dananjohnson would you be able to take look at this PR? Is that huge number of lines added by the snapshot updates because of how I'm importing the locale object? Also, I'd love a second set of eyes of `global/forms/DatePicker`. Thanks!

dananjohnson:
@1aurend I took a quick look at rebasing on `master` and updating snapshots but encountered a number of failing tests. Please review when you're back and then merge once all actions are passing.

PR #3303, opened 2022-03-30
[F] Create schemas journals + issues

1aurend:


client/src/global/components/schema/Types/Journal.js
```diff
@@ -0,0 +1,108 @@
...
+    createdAt,
+    updatedAt,
+    avatarStyles
+  } = attributes;
+
+  const journalData = {
+    "@type": "Periodical",
+    "@id": metadata.issn,
+    name: title,
+    url: lh.link("frontendJournalDetail", slug),
```
dananjohnson, line 48: This needs to be a full URL, not a relative one.

client/src/global/components/schema/Types/Journal.js
```diff
@@ -0,0 +1,108 @@
...
+    issn: metadata.issn,
```
dananjohnson, line 49: Let's add ISBN and DOI properties using `identifier` (https://schema.org/identifier).

client/src/global/components/schema/Types/Issue.js
```diff
@@ -0,0 +1,89 @@
...
+  } = attributes;
+  const { creators, contributors } = relationships;
+
+  const issueData = {
+    "@type": "PublicationIssue",
+    "@id": metadata.doi,
+    name: title,
+    issueNumber: number,
+    url: lh.link("frontendProjectDetail", slug),
+    issn: metadata.issn,
```
dananjohnson, line 39: I didn't see that `issn` was a valid property for PublicationIssue. You may need to add it using `identifier`. Let's add ISBN and DOI at the same time.

client/src/global/components/schema/Types/Journal.js
```diff
@@ -0,0 +1,108 @@
...
+    copyrightHolder: metadata.rightsHolder,
+    copyrightNotice: metadata.rights,
+    dateCreated: createdAt,
+    dateModified: updatedAt,
+    datePublished: publicationDate,
+    publisher: metadata.publisher,
+    hasPart: [...renderIssues(), ...renderVolumes()],
```
dananjohnson, line 56: This is great! One thing I did notice in the documentation is that issues are nested in volumes: https://schema.org/PublicationIssue#eg-0398. If it's not too complicated, I think that would be good to do (at least for issues that do belong to volumes).

client/src/global/components/schema/Types/Issue.js
```diff
@@ -0,0 +1,89 @@
...
+    number
+  } = attributes;
+  const { creators, contributors } = relationships;
+
+  const issueData = {
+    "@type": "PublicationIssue",
+    "@id": metadata.doi,
+    name: title,
+    issueNumber: number,
+    url: lh.link("frontendProjectDetail", slug),
```
dananjohnson, line 38: need full, not relative, URL

client/src/global/components/schema/Types/Project.js
```diff
@@ -118,35 +79,22 @@ export default class Project extends PureComponent {
       name: title,
       url: lh.link("frontendProjectDetail", slug),
```
dananjohnson, line 80: need full, not relative, URL (my bad! 🙃 )

dananjohnson (changes requested):
This looks really good! I like how you've kept things DRY as you've built out the different schemas. See just a few small changes below.

dananjohnson (approved):

PR #3306, opened 2022-03-31
[C] Refactor FE containers

1aurend:


1aurend:
@dananjohnson here are the FE containers minus `ResourceDetail` and `ResourceCollectionDetail`. There's one test that I can't get. For some reason, the `ProjectResources` container doesn't seem to mount the same way as the others. Do you mind taking a look before I spend more time on it?

client/src/frontend/containers/EventList/index.js
```diff
@@ -1,111 +1,72 @@
...
+    </>
+  );
 }
 
-export default withTranslation()(connectAndFetch(EventList));
+EventList.displayName = "Frontend.Containers.EventList";
+
+EventList.propTypes = {
+  project: PropTypes.object,
+  journalBreadcrumbs: PropTypes.object
```
dananjohnson, line 71: Seeing a prop-type warning for this one:
```
Failed prop type: Invalid prop `journalBreadcrumbs` of type `array` supplied to `Frontend.Containers.EventList`, expected `object`.
```

dananjohnson (changes requested):
Great work, @1aurend! Just noticed one issue: something's up with the Resources container. I can navigate tot= that route client-side, but if I try to load the URL directly there's a 500 error. Check that out and the little prop-type warning below. Otherwise this looks good to go!

1aurend:
@dananjohnson turns out I was missing an import null guard in the resources container. When you navigate via the project, the project is already fetched by the wrapper but not when loading the page directly. Thanks for the catch!

zdavis:
@1aurend, as far as you're concerned, is this good to go?

1aurend:
> @1aurend, as far as you're concerned, is this good to go?

It's ready except for these markup warnings that I couldn't figure out how to solve:
Screen Shot 2022-05-04 at 9 01 49 AM

zdavis:
Ok, I'll see if I can figure out the warning.

dananjohnson:
@1aurend when I pulled this down to take a look at it again, I noticed exceptions when visiting the resources page for a project that doesn't have any resources. Interestingly I didn't notice the server warning you mentioned above with resource collections. The placeholder for that page seems to render without issue. Do you want to take a look at the resources error? If there's not an easy fix or you don't have time, then just open up a new PR with the FatalError fix and we'll merge that in separately for v7.

dananjohnson (approved):

PR #3323, opened 2022-04-07
[B] Show features on homepage

### LINKED ISSUE #3320: v7-alpha doesn't promote published Features to the HP
When a Feature is toggled to `Published`, it does not appear as expected on the Manifold HP in v7-alpha.

See the Apparatus instance:

- Feature: `https://apparatuspress.manifoldapp.org/backend/records/features`
- Homepage: `https://apparatuspress.manifoldapp.org`

1aurend:
Resolves #3320


1aurend:
@zdavis The features fetch seemed like it was no longer happening, so I refactored the remaining home containers to use the new hooks. Would you mind taking a look before I merge?

zdavis:
Thanks, Lauren. I'll take a look at this when I'm back on Manifold. I want to understand why the fetch isn't happening, since it could be indicative of a problem elsewhere.

zdavis:
Looks like I removed the fetch call when I added the fetch data hook—which is to say, I'm not seeing a widespread problem with the deprecated fetch data approach.

PR #3324, opened 2022-04-07
[B] Fix style regressions in search results

1aurend:


dananjohnson (changes requested):
A handful of little corrections. I won't be around tomorrow but if you get them made, I can merge in the PR on Monday.

* padding at bottom of result looks too small
Screen Shot 2022-04-07 at 5 42 26 PM

* byline font-size is 18px in design vs 16px in implementation
* the parent title color matches design, but it doesn't meet contrast requirements. Let's use `var(--color-neutral-text-dark)`.
* the wrapping here looks weird. Probably makes sense to ditch `flex-wrap: wrap` here. v6 doesn't use it (https://manifold.umn.edu/search).
Screen Shot 2022-04-07 at 5 52 18 PM

* increase the margin separating the list and pagination to match the comps.
Screen Shot 2022-04-07 at 5 57 18 PM

1aurend:
@dananjohnson made the changes. Should be ready for you to double check and merge.

dananjohnson (approved):

PR #3293, opened 2022-03-18
[F] Localize backend components

LeonorPDX:

PR #3345, opened 2022-04-19
[C] Migrate FE resource + resource collection styles

1aurend:
Resolves MNFLD-350


client/src/frontend/components/resource/Card/styles.js
```diff
@@ -0,0 +1,105 @@
...
+  width: 62.2%;
+  padding: 12px 15px;
+  text-decoration: none;
+  background-color: var(--color-base-neutral05);
+
+  ${respond(`width: calc(100% - 155px);`, 40)}
+
+  ${respond(`width: calc(100% - 135px);`, 40)}
+
+  ${respond(`width: calc(100% - 155px);`, 40)}
```
dananjohnson, line 40: These are all set at the same breakpoint! 🤔

1aurend: Ha! Did I do that?

client/src/frontend/components/resource-collection/Description/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+  return (
+    <Styled.Wrapper>
+      {date && (
+        <Styled.DateWrapper>
+          <FormattedDate
+            prefix={t("dates.collection_created")}
+            format="MMMM yyyy"
+            date={date}
+          />
```
dananjohnson, line 19: Make sure to use a valid format for the `datetime` attribute (needs dashes not slashes). https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time#valid_datetime_values
Screen Shot 2022-04-21 at 10 43 07 AM

client/src/frontend/components/resource-collection/Description/index.js
```diff
@@ -0,0 +1,33 @@
...
+        </Styled.DateWrapper>
+      )}
+      {description && (
+        <Styled.Description
+          dangerouslySetInnerHTML={{
+            __html: description
+          }}
+        />
+      )}
```
dananjohnson, line 28: This rendered for me when `description` was just an empty string. Might need to check `!!description` instead.

client/src/frontend/components/composed/EntityCollection/patterns/ProjectResourceCollectionDetail.js
```diff
@@ -43,10 +43,9 @@ function ProjectResourceCollectionDetail({
...
       collectingProps={{ collectable: resourceCollection }}
       DescriptionComponent={props => (
-        <FormattedDate
-          prefix={t("dates.collection_created")}
-          format="MMMM yyyy"
+        <ResourceCollection.Description
           date={resourceCollection.attributes.createdAt}
+          description={resourceCollection.attributes.descriptionFormatted}
           {...props}
         />
```
dananjohnson, line 50: Yes I like this better than what I had done!

client/src/frontend/components/resource-collection/Description/styles.js
```diff
@@ -0,0 +1,24 @@
...
+`;
+
+export const Wrapper = styled.div`
+  margin-block-start: -30px;
+`;
+
+export const Description = styled.div`
+  margin-block-start: 35px;
+  font-family: var(--font-family-copy);
+`;
```
dananjohnson, line 24: Compare color and font size to design, which is a bit larger and darker.

dananjohnson (changes requested):
- [ ] Update Add Comment design (see Frontend - Resources page in Sketch)
Screen Shot 2022-04-21 at 10 18 51 AM

- [ ] fix comment textarea font size and width
Edge:
Screen Shot 2022-04-21 at 10 21 00 AM

Implementation:
Screen Shot 2022-04-21 at 10 21 06 AM

- [ ] Are the audio waves that have already been played supposed to have a background color? IIRC they used to be the Manifold accent color.
Screen Shot 2022-04-21 at 10 24 27 AM

- [ ] Latest design has moved the media object into the left column, with the right column aligning to the top with the title in the left column. Take the time to implement a mode modern layout than floats, as you suggested. I would recommend a Flexbox with wrapping enabled to avoid dealing with breakpoints, but if you prefer to use Grid, go for it.
Screen Shot 2022-04-21 at 10 28 40 AM

- [ ] A couple metadata presentation issues: 1) some labels haven't been localized correctly (I didn't test them all, just a few). 2) In latest design, tags have a "Tags:" label.
Screen Shot 2022-04-21 at 10 30 36 AM

- [x] There are some weird things going on with the slideshow. See the pagination buttons and the caption. I don't think "Read more" is supposed to appear for a short caption like this. The bottom left buttons should also be at the bottom of the container, whether there is a long or short caption.
Screen Shot 2022-04-21 at 10 50 39 AM
Screen Shot 2022-04-21 at 10 53 15 AM

client/src/frontend/components/resource-collection/Description/index.js
1aurend: This seemed hard. Both `date-fns` and `Date.toLocaleDateString()` use slashes. I couldn't come up with a programmatic way to get localized format and dashes. Ended up writing this function:
```
const dateTime = () => {
    try {
      let dateTimeFormat;
      switch (format) {
        case "MMMM, yyyy":
        case "MMMM yyyy":
          dateTimeFormat = "M yyy";
          break;
        case "PPpp":
          return parsedDate.toISOString();
        default:
          dateTimeFormat = "P";
          break;
      }
      return formatDate(parsedDate, dateTimeFormat, {
        locale: t("date_fns", { returnObjects: true })
      })
        .replaceAll("/", "-")
        .replaceAll(" ", "-");
    } catch {
      return "";
    }
  };
```
Since we only use a few formats in Manifold, it seems to work, but it also feel fragile. Let me know if you have other ideas!

1aurend:
Before I forget, just wanted to note that the missing metadata translation keys should be fixed by #3293 when that goes in.

1aurend:
OK— @dananjohnson back over to you for a final review and if you want to take a closer look at the collapse implementation.

Also, I'm stuck with this prop mismatch warning after updating `FormattedDate` and I don't remember what that fix is. Can you remind me?
Screen Shot 2022-04-22 at 4 53 55 PM

client/src/frontend/components/resource-collection/Description/index.js
dananjohnson: I agree, this is probably too brittle. If you look at that [MDN doc I mentioned](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time#valid_datetime_values), it doesn't have to be in a `03-03-2022` format. There are a lot of possibilities, just not slashes. I assume one of the other options will be easier to achieve?

dananjohnson (changes requested):
@1aurend the design revision to the resource detail route looks great! A couple little issues with the "Add Comment" section:

- [x] in design, it appears within in the left column, not below:
Screen Shot 2022-04-25 at 10 13 40 AM

- [x] For a11y, it's necessary to link the disclosure to the hidden content, and advisable to shift focus when the content is open and closed. That probably makes this a good candidate for a Collapse context. When the user authenticated, focus should move to the textarea. When not, it should go to the Log In button. It's helpful to give the form an ARIA label too. Simplest would be to use `aria-labelledby` with the ID of the disclosure.

Things look really close in general. I took a closer look at the slideshow caption—thoughts below—and noticed a couple other little issues with the slideshow too:

- [ ] Something's up with the resource collection slideshow fetching subsequent slices of resources. I made a collection of 11 items. Sliding through the first 10 works great, but when I try to go to the 11th, it goes back to the first slide, and the first 10 are now in a loading state. The API request looks good, and if I go back to the 11th slide it's there. But it seems like the second batch of 10 slides is replacing the first 10 in state rather than being appended. Or something like that.
Screen Shot 2022-04-25 at 10 26 24 AM

- [ ] Looks like a l10n issue in the slide description.
Screen Shot 2022-04-25 at 10 34 30 AM

- [ ] After taking a closer look at the collapsible slide caption, my feeling is that we shouldn't use the Collapse components here. As you found, visually it doesn't quite work in this context since the "collapsed" content is never really hidden. So you had to come up with a workaround, which has this flicker we need to try to solve. More importantly IMO, the a11y features of the Collapse components don't really make sense here, since this is really just a visual effect, and we're never showing/hiding content from screen readers. Here's what I would do: 1) restore the original implementation of the collapse effect; 2) use `aria-hidden` on the show/hide toggle, since it's not useful for SR users. How does that sound?

dananjohnson:
@1aurend could you tackle this issue while you're finalizing your PR? https://github.com/ManifoldScholar/manifold/issues/2821

Should be quick. Thanks!

dananjohnson (approved):

PR #3354, opened 2022-04-28
[C] L10n cleanup

1aurend:


dananjohnson (commented):
@1aurend, is there a reason you couldn't delete `localeOld`? Ideally, the keys there would be merged in with those in `client/src/config/app/locale`.

1aurend:
> @1aurend, is there a reason you couldn't delete `localeOld`? Ideally, the keys there would be merged in with those in `client/src/config/app/locale`.

I can get most of the keys merged. When I realized the file is used by a couple of hocs, and we decided to leave those for v8, I stopped. Happy to take one more pass through tomorrow and get a clearer sense of what it would take to be done with the file.

dananjohnson:
OK that'd be good. If we can't lose the file, let's at least delete the commented out parts and rename the file. Could it be moved into `client/src/config/app/locale` as well? I just think it'll be confusing for localizers/future us what this file is.

1aurend:
OK— file ditched! For the hocs that used it, I just moved the objects into the respective files with a note to localize later.

dananjohnson (approved):

PR #3357, opened 2022-04-29
[C] Reorganize atomic/composed components

dananjohnson:


dananjohnson:
@1aurend thank you for your feedback! I really liked all your suggestions and went ahead and implemented them. 🙏🏻

PR #3361, opened 2022-05-04
[F] Show tags on resource cards

1aurend:
Resolves MNFLD-364


1aurend:
@zdavis Did I do this right? Is it okay that I did?

PR #3428, opened 2022-08-30
[B] Prevent default on clicks inside drawers

### LINKED ISSUE #3399: Journals layout add background image "Upload a file" button just jumps up the page
We're playing around with JITP test journals and when we go to add a background image in Layout (tested in Chrome and Firefox), the "upload a file" link hops up the page. Drag and drop behaves as expected.

## Expected Behavior
You would expect Manifold to open your file finder.

## Steps to Reproduce
1. Enter Admin mode
2. In a Journal, navigate to Layout>Description+Images
3. in shelf, at "background image"
4. click "upload a file"

https://cuny.manifoldapp.org/backend/journals/af82626f-119f-484b-8665-bd8f358d39f1/layout

## Context (Environment)
In both Chrome and Firefox the "upload a file" jumps up the page rather than opening a window to browse files.

1aurend:
Fixes #3399

### DIFF
--- a/client/src/global/components/drawer/Wrapper.js
+++ b/client/src/global/components/drawer/Wrapper.js
@@ -106,6 +106,7 @@ class DrawerWrapper extends PureComponent {
 
   componentDidMount() {
     window.addEventListener("keyup", this.handleLeaveKey);
+    window.addEventListener("mousedown", this.preventDefaultOnChildClicks);
     this.enableScrollLock();
   }
 
@@ -128,6 +129,7 @@ class DrawerWrapper extends PureComponent {
 
   componentWillUnmount() {
     window.removeEventListener("keyup", this.handleLeaveKey);
+    window.removeEventListener("mousedown", this.preventDefaultOnChildClicks);
     this.disableScrollLock();
   }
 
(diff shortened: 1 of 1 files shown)


1aurend:
@dananjohnson I was able to reproduce this, but only when clicking on a child element with the top of the drawer scrolled out of view. This fix seems to work. I used [this thread](https://github.com/focus-trap/focus-trap-react/issues/14) as a reference. But I think there has to be a better way. Would you mind taking another look?

dananjohnson:
This solves the focus trap issue, but it introduces other problems. For instance, (`textarea`) elements in the drawer no longer receive focus on click, so they can't be edited. I'd expect there to be other issues if we prevent default on all `mousedown` events.

The particular form controls Krysia mentioned as not working correctly are radios and image upload fields. In both cases, the design of the component is such that the interactive element itself (`<input>`) doesn't receive the click directly, so `focus-trap` seems to treat clicks on them as happening outside the trap. As a result, it resets focus back to the first interactive element.

I'll take a pass at refactoring these controls so that the inputs themselves get the click. If that doesn't work, maybe we can do a more scoped `preventDefault` in the drawer.

What a pesky issue!

1aurend:
Closing because this is resolved by #3431.

Issue #2900, opened 2020-12-08
Add support for Google Analytics 4


tsmyre:
Per our conversation in Slack, we want to add support for the newest version of Google Analytics (GA4), which came out this past October.

zdavis:
Thank you for taking the time to open this feature request. The Manifold team reviewed this issue during our bi-weekly meeting and the consensus is that this feature makes sense and is in keeping with our overall vision for the platform. Moreover, we see this request as a viable candidate for development under our current available funding. We’re adding a “planned” label to this feature to indicate that we plan on completing it within our current funding cycle.

_This was an automated message, but please don't hesitate to reply. Our team watches these issues closely and will respond as soon as we're able to!_

zdavis:
@1aurend this issue relates to your GDPR work. Could you reference it in your PR so we're sure it's closed when that branch goes in?

1aurend:
Done

1aurend:
Released in v7.1

PR #3430, opened 2022-09-01
[F] Add GDPR functionality to Manifold

1aurend:
Resolves:
- #2900


client/src/global/components/sign-in-up/form-inputs/AvatarDropzone/AvatarDropzone.js
```diff
@@ -0,0 +1,100 @@
...
+
+  const { watch, setValue } = useFormContext();
+
+  const showAvatar = !!watch("avatar");
+  const removeAvatar = () => {
+    setValue("avatar", null, { shouldDirty: true, shouldValidate: false });
+  };
+  const handleDrop = file => {
+    setValue("avatar", file[0], { shouldDirty: true, shouldValidate: false });
+  };
```
jen-castiron, line 22: This is looking good! You could set the avatar's values here to what you're transforming them into [on submit](https://github.com/ManifoldScholar/manifold/pull/3430/files#diff-4596c2e59267d1d0e39a5c495c48b17ba021fc2774d711199f48ee0bc7526d4aR73):
```
setValue("avatar", {
   file: file[0],
   content_type: file[0].type,
   filename: file[0].name
})
```
But since you still need to get the image data from the avatar, and you don't need to read the file data until submit, I don't think this change would make much of a difference.

For removing the avatar, you might be able to set that value here. I think that's what we're doing on NGLP - setting the remove value inside the image component.
```
const removeAvatar = () => {
   setValue("removeAvatar", true);
   ...
}
const handleDrop = () => {
   setValue("removeAvatar", false);
   ...
}
```
If this component is shared, you could pass in a "removeFieldName" to this component.

All in all, my suggestions are minor improvements. This solution looks good to me as is 👍

Issue #3489, opened 2022-11-07
Can only view first 20 Project Collections in the backend


Ian-BD:
When viewing project collections in the Backend you are limited to the first 20 based on the sort order with no way to to paginate. these means there's no easy to view all Project collections once you have more than 20.

## Expected Behavior
Like the front end there should be options to click through pages to view all Project collections.

## Current Behavior
Only the first 20 Project collections are viewable.

## Possible Solution
The simplest solution seems to be adding pagination below the list.

## Steps to Reproduce

1. Access backend.
2. Create 21 Project collections.
3. Notice only 20 are viewable

## Context (Environment)
Our client is unable to add and modify project collections easily, my current workaround is for them to email me when they've added a new one and I can find a direct link by calling the Manifold API.

## Notes
I noticed this issue when looking at the API request in the Network tab of Firefox's Dev console, it appears that pagination is a feature of the API but there's no way to alter it.
![image](https://user-images.githubusercontent.com/89195159/200337670-14a5a759-1bc4-417e-8e74-b0e11a84cd41.png)

zdavis:
Fixed in #3493

Issue #3487, opened 2022-11-02
Scrolling issue on visibility drawer


mkgold:
When I am on the DDH Manifold instance, I find that if I am looking at a text as a logged in viewer and try to hit the viewing options modal drawer (the eye icon), the drawer opens, but I'm unable to scroll down to see all of the options because the scroll focus is on the background text rather than the modal.

Please see attached screenshot and let me know whether you need any other info to replicate.

<img width="1464" alt="Screen Shot 2022-11-02 at 5 36 22 PM" src="https://user-images.githubusercontent.com/278503/199607313-726cb31c-a5e4-484a-9a20-83357b7ff2c2.png">

tsmyre:
Just confirming that I can replicate this behavior on Brave.

zdavis:
Fix will go out in next version.

PR #3493, opened 2022-11-08
[B] Allow scroll in reader menu + project collections list

1aurend:
Also resolves:
- MNFLD-442
- #3487, #3489

### DIFF
--- a/client/src/theme/styles/components/backend/project-collection/collectionList.js
+++ b/client/src/theme/styles/components/backend/project-collection/collectionList.js
@@ -26,6 +26,11 @@ export default `
         breakpointLessOne(75),
         "max"
       )}
+
+      ul {
+        max-height: 1270px;
+        overflow-y: scroll;
+      }
     }
 
     .actions {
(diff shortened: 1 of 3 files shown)


client/src/theme/styles/components/backend/project-collection/collectionList.js
```diff
@@ -26,6 +26,11 @@ export default `
         breakpointLessOne(75),
         "max"
       )}
+
+      ul {
+        max-height: 1270px;
+        overflow-y: scroll;
```
dananjohnson, line 32: Always use `auto` over `scroll` unless you are certain that you want a scrollbar to be visible under all circumstances. `auto` will only show one if there is actually content overflowing the container.

client/src/theme/styles/components/reader/controlMenus/visibility.js
```diff
@@ -5,6 +5,8 @@ export default `
   .visibility-menu {
     font-family: var(--font-family-heading);
     ${uiPanelMenu}
+    overflow-y: scroll;
```
dananjohnson, line 8: Use `auto`.

dananjohnson (changes requested):

PR #3510, opened 2022-11-29
[B] Fix drawer top position

dananjohnson:
Value was set to the wrong CSS custom property


1aurend:
FYI: I included this fix in my migration of drawer styles that will go in #3512.

dananjohnson:
Sounds good. I'll close this then, yeah?

PR #3512, opened 2022-11-30
[C] Update focus-trap-react

1aurend:


1aurend:
@dananjohnson Thank you for your careful review, as always! Taking your notes in order:

- I noticed that switch not behaving once or twice, but not consistently. It seems to be working fine for me this morning? But regardless, I'll update the css.

- This is tricky! It's broken on Edge too, though in a different way. I think I understand now what's going on, but still working on how to fix.

- Agreed re: making drawers consistent. There are a few other drawers like this one; it's any drawer that doesn't close with a route change. However, it's not a trivial fix, and I'd be tempted to do it as a separate PR. I think there are two options: (1) Add routes so that all drawers with a Save trigger the route change prompt in the same way using global Form, or (2) use the `withConfirm` hoc on drawers that aren't routes and check form state in the close handler. Do you have a preference?

1aurend:
Ok! I think I got the ESC drawer + confirm right for real this time 🙃

dananjohnson:
Makes sense about doing a separate PR. I think we've done option #2 elsewhere, so that sounds fine to me. (The archive and delete buttons in RG settings use the HOC, I think.)

1aurend:
@dananjohnson Do you mind giving this another quick 👀 and unblocking merging when you get a chance? Thanks!

dananjohnson (approved):
Looks good to me!

PR #3514, opened 2022-12-01
[C] Migrate drawer styles

1aurend:


dananjohnson (changes requested):
Nice! Just noticed a couple oddities with the reader menus:

- [x] The TOC menu slides in from the right.
- [ ] If the Back to Project menu (the far left one) is open and you click "Menu" to close it, it immediately reopens. Might not be a new bug.

1aurend:
Thank you @dananjohnson for catching that issue with the back menu. It was a symptom of a larger issue with how I had refactored the global click handler into the outside click handler. I think all these things will be easier to handle going forward now that they're all located in the trap options, but finding all the event handlers and consolidating them... 🤯

dananjohnson:
No problem! So much wranging… 😰

PR #3518, opened 2022-12-05
[F] Implement entitlements import API

scryptmouse:

PR #3521, opened 2022-12-07
[C] Speed up a number of slower API tests

scryptmouse:
* Add test-prof / let_it_be to do one-time initializations of mostly read-only data during slower tests
* Some idiomatic clean-ups of older tests


scryptmouse:
This should go in after / over #3518 because it relies on a few updates in that branch.

zdavis:
Comes in after #3549 now.

PR #3526, opened 2022-12-19
[F] Client updates for entitlement imports

1aurend:
*Includes commits from #3518


api/app/controllers/api/v1/pending_entitlements_controller.rb
```diff
@@ -13,7 +13,7 @@ class PendingEntitlementsController < ApplicationController
       INCLUDES = %i[entitlement user subject].freeze
 
       resourceful! PendingEntitlement, authorize_options: { except: [:index] } do
-        PendingEntitlement.preload(*INCLUSIONS).in_default_order.filtered(
+        PendingEntitlement.preload(*INCLUSIONS).by_state(:pending).filtered(
```
1aurend, line 16: Just flagging that it looks like I caused tests to fail by doing this. I (think I) needed to remove `in_default_order` for the order filter to work. I added the state scope because the default in strong params never gets used; there's always a filters param from the client. But I can also always send `state: pending` from the client.

api/app/serializers/v1/concerns/project_serializer.rb
```diff
@@ -26,6 +26,7 @@ module ProjectSerializer
         typed_attribute :avatar_styles, Types::Serializer::Attachment.meta(read_only: true)
         typed_attribute :is_journal_issue, Types::Bool.meta(read_only: true), &:journal_issue?
         typed_has_many :creators, serializer: ::V1::MakerSerializer, record_type: :maker
+        typed_attribute :entitlement_subject_url, Types::String.meta(read_only: true)
```
1aurend, line 29: This also seems to have caused tests to fail.

PR #3530, opened 2023-01-04
[F] Add support for MathML

1aurend:


1aurend:
@dananjohnson No rush, but I'm hoping you can give this an a11y review when you get a chance. I made some progress applying annotation styles to math elements and replacing their content in annotation lists. However, in addition to the things we've already discussed, I've noticed a couple of possible issues with keyboard support:
- If we add the same interactive attributes / onClick to the math part of the highlight as to the surrounding text spans, there are keyboard controls for the span and then keyboard controls for its mjx element child, if accessibility is turned on. This might be fine, but it took me a second to figure out. Are the aria attributes correct and clear enough here?
  Screen Shot 2023-01-09 at 10 19 52 AM
  Pressing `enter` here on the span opens the annotation drawer:
  Screen Shot 2023-01-09 at 10 25 43 AM
  Pressing `enter` here on the mjx element opens the MathJax explorer:
  Screen Shot 2023-01-09 at 10 23 30 AM

- It takes a number of presses of tab to tab through a math element because of tabbable internal elements MathJax creates. However, there are no associated focus styles, so it's hard to tell where you are.

Also, not a11y, but selecting sentences or paragraphs that contain math seems to work in Chrome/Safari, even though the math does not appear selected. It's buggy, however, in FF. FF seems to not accurately select text characters following the math elements. Can we improve this?

1aurend:
Note to self: remove MathJax deps before merging if we decide to ahead without it.

client/src/reader/components/section/body-nodes/nodes/Math/Node.js
```diff
@@ -0,0 +1,74 @@
...
+  const {
+    classes,
+    removableHighlightId,
+    textAnnotationIds,
+    annotationIds,
+    interactiveAttributes
+  } = getAnnotationStyles(localAnnotations, uuids, t, hasInteractiveAncestor);
+
+  const Wrapper = attributes.display === "inline" ? "span" : "div";
+  const wrapperStyles = Wrapper === "div" ? { width: "max-content" } : {};
```
dananjohnson, line 52: You may want to add `{ maxWidth: "100%" }`. `width: max-content` will cause an element to ignore the width of its parent and extend right off the page if it's really that wide.

client/src/theme/assets/fonts/FiraMath/mathfonts.css
```diff
@@ -0,0 +1,21 @@
...
+         url('FiraMath-Regular.woff2'), url('FiraMath-Regular.woff');
+}
+
+.htmlmathparagraph, m|mtext {
+    /* TODO: use FiraGO instead? */
+    font-family: Fira Math;
+}
+m|math {
+    font-family: Fira Math;
+}
```
dananjohnson, line 21: It doesn't look like this file is being used anywhere.

dananjohnson (changes requested):
@1aurend super cool feature! Great work on this. I just reviewed the code, as we've walked through it together and it appears to work as expected in the application. If there's any browser or a11y testing you'd like me to do though, let me know.

dananjohnson (approved):
APPROVED!

PR #3539, opened 2023-01-24
[F] Implement announcements & focus management for client-side routing

1aurend:
@dananjohnson here's this. I think the two things left to do before it goes in are (1) figure out the blur bug, and (2) standardize page titles in admin, so they're not so verbose.


dananjohnson:
@1aurend just wanted to say that this is still on my radar! I will take a look next week when I'm back to full-time.

dananjohnson:
I also think you can remove the `useBlurOnLocationChange` hook, since it's not being used anywhere. It seems like we'll just manage this in one place in the application, but if we really need to abstract it later, we can add the hook then.

dananjohnson (approved):

PR #3541, opened 2023-01-25
[F] Handle text section body updates

1aurend:
Adds callbacks for updating `body_json` and calculating `has_mathml` when section body is changed.


zdavis:
This looks good to me, although it might be nice to have a spec for the mathml check.

1aurend:
Was wondering about that. I'll add one.

PR #3559, opened 2023-02-15
[F] Extrapolate nodes from text sections; add to annotations

scryptmouse:
* Use SQL formatted-schema instead of Ruby
* Calculate a consistent `body_hash` based on the stringified version
  of a text section’s `body_json`.
* Add table and maintenance operation / job to extrapolate nodes
  from a TextSection’s `body_json`
* Infer text section node connections for any given annotation with
  the AnnotationNode view
* If nodes get orphaned / moved, try to reconstruct the node from
  the previously extrapolated nodes belonging to another `body_hash`.


1aurend:
Included in #3600

PR #3569, opened 2023-03-09
[F] Render annotation detail from nodes

1aurend:
This PR updates how annotation subjects are rendered both in the reader notes drawer and in full lists of annotations, such as the My Notes overlay or in a reading group. Existing highlights and annotations are rendered using the new `annotation_node` property. Pending annotations are rendered directly from `body_json`. As a result, annotations are rendered with at least a sentence of surrounding text context, and we can render MathML, thumbnails of images, etc. in the notes drawer. I've also implemented an element blacklist that can be used to prevent elements from showing up in the drawer. Right now, the only blacklisted element is `iframe`, so that should be updated before this is merged.


1aurend:
Includes commits from #3559.

PR #3581, opened 2023-04-06
[F] Support text section ingest

1aurend:


api/app/services/ingestions/text_section/updater.rb
```diff
@@ -0,0 +1,83 @@
...
+        text.ingestion_sources.find_by(source_identifier: attributes[:source_identifier])
+      end
+
+      def user_stylesheets
+        text_section.stylesheets.where(ingested: false)
+      end
+
+      def stylesheets
+        text.stylesheets.where(hashed_content: attributes[:stylesheet_contents])
+      end
```
1aurend, line 70: *change from `text` to `text_section`

1aurend: Nvm— I was wrong about this.

api/app/enums/ingestion_target_kind.rb
```diff
@@ -0,0 +1,77 @@
...
+class IngestionTargetKind::Text < IngestionTargetKind
+end
+
+class IngestionTargetKind::TextSection < IngestionTargetKind
+  def reingest?(ingestion)
+    ingestion.text_section.present?
+  end
+
+  # rubocop:disable Metrics/MethodLength
+  def begin_processing(user, ingestion)
```
zdavis, line 47: Is there any difference between the two `begin_processing` methods in these classes other than the fact that one sets `ingestion.text` and the other sets `ingestion.text_section`?

If not, I would remove the redundant code in here by doing something like this:

```
class IngestionTargetKind < ClassyEnum::Base
  [...]
  # rubocop:disable Metrics/MethodLength
  def begin_processing(user, ingestion)
    serialization_options = { current_user: user }
    serialization = V1::IngestionSerializer.new(
      ingestion,
      serialization_options
    ).serializable_hash
    IngestionChannel.broadcast_to ingestion, type: "entity", payload: serialization

    outcome = nil

    begin
      PaperTrail.request(whodunnit: user) do
        outcome = Ingestions::Ingestor.run ingestion: ingestion
      end
      outcome
    rescue StandardError => e
      return ingestion.handle_ingestion_exception(e)
    end

    if outcome.valid?
      ingestion.info("\nIngestion Complete.")
      attach_result(ingestion, outcome.result)
      ingestion.processing_success
    else
      ingestion.handle_ingestion_exception(outcome.errors)
    end
  end
  # rubocop:enable Metrics/MethodLength

  def attach_result(ingestion, result)
    raise NoMethodError("IngestionTargetKind enum must implement attach_result")
  end

  [...]

end

class IngestionTargetKind::Text < IngestionTargetKind

  def attach_result(ingestion, result)
    ingestion.text = result
  end

end

class IngestionTargetKind::TextSection < IngestionTargetKind

  def attach_result(ingestion, result)
    ingestion.text_section = result
  end

end

zdavis: (I might also use this as an opportunity to break this long method into multiple methods to avoid having to disable the MethodLength lint rule.)

PR #3597, opened 2023-05-24
[F] Support asset management

1aurend:


api/app/uploaders/concerns/shared_uploader.rb
```diff
@@ -10,8 +10,10 @@ module SharedUploader
     host: MANIFOLD_CONFIG.api_url&.sub(%r{/\z}, "") || ""
   }.freeze
 
-  BETTER_MARCEL = ->(io, *_) do
-    Marcel::MimeType.for(io, name: io.path, extension: File.extname(io))
+  BETTER_MARCEL = ->(io, analyzers) do
+    return Marcel::MimeType.for(io, name: io.path, extension: File.extname(io)) if io.respond_to? :path
+
+    analyzers[:marcel].call(io)
   end
```
1aurend, line 17: Change that needs a deeper look @zdavis

client/src/backend/components/authoring/AddEditAssetForm/index.js
```diff
@@ -0,0 +1,134 @@
...
+        label={t("texts.assets.upload_label")}
+        instructions={t("texts.assets.upload_instructions")}
+      >
+        <Form.Upload
+          fileNameFrom="attributes[attachmentData][metadata][filename]"
+          value={
+            asset
+              ? `http://manifold.lvh/system/${asset?.attributes?.attachmentData.id}`
+              : undefined
+          }
```
1aurend, line 116: Fix this, but also, why didn't/couldn't I use the api proxy here?

zdavis:
@scryptmouse please review the ruby parts of this when you get a chance.

api/app/uploaders/concerns/shared_uploader.rb
scryptmouse: Is this basically because some things might get uploaded that don't have a `#path` method, and so we need to defer to the default marcel analyzer for those?

Regardless, rather than using a guard clause here, I'd just use a standard if/else block. It's easier to understand.

```ruby
if io.respond_to?(:path)
  Marcel::MimeType.for(io, name: io.path, extension: File.extname(io))
else
  analyzers[:marcel].call(io)
end
```

scryptmouse (changes requested):
The controller stuff looks fine to me, just that one Q and slight rewrite for the proc.

api/app/uploaders/concerns/shared_uploader.rb
1aurend: I'm using the updaters to get the file data for the IngestionSource. The type of data object that gets created via this method is different, I think, than what gets created when a source file is read during ingestion, and so doesn't have the `path` method. Everything I uploaded via the api lacked the path method, as far as I could.

```
module Updaters
  # Updates an IngestionSource model from JSON-API style params
  class IngestionSource
    include ::Updaters

    def attachment_fields
      [:attachment]
    end

  end
end
```

1aurend: But then because of the model it uses `IngestionSourceUploader` rather than `AttachementUploader`. Is this a problem?

scryptmouse: No, that's fine. If it doesn't have path, it doesn't have path. I'd just restructure the proc to use the above if/else construct then so it's more clear that it's branching on `respond_to?(:path)`

scryptmouse (approved):

PR #3620, opened 2023-07-06
[F] Add css layers

1aurend:
Resolves MNFLD-544; also, updates `font-display` to `swap`


1aurend:
@dananjohnson there's not much to see here, but maybe we can talk about this one next week. I tried a few different ways of adding the layers, and this very simple approach seemed to work best.

PR #3629, opened 2023-08-09
[F] Add alt text fields for uploaded images

### LINKED ISSUE #2323: Add alt text fields for project cover, background, and thumbnail images.

1aurend:
Resolves #2323


1aurend:
Resolves #2323

api/app/uploaders/attachment_uploader.rb
```diff
@@ -49,6 +50,10 @@ class AttachmentUploader < Shrine
     outcome
   end
 
+  add_metadata :alt_text do |io, _context|
+    io.respond_to?(:alt_text) ? io.alt_text : nil
```
scryptmouse, line 54: it's a very minor thing, but you could do `io.try(:alt_text)` to do this with the same result, but more succinct and idiomatic rails

scryptmouse: Also while we're not currently using it, I wanted to go ahead and check to make sure that if we end up adding `refresh_metadata` plugin, it won't break anything. Because `add_metadata` defines a reader on the UploadedFile, and `io` when using refresh metadata is the `UploadedFile` instance, it will just transparently work with the way the block is written. For new uploads, `io.alt_text` will be reading from the data file. For existing uploads, when refreshing, it will be the previous alt text value and we won't lose it when refreshing.

api/app/lib/updaters.rb
```diff
@@ -72,35 +72,48 @@ def assign_attributes!(model)
...
     end
   rescue ::Shrine::Plugins::DataUri::ParseError
     nil
   end
 
+  def update_attachment_alt_text(alt_text, key, model)
+    attachment = model.send(key)
+    return unless attachment.present?
+
+    attachment.metadata[:alt_text] = alt_text
```
scryptmouse, line 112: While I think this will work, Shrine's intended API is to use `add_metadata`. so we should instead do:

```ruby
attachment = model.send(key)

return unless attachment.present?

attachment.add_metadata alt_text: alt_text

attachment
```

api/app/serializers/v1/concerns/project_serializer.rb
```diff
@@ -24,6 +24,9 @@ module ProjectSerializer
         typed_attribute :recently_updated, Types::Bool.meta(read_only: true), &:recently_updated?
         typed_attribute :updated, Types::Bool.meta(read_only: true), &:updated?
         typed_attribute :avatar_styles, Types::Serializer::Attachment.meta(read_only: true)
+        typed_attribute :avatar_alt_text, Types::String.optional do |object, _params|
+          object.avatar.metadata["alt_text"] unless object.avatar.nil?
```
scryptmouse, line 28: When you find yourself writing method calls like this that involve a lot of deep navigation, that's usually a sign you want to use `delegate`.

Shrine also helps us here, because using `add_metadata` will automatically define a reader on the `UploadedFile` instance.

in `app/models/concerns/attachments.rb`, in the `manifold_has_attached_file` method, after `include AttachmentUploader::Attachment.new(field)`, add the following line:

```ruby
delegate :alt_text, to: field, prefix: true, allow_nil: true
```

That will create methods like `hero_alt_text`, `avatar_alt_text` etc that already safely navigate null attachments because of the `allow_nil: true` option. You can then remove all these blocks from this and all the similar ones. It will just read the method directly on the model since it's the same name.

scryptmouse (changes requested):
This looks great! Few minor niggles WRT Shrine's add_metadata API and using delegate, but I think it's GTG after that.

I looked over the FE stuff and it looked good, too. (Plus you know that way better than I anyway)

1aurend:
> 

I'm glad the FE part made sense to you. I went back and forth on that a few times trying to figure out a good way to make the UI reflect the fact that the alt text is being stored on the attachment and not really as a separate field.

api/app/uploaders/attachment_uploader.rb
1aurend: Just making sure I'm clear, you already checked `refresh_metadata` and we're good? Fwiw, I think I came across a related issue making `IngestionSources` uploadable from the client: https://github.com/ManifoldScholar/manifold/pull/3597/files#r1237821815

api/app/lib/updaters.rb
1aurend: Ah, didn't realize the plugin made a callable method, but that makes sense!

api/app/uploaders/attachment_uploader.rb
scryptmouse: Yeah, I pulled your branch down and tested adding refresh_metadata just to make sure the alt text was persistent. Not that we need to add it now, but I wasn't 100% on how shrine would behave with ephemeral metadata like this.

api/app/serializers/v1/concerns/project_serializer.rb
1aurend: Nifty!

api/app/lib/updaters.rb
scryptmouse: Yeah, Shrine is really powerful and handy: [add_metadata](https://shrinerb.com/docs/plugins/add_metadata)

scryptmouse (approved):

PR #3631, opened 2023-08-10
[F] Add endpoint for triggering ingestions

### LINKED ISSUE #3114: Allow Ingestion via REST API with a simple one-step interface
Following the discussion at https://manifold-community.slack.com/archives/C8LFMHM0X/p1636035666015300 it became clear that ingestions via Swagger RESTful API is currently not possible. The API does not trigger the necessary events in the "state machine of the ingestion model". This capacity would need to be added to the RESTful API

Since ingestion via RESTful API turns into a rather complex process (POSTing against the `api/files` endpoint, POSTing agains the ingestion endpoint ...) it might be helpful to have these steps merged behind 1 endpoint method.

1aurend:
Resolves #3114


1aurend:
@scryptmouse no rush on reviewing this one; it's a standalone thing. Some questions I have:
- Does it make sense to handle both text and text section ingests in one controller action like this?
- What should this endpoint actually return since it's going to be used for running ingestions outside of the client?
- Is it right to use `current_user` as creator and for auth here?

api/app/controllers/api/v1/ingests_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+      before_action :set_project, only: [:create]
+      before_action :set_text, only: [:create]
+
+      resourceful! Ingestion do
+        @project&.ingestions
+        @text&.ingestions
+      end
+
+      # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+      def create
```
scryptmouse, line 14: To answer your larger question, I would make this a separate controller. By all means, you could make a concern to DRY things up, but the controller should only handle one kind of ingestion parent. You had basically something perfect, I just broke it out into a gist because typing it up in these little windows isn't a good time:

https://gist.github.com/scryptmouse/75d4d1cbd8f487bfa68a2594631e2177

api/app/controllers/api/v1/ingests_controller.rb
```diff
@@ -0,0 +1,53 @@
+module API
+  module V1
+    # Trigger ingestion processing via the API
+    class IngestsController < ApplicationController
+      before_action :set_project, only: [:create]
+      before_action :set_text, only: [:create]
+
+      resourceful! Ingestion do
+        @project&.ingestions
+        @text&.ingestions
```
scryptmouse, line 10: See below for why one-size-fits-all is not an ideal approach, but this also won't necessarily do what you think: ruby has implicit return for the last statement in a lexical block, so if this were called for a project, the block would still return nil, because `@text&.ingestions` is the last statement, and it'd be nil in that case.

If this were all to be the same controller, you'd want `@project&.ingestions || @text&.ingestions`.

api/app/controllers/api/v1/ingests_controller.rb
scryptmouse: The biggest reason is the SRP (single responsibility principle). Even though the concern is almost the same code, the difference is you're specifying that for `api/v1/projects/ingestions_controller.rb` it's handling project ingestions, and for `api/v1/texts/ingestions_controller.rb` it's handling text ingestions.

api/app/controllers/api/v1/ingests_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+      end
+
+      # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+      def create
+        parent = @project || @text
+        @ingestion = ::Updaters::Ingestion
+          .new(ingestion_params)
+          .update_without_save(parent.ingestions.new)
+
+        @ingestion.creator = current_user
```
scryptmouse, line 20: Using `current_user` here and below is totally fine and fits with how things are done elsewhere in Manifold. This comes in on the API and the user who triggered the ingestion is the one who should be its creator / processor.

api/app/controllers/api/v1/ingests_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+        else
+          render_single_resource @ingestion, serializer: ::V1::IngestionSerializer
+        end
+      end
+      # rubocop:enable Metrics/AbcSize, Metrics/MethodLength
+
+      private
+
+      def set_project
+        @project = Project.friendly.find(params[:id], allow_nil: true)
```
scryptmouse, line 45: Breaking things out into multiple controllers also allows `Model.find` to raise `ActiveRecord::RecordNotFound` where it should. As is, if someone tried to trigger an ingestion through the API on a record that had been deleted, or had a typo, or any number of other things, you could get into a place where `@project` and `@text` are both nil, and the approach here doesn't really know how to deal with that.

Splitting it out in a concern with an interface that says which type it handles allows us to raise the not found exception and process things as we do in other APIs.

api/spec/requests/ingests_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  context "when a text ingest is triggered" do
+    let!(:path) { "/api/v1/projects/#{project_id}/ingest" }
+
+    before(:each) do
+      post path, headers: admin_headers, params: valid_params
+    end
+
+    it "creates and processes an ingestion" do
+      expect(response).to have_http_status(201)
```
scryptmouse, line 28: This is a strong personal preference, and I know there are other Manifold tests written like this so I get that it might seem ideal, but I strongly believe it can be a little confounding to have requests in specs happen in before blocks. The side effects should happen within a single example.

```ruby
it "creates an processes an ingestion" do
  expect do
    post path, headers: admin_headers, params: valid_params
  end.to change(Ingestion, :count).by(1)

  expect(response).to have_http_status(:created)
end
```

Rack / Rails also defines nice, human-readable symbols for status codes, so you can pass in things like `:ok` (200), `:created` (201), `:no_content` (204), `:bad_request` (400), `:unauthenticated` (401), `:forbidden` (403), `:not_found` (404), `:gone` (410), `:unprocessable_entity` (422), `:internal_server_error` (500).

You can still use the numbers if you prefer, I just find the symbols a little more readable.

api/spec/requests/ingests_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+  end
+
+  let!(:text) { FactoryBot.create(:text) }
```
scryptmouse, line 32: This should go in the below context. Otherwise you're creating a text for the above test as well, which doesn't need it.

Likewise, the `let!(:project) { ... }` bit should go in its associated context. `valid_params` and `attributes` are shared between the two types, but text and project are specific to each nested example.

scryptmouse: Though also, if things are broken out into their own bespoke controllers, you should probably wind up with two very similar specs:

* spec/requests/api/v1/projects/ingestions_spec.rb
* spec/requests/api/v1/texts/ingestions_spec.rb

api/app/controllers/api/v1/ingests_controller.rb
```diff
@@ -0,0 +1,53 @@
+module API
+  module V1
+    # Trigger ingestion processing via the API
+    class IngestsController < ApplicationController
```
scryptmouse, line 4: Because this creates an `Ingestion` model, the controller should be called `IngestionsController` to be proper and restful

api/config/routes.rb
```diff
@@ -187,6 +187,9 @@
         end
       end
 
+      post "projects/:id/ingest", to: "ingests#create"
+      post "texts/:id/ingest", to: "ingests#create"
```
scryptmouse, line 191: Breaking these out, you're gonna end up with routes that look like this instead:

```ruby
namespace :projects do
  resources :ingestions, only: %i[create]
end

namespace :texts do
  resources :ingestions, only: %i[create]
end
```

api/spec/requests/ingests_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      source: markdown_source_params,
+      ingestionType: "epub"
+    }
+  }
+  let!(:valid_params) {
+    build_json_payload(attributes: attributes)
+  }
+
+  context "when a text ingest is triggered" do
+    let!(:path) { "/api/v1/projects/#{project_id}/ingest" }
```
scryptmouse, line 21: We shouldn't build our paths raw in request specs: `api_v1_projects_ingestions_path(project)` will get you a path.

scryptmouse (changes requested):
Because this was a smaller thing I went a little more in-depth on the idiomatic part of the review and tried to explain why. I hope it is helpful but please let me know if there are any questions!

scryptmouse:
I forgot to answer your question about what it should return. Honestly, that's up to you. We might want to return the ingestion itself so the client can maybe track the ID and use it to poll the API for changes.

scryptmouse:
Made a quick update on the linked gist as well—forgot to account for the parent id key changing from `params[:id]` to `:project_id`, `:text_id`, etc.

zdavis:
@scryptmouse thank you for this awesome review!

1aurend:
Yes, thank you!!

scryptmouse (approved):

scryptmouse:
I think this looks great now, and should we end up adding another ingestable thing, it'll give us more freedom to handle it however we need to in its own controller.

PR #3633, opened 2023-08-15
[F] Render annotations as links

### LINKED ISSUE #3533: Assign URL to view of page with annotation open
## Feature Description
This relates to feedback from the Mellon-funded Embedding Preservability project - it was suggested we log our ideas here for further conversation/consideration. Currently, when you click to open an annotation the URL does not change - this means there would be no way to share a link to an annotation. Also, because annotations appear as a result of an onclick() event on a `<span>` tag, this makes it more challenging for web crawlers to discover annotation content, which also impacts the ease with which this content is crawled/archived. 

Following the pattern used to open a resource modal, replacing the `<span>` used for annotations with something like the following would allow the annotation view of a page to be bookmarked and simplify crawling:
`<a href="https://manifold.umn.edu/read/metagaming/section/00d65eb7-926c-4159-9d95-5fee5285e7b8/annotation/abcdef-1234-5678-ghij-abc123def456"... >{highlighted text}`

Allows users to bookmark and share annotations. Enables crawlers to discover annotations for indexing and archiving.

## User Stories
As a preservation archive I would like to be able to auto-crawl and archive Manifold publications, including annotation content when in scope for preservation, without having to create a customized crawl process to support content that is only visible via an onclick() event.
As a user of Manifold I would like to be able to bookmark and share annotation links.

## Design Notes
Discovery of href URLs in a-links is broadly supported by web crawlers but firing onclick() events to discover content is less frequently supported or requires custom solutions. Using `<a href` supports fully automated crawling, improving the likelihood that archives can get a high-fidelity copy of Manifold publications for preservation.

## Development Notes
N/A

1aurend:
Resolves #3533

### DIFF
--- a/client/src/reader/components/annotation/popup/menus/FollowLink.js
+++ b/client/src/reader/components/annotation/popup/menus/FollowLink.js
@@ -3,6 +3,7 @@ import { useTranslation } from "react-i18next";
 import PropTypes from "prop-types";
 import Menu from "../parts/Menu";
 import MenuItem from "../parts/MenuItem";
+import { useHistory } from "react-router-dom";
 
 function FollowLink({
   menu,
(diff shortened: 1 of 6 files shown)


1aurend:
@dananjohnson could you take a look and see if you see any potential accessibility issues here? Or any other reason not to make this change?

PR #3634, opened 2023-08-16
[B] Ensure default restricted access notice content

1aurend:


api/app/models/settings.rb
```diff
@@ -17,6 +17,8 @@ class Settings < ApplicationRecord
       social_share_message: "Shared from Manifold Scholarship",
       library_disabled: false,
       restricted_access: false,
+      restricted_access_heading: "Access to this project is restricted.",
+      restricted_access_body: "Only users granted permission may view this project's texts, resources, and other content.",
```
scryptmouse, line 21: I would enhance this model in a couple small ways. We should define these strings as constants, and add accessors for them for use in the project.

```
class Settings < ApplicationRecord
  DEFAULT_RESTRICTED_ACCESS_HEADING = "Access to this project is restricted.".freeze
  DEFAULT_RESTRICTED_ACCESS_BODY = ""Only users granted permission may view this project's texts, resources, and other content.".freeze

  # Add these constants in the defaults hash where you have the strings, then, later in the class:

  # @return [String]
  def default_restricted_access_heading
    general.fetch(:restricted_access_heading, DEFAULT_RESTRICTED_ACCESS_HEADING)
  end

  # @return [String]
  def default_restricted_access_body
    general.fetch(:restricted_access_body, DEFAULT_RESTRICTED_ACCESS_BODY)
  end
end

File: api/app/models/project.rb [hunk 2]
```diff
@@ -124,6 +124,7 @@ class Project < ApplicationRecord
   before_create :assign_publisher_defaults!
   after_commit :trigger_creation_event, on: [:create]
   after_commit :queue_reindex_children_job
+  before_validation :ensure_restricted_access_notice_content
```
scryptmouse, line 127: Minor things, but I really think having ! at the end of hooks is a helpful idiom to follow. Not saying change them where you see them in existing code, but it's a pattern we follow in newer Rails apps and any new hooks should do it.

Additionally, it isn't a problem here, but in later versions of rubocop-rails, `before_validation` hooks should come before `before_create` / `before_update` (and certainly before `after_commit`). This helps developers reading the hooks follow the logical path of execution for callbacks, so you can read something like:

```
before_validation :ensure_restricted_access_notice_content!
before_create :assign_publisher_defaults!
before_update :prepare_to_reindex_children, if: :draft_changed?
after_commit :trigger_creation_event, on: [:create]
after_commit :queue_reindex_children_job
```
... and be able to see each step more or less in the order it executes.

api/app/models/project.rb
```diff
@@ -416,6 +417,16 @@ def trigger_creation_event
     Event.trigger(EventType[:project_created], self)
   end
 
+  def ensure_restricted_access_notice_content
```
scryptmouse, line 420: Since we're making sure we have easy access to the default strings from settings, we can use those rather than repeating hard-coded strings in code. Otherwise, if we ever want to mess with this wording, we have to update it in two places.

```ruby
def ensure_restricted_access_notice_content!
  return unless restricted_access?

  settings = Settings.instance
    
  self.restricted_access_heading = settings.default_restricted_access_heading unless restricted_access_heading?
  self.restricted_access_body = settings.default_restricted_access_body unless restricted_access_body?
end
```

api/app/models/project.rb
scryptmouse: In addition, it feels like we should probably add presence validation for these fields just to make sure nothing is broken:

```ruby
validates :restricted_access_heading, :restricted_access_body, presence: { if: :restricted_access? }
```

That'll go down around line 138 with the other validations.

scryptmouse (changes requested):
Looking good! Some minor idiomatic fixes and DRY-ing up.

scryptmouse (approved):

Issue #3590, opened 2023-05-03
Feature request: allow units in resource sizes


benmiller314:
In the settings for interactive resources, there are fields to support an optional minimum height and minimum width. These fields, however, seem to only accept numbers, which are then converted to a fixed number of pixels. I discovered this when trying to use a viewport unit, because declaring an absolute minimum number of pixels is a recipe for awkward overflow on screen sizes smaller than the one on which we're preparing the document. (Indeed, setting a minimum rather than a maximum width also runs the risk of the resource overflowing the available space, preventing the iframe from responsively scaling for mobile devices or side-by-side browser windows.) 

I've confirmed that [iframes can take viewport units for their own widths and heights](https://developer.mozilla.org/en-US/docs/Web/CSS/Viewport_concepts), though in that case width and height have to be set in the CSS rather than in the HTML itself. I hope this isn't too big of a barrier.

If the issue is instead more about the challenges of parsing input from the webform, would it be possible instead to add a pulldown menu on the form that allows us to specify units? This would also clarify to users that px is the default, if that's what you prefer.

Grateful for your consideration of this enhancement!

1aurend:
Resolved in #3636

PR #3636, opened 2023-08-17
[F] Interactive resource updates

### LINKED ISSUE #3130: interactive resources limit iframe functionality
interactive resource feature could expand functionality by creating a field where other  attributes could be added (e.g. `allow=camera` for interactive / camera based sketches).

1aurend:
Resolves #3130, #3590


api/db/migrate/20230817212021_add_iframe_allows_to_resources.rb
```diff
@@ -0,0 +1,11 @@
+class AddIframeAllowsToResources < ActiveRecord::Migration[6.1]
+  def up
+    add_column :resources, :iframe_allows, :string, array: true, null: false, default: ["fullscreen"]
+    remove_column :resources, :iframe_allow_fullscreen, :boolean
```
1aurend, line 4: The client doesn't use this field, so it seemed safe to incorporate into the new one.

scryptmouse: Always a fan of merging things. Since it isn't being it sounds like we don't need to worry about migrating any false values (since they shouldn't exist).

scryptmouse (approved):
This looks great to me.

PR #3639, opened 2023-08-22
[F] Improve packaging

1aurend:


api/app/models/resource.rb
```diff
@@ -247,6 +247,16 @@ def vimeo?
...
   end
 
+  # @return [{ Symbol => Object }]
+  def packaging_metadata
+    metadata.with_indifferent_access
+      .merge(id: id)
+      .merge(slug: slug)
+      .merge(title: title_plaintext)
+      .merge(description: description_plaintext)
+      .merge(caption: caption_plaintext)
```
scryptmouse, line 257: I'm not gonna say we need to change this here, but in future, you can just do `.merge(id: id, slug: slug, title: title_plaintext, description: description_plaintext, caption: caption_plaintext)`

scryptmouse (approved):
This looks good! One suggestion for the future but it's fine to leave as-is.

PR #3653, opened 2023-09-06
[F] Use query params in admin

1aurend:
This PR unifies how list pagination and filters are handled between the frontend and admin mode, using query params wherever a route's primary purpose is to display a single list. Pagination links are also changed to buttons rather than anchors where no valid href exists, e.g. on pages displaying multiple paginated lists.


1aurend:
@dananjohnson It looks to me like this is basically done, and the only conflicts are snapshots that can be deleted. I'd need to update the new spam mitigation list views to match this and do a little testing, but it seems worth getting in?

dananjohnson:
@1aurend that's great! Yeah I agree—it's worth merging and then utilizing in the new list views.

client/src/hooks/useListQueryParams/index.js
dananjohnson, line 1: Looks good!

1aurend:
@dananjohnson the project exports bug was related to this PR and was an easy fix. The PC drawer thing is an issue on Edge too, so I'm copying your notes into a separate Linear issue. I think it's an issue with how `onDirty` is being set on that form. It's one of the very old ones.

PR #3665, opened 2023-10-05
[F] Add option to hide sections from reader nav

1aurend:
Also resolves:
- MNFLD-852


api/app/serializers/v1/text_serializer.rb
```diff
@@ -134,7 +134,16 @@ def ingestion_external_source_url(object, params)
...
-        sections_ids.map { |id| Hash[id: id.to_s, name: object.text_sections.find(id).name] }
+
+        sections_ids.map do |id|
+          section = object.text_sections.find(id)
+
+          {
+            id: id.to_s,
+            name: section.name,
+            hidden: section.hidden_in_reader
+          }
```
scryptmouse, line 145: This whole method is better expressed as:

```ruby
object.text_sections.where(id: object.spine).map do |text_section|
  text_section.slice(:id, :name).merge(hidden: text_section.hidden_in_reader)
 end
```

api/db/migrate/20231005175407_add_hidden_in_reader_to_text_sections.rb
```diff
@@ -0,0 +1,5 @@
+class AddHiddenInReaderToTextSections < ActiveRecord::Migration[6.1]
+  def change
+    add_column :text_sections, :hidden_in_reader, :boolean, default: false
```
scryptmouse, line 3: This should have `null: false` as well as `default: false`. You can bring it down in your local DB by running:

```sh
bin/rake db:migrate:down VERSION=20231005175407
```

scryptmouse (changes requested):

api/app/serializers/v1/text_serializer.rb
scryptmouse: Technically I might add a method like:

```ruby
def to_section_map
  slice(:id, :name).merge(hidden: hidden_in_reader)
end
```

to `TextSection`, then you can just have:

```ruby
object.text_sections.where(id: object.spine).map(&:to_section_map)
```

scryptmouse (approved):

PR #3773, opened 2024-09-09
[E] Expose reader font-size adjustment variable

dananjohnson:


dananjohnson:
To test:
1. Add to a text's custom stylesheet: 
```css
.manifold-text-section {
    --font-size-adjustment: 0.625;
}
```
2. View that text in the reader and confirm that the font size is being adjusted accordingly. Change the user font size setting in the header and make sure the adjustment value continues to work.
3. Try an invalid value for the variable (a keyword, percentage, etc) and make sure it's ignored and the default reader font sizing is used.

1aurend (approved):
@dananjohnson I tried tests you suggested, and it seems like it's working to me. 

A couple of thoughts that you can take or leave:

1. Should we set an absolute min or max on font size? Depending on the adjustment value, the user size adjustments can allow font sizes that are really unreadable:
Screenshot 2024-09-16 at 11 05 48 AM

2. It seems like it takes a hard refresh to apply the font size adjustment. If you save a stylesheet, and then click the "view" link in admin, you don't see the adjustment without reloading the reader. Can/should we do anything about this?

dananjohnson:
@1aurend good idea! I'll update the CSS to clamp the `--font-size-adjustment` value. Re: #2, that's interesting… what part is missing when you click 'View'? Also, I wonder if it's a local dev issue only.

1aurend:
> @1aurend good idea! I'll update the CSS to clamp the `--font-size-adjustment` value. Re: #2, that's interesting… what part is missing when you click 'View'? Also, I wonder if it's a local dev issue only.

When I clicked view after changing the `--font-size-adjustment` value in the stylesheet and saving, I didn't see the reader font size change until after a hard refresh. Is that what you mean by "what part"? Could be a local issue. Are you able to reproduce?

dananjohnson:
Gotcha. Yeah, that's what I assumed you meant, just wanted to clarify that that value was missing. I've seen it apply/update immediately on edge, so I think we're probably good.

PR #3775, opened 2024-09-20
Italian translation

atomotic:
italian translation, provided by [Ledizioni](https://www.ledizioni.it/) publisher.


1aurend:
Translations moved into #3774

PR #3782, opened 2024-10-31
[F] Spam mitigation

1aurend:
This PR improves instance admins' ability to deal with spam by:
- Adding new searchable lists in admin for
  - Annotations
  - Comments
  - Reading Groups
  - Users
- Allowing admins to delete those records
- Updating flags functionality to make it clearer to users
- Surfacing flags and flagged records in admin and allowing admins to resolve flags


api/app/models/annotation.rb
```diff
@@ -357,5 +374,17 @@ def build_by_reading_group_membership_scope(rgm)
...
+      escaped = value.gsub("%", "\\%")
+
+      needle = "%#{escaped}%"
+
+      body_matches = where arel_table[:body].matches(needle)
+
+      creator_matches = joins(:creator).where(User.arel_table[:first_name].matches(needle).or(User.arel_table[:last_name].matches(needle)))
+
+      creator_matches.or(body_matches).distinct
+    end
```
1aurend, line 388: In case anyone looks at this while I'm out— this is the main piece of this work that I know needs to change. I'm fairly certain this way of setting up a keyword scope won't coexist with the pre-existing search. Annotations are indexed for search, but there are various conditions around visibility that exclude annotations from search that we don't want to apply in the admin view. In addition, for the admin list search, I don't think we need full text indexing of the annotation subject, etc. A simple keyword match on a couple of fields like creator name and body seems likely to get the results users are after.

1aurend:
@scryptmouse I know you've looked at a lot of this already, but I want to make sure to get your final 👍 before this gets merged. No rush, though, since Dana's also going to take a look when he's back.

scryptmouse (approved):
I think this looks great!

dananjohnson (approved):

PR #3794, opened 2024-12-19
[F] WebVTT support

1aurend:


scryptmouse:
I can't really say. If it's easy, we should support as many as makes sense. If it's hard to model or expose it in the UI, I think it depends on how likely it is folks are going to need multiple tracks.

dananjohnson:
Yeah. I think the known use case we have is for a single track. Support for multiple sounds good, but only if it's not too difficult to do. On the frontend, I guess we'd check for a track in the currently set language and fall back to "en"?

We should upgrade the package, but lets put a pin in it and revisit when we work on dependency upgrades next quarter.

1aurend:
Sound good re: package upgrade.

And probably makes sense to wait on multiple languages/tracks until we have a use case. The video player package displays a dropdown list of languages so the users can choose which to show, if it's supplied with more than one text track of `kind="captions"`. So, that part could be pretty trivial. As for modeling it, we could do something like what we did with `alt_text` to associate a `text_track_lang` and/or `text_track_kind` field with the relevant attachment. Anyhow... thoughts for if we get there.

client/src/backend/components/resource/form/kind/Video/Video.styles.js
dananjohnson, line 1: I think a lot of the styles here can be dropped. They override other styles that seem to work perfectly fine. I would keep the flexbox properties for the Group div, however, and replace the `inline-size` property on the child selector with `flex-basis: 350px`. This should yield something like this, which removes the wide gap between the children on desktop and makes the upload fields more consistent on mobile with how they look for other resource types.

Screenshot 2025-01-09 at 12 33 42 PM
Screenshot 2025-01-09 at 12 33 56 PM

client/src/frontend/components/resource-player/Video/index.js
```diff
@@ -104,23 +104,49 @@ class ResourcePlayerVideo extends Component {
...
-      attachmentStyles
+      attachmentStyles,
+      captionsTrackUrl
     } = this.props.resource.attributes;
 
+    // Use the proxy if running over ports in dev to avoid CORS error
+    const captionsSrc =
+      process.env.NODE_ENV === "development"
+        ? captionsTrackUrl.replace("3020", "3010")
+        : captionsTrackUrl;
```
dananjohnson, line 132: I know it's just for dev, but string replacing like this seems liable to alter some urls that we don't want to alter. Could you use the URL interface here and change the `port` property instead?

client/src/frontend/components/resource-player/Video/index.js
```diff
@@ -104,23 +104,49 @@ class ResourcePlayerVideo extends Component {
...
 
+  trackRef = el => {
+    if (!el) return;
+
+    el.addEventListener("load", () => {
+      if (!el.track?.cues) return;
+      Array.from(el.track.cues).forEach(c => {
+        // eslint-disable-next-line no-param-reassign
+        c.line = -3;
+      });
```
dananjohnson, line 115: What is this for? Might be good to add a comment for future developers.

dananjohnson (changes requested):
Working well! Just noted a couple minor things. The FE part of this is good to go once addressed, as far as I'm concerned.

client/src/frontend/components/resource-player/Video/index.js
1aurend: Is your question rhetorical? If it isn't, that moves the captions up from their default position, so the controls overlay doesn't cover them. I didn't know before I looked into this that they can't be positioned with css.

dananjohnson: It is! I mean, I didn't know that, but I don't think others will either, so a brief comment in your code would be good.

1aurend: Yeah, kind of interesting. `::cue` accepts only a limited set of css properties: https://developer.mozilla.org/en-US/docs/Web/CSS/::cue

client/src/backend/components/resource/form/kind/Video/Video.styles.js
1aurend: I agree— I don't love what I did with the styles here, but issue I was trying to solve was this. When the uploaded files have different length filenames the inputs can be different heights:
Screenshot 2025-01-13 at 9 56 50 AM
I replaced `inline-size` with `flex-basis`. That seems cleaner, but what do you think about leaving the other child styles to address the differing heights?

dananjohnson: Makes sense! That's fine.

1aurend:
@dananjohnson can you unblock merging whenever you get a moment? Thanks!

dananjohnson (approved):

PR #3801, opened 2025-01-15
Ts/add collaborator attributions

timbot1789:


api/app/models/collaborator.rb
```diff
@@ -25,4 +26,8 @@ def packaging_metadata
   def to_s
     "#{role} #{maker}"
   end
+
+  def set_role_priority
+    self.priority = CollaboratorRole.find_index(role)
```
scryptmouse, line 31: I don't think this should be using `find_index`—on the call I'd mentioned `ActiveSupport::Configurable` as the ideal way to define priorities. Relying on the index may seem like less effort, but it will make any future updates a lot more work and hard to reason about. Folks are absolutely gonna want to change the prioritization of some of these roles, add new ones that may be higher priorities than other ones, etc.

Relying on the class definition order means:
1. the priority is densely ranked: adding something at say, index 9 is going to require writing a migration that shifts a dozen+ other enum values instead of filling in a gap
2. the priority is implicit instead of explicit: figuring out which index something currently has relies on either counting the order in the file or executing code in the console instead of just a simple declaration in the enum classes.

In Manifold, all classy enums [already have ActiveSupport::Configurable](https://github.com/ManifoldScholar/manifold/blob/8536201dd4990d3f66945111cc58fa4a5a403a38/api/lib/patches/better_enums.rb#L8) included as part of a patch to the library, so it's really just a matter of adding the the priorities to the enum classes.

`integer` in PG gives us around -2 billion to 2 billion to work with, so a sparse priority instead of dense also lets us put things in between other enums if we need to rearrange stuff without having to radically restructure things. These values aren't visible in any meaningful way to users, so we don't have to worry about them being consecutive like 1, 2, 3, 4.

```ruby
class CollaboratorRole < ClassyEnum::Base
  config.priority = 1_000_000 # default at the bottom

  config_accessor :priority, instance_writer: false

  def gepub_add_method
    :add_contributor
  end
end

class CollaboratorRole::Author < CollaboratorRole
  config.priority = 1000

  def gepub_add_method
    :add_creator
  end
end

class CollaboratorRole::Editor < CollaboratorRole
  config.priority = 2000
end

# and so on
```

Then for `set_role_priority` in `Collaborator`, it becomes `self.priority = role.priority`.

api/db/migrate/20250115212958_add_other_description_priority_importance_to_collaborator.rb
```diff
@@ -0,0 +1,7 @@
+class AddOtherDescriptionPriorityImportanceToCollaborator < ActiveRecord::Migration[6.1]
+  def change
+    add_column :collaborators, :other_description, :string
+    add_column :collaborators, :priority, :integer
+    add_column :collaborators, :importance, :integer
+  end
```
scryptmouse, line 6: These new fields should be indexed, I'd make a composite for the order:

```ruby
change_table :collaborators do |t|
  t.index %i[collaboratable_id collaboratable_type priority role importance], name: "index_collaborators_sort_order"
end
```

This will optimize for cases like the view, so we can do `WHERE collaboratable_id = '1234' AND collaboratable_type = 'Text' ORDER BY priority ASC, role ASC importance ASC`

api/db/migrate/20250115214357_update_collaborator_role_values.rb
```diff
@@ -0,0 +1,14 @@
+class UpdateCollaboratorRoleValues < ActiveRecord::Migration[6.1]
+  def up
+    execute <<~SQL
+      UPDATE public.collaborators
+      SET role = 'author', priority = 0
+      WHERE role = 'creator';
+    SQL
+    execute <<~SQL
+      UPDATE public.collaborators
+      SET role = 'other', other_description = 'contributor', priority = 20
```
scryptmouse, line 10: With the priority updates, just making a note here that this migration will need to be updated.

api/db/views/makers_with_collaborator_roles_v01.sql
```diff
@@ -0,0 +1,57 @@
...
+  public.makers.prefix AS prefix,
+  public.makers.suffix AS suffix,
+  public.makers.display_name AS display_name,
+  public.makers.first_name AS first_name,
+  public.makers.middle_name AS middle_name,
+  public.makers.last_name AS last_name,
+  public.makers.id AS maker_id
+FROM public.journals
+JOIN public.collaborators ON public.collaborators.collaboratable_id = public.journals.id
+JOIN public.makers ON public.collaborators.maker_id = public.makers.id
```
scryptmouse, line 57: Joining against the collaboratable tables themselves, as well as the union, is definitely overkill here. The point of this view is to be fetched by collaboratable records, and there's no benefit to joining from them. In fact, no joins are necessary at all. This also isn't producing the ideal flattened structure (one row per maker/collaboratable combo) and is fetching maker information unnecessarily.

This query gives us all we need to restructure the serializer so it includes a priority-ordered list of roles and an association to the makers table so that we can return something FE can use:

```sql
SELECT
  (collaboratable_id::text || '-' || maker_id::text) AS id,
  collaboratable_id,
  collaboratable_type,
  maker_id,
  array_agg(DISTINCT role ORDER BY priority ASC) AS roles,
  MIN(priority) AS priority,
  MIN(importance) AS importance
  FROM collaborators
  WHERE maker_id IS NOT NULL -- not ideal, but the table currently has no null constraint on that field and fixing that is a separate issue
  GROUP BY collaboratable_id, collaboratable_type, maker_id
  ```

This view only exists to flatten the collaborator records between makers and collaboratables. The serializer should return the maker as a separate eager loaded association 

1. Joining against the collaboratable table is unnecessary because it requires using unions, requires duplicating those columns in each branch of the unions, and is a more expensive query. It's true that technically it might hide cases where there are orphaned collaborators, but that's not a concern the view needs to handle since it will effectively only be called from `collaboratable.makers_with_collaborator_roles.includes(:maker)`. It would also require us to have to update this view if a new collaboratable is added, which is unnecessary.
2. Joining against makers is likewise unnecessary since it means if we change the shape of maker we need to update this, when we could just eager load the association and keep letting maker manage its own shape.

File: api/app/models/makers_with_collaborator_role.rb [hunk 5]
```diff
@@ -0,0 +1,14 @@
+class MakersWithCollaboratorRole < ApplicationRecord
```
scryptmouse, line 1: I realize this has been used through a lot of this branch, but I think the name is a little confusing and awkward for Rails norms. The pluralization is a little hard to grok, and while it's associated with `maker`, calling it `maker` is a little misleading for its purpose.

I think something like `FlattenedCollaborator`, `CompositeCollaborator`, `AggregateCollaborator`, etc. would be much more clear about its purpose.

scryptmouse (changes requested):
Some necessary changes, I think, but it's looking good!

api/app/models/makers_with_collaborator_role.rb
```diff
@@ -0,0 +1,14 @@
...
+  include WithParsedName
+  belongs_to :maker
+  belongs_to :collaborator
+  belongs_to :commentable, polymorphic: true
```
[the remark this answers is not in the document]
timbot1789: whoops. I think I knew that and typed the wrong thing.

api/app/models/collaborator.rb
timbot1789: This seems reasonable. We used the `find_index` pattern on reliable credit, so I thought that was a common design here. This pattern seems better though

api/app/models/makers_with_collaborator_role.rb
timbot1789: I didn't like the name either, but couldn't think of a better one. And I don't like spending too much time thinking about naming. I'll change it to `FlattenedCollaborator`

api/app/models/collaborator.rb
scryptmouse: Yeah, I think `find_index` is fine for enums that are likely to never really change, but for a larger, more complicated enum that might go through a few revisions once it starts getting tested, we'll want this extra explicitness / flexibility.

api/app/models/makers_with_collaborator_role.rb
scryptmouse: it's definitely a pain to figure out (especially for models, since so much ends up being conventions based with them)

api/db/structure.sql
```diff
@@ -488,7 +488,10 @@ CREATE TABLE public.collaborators (
     updated_at timestamp without time zone NOT NULL,
     "position" integer,
     collaboratable_type character varying,
-    collaboratable_id uuid
+    collaboratable_id uuid,
+    other_description character varying,
+    priority integer,
+    importance integer
```
timbot1789, line 494: Ticket said to call this field 'position'. However, since 'position' is already taken in this record, I changed the name to 'importance'.

timbot1789:
@scryptmouse this should be ready for another pass

scryptmouse (approved):
this looks great!

PR #3802, opened 2025-01-22
[F] Add citation override field to entities

1aurend:


1aurend:
@dananjohnson there's not much code to review here, but I would appreciate your thoughts on whether we should tweak the presentation of any of these citations in the reader/frontend. So far this PR includes the minimum needed to add citation to existing metadata display. (I can add some more data, so the screenshots are more realistic if that helps.)

Text:
Screenshot 2025-01-22 at 3 51 40 PM
Text section:
Screenshot 2025-01-22 at 3 52 16 PM
Project/journal:
Screenshot 2025-01-22 at 3 52 54 PM
Resource:
Screenshot 2025-01-22 at 3 53 36 PM

dananjohnson:
@1aurend this looks good! My only suggestion (and feel free to disagree) might be to move the citation at the Text level above the copyright info, and remove the "Cite as:" text. I think it's self-explanatory without the label.

1aurend:
No, I agree this looks better (with/out a rights statement):
Screenshot 2025-01-28 at 1 26 31 PM
Screenshot 2025-01-28 at 1 29 05 PM

PR #3805, opened 2025-01-28
[F] Update UI for collaborators

1aurend:


1aurend:
@scryptmouse @dananjohnson — this PR review requires a bit of explanation: My initial plan for updating the collaborators UI was to implement a pattern similar to what we've recently done on other projects, where the admin list shows individual contributions. However, as I got into it, there were two key differences: (1) while the `Collaborator` model already existed, it was never exposed to the client (unless I really missed something). The client sent an array of maker ids for `creators` and `contributors`, which the updater handled. (2) `position` was managed at list-level via a DnD interface rather than as a field that users set on an individual collaborator record. So, after trying a few different setups for the UI, I landed on displaying flattened collaborators in admin as well as in the frontend. It seemed to be the smaller change for users, and I couldn't think of a way to handle position with multiple entries for a single maker in the list, short of changing how position is calculated, doing away with the DnD list, etc. While I like where this ended up, I'm happy to rethink if the endpoints I setup for creating and deleting collaborators by maker are problematic.

For context, here's a screen grab of the updated UI:
https://github.com/user-attachments/assets/ec5fedc1-10ee-4d0d-abf4-ae4fd0ae8dbc

dananjohnson (approved):
I thought the UI you came up with (as you demoed last week) was solid! I don't have any issues with proceeding with this approach.

scryptmouse (approved):
Looks great!

PR #3807, opened 2025-01-29
[F] Add markdown blocks to RG homepage

1aurend:
Also, fixes..
1. Missing focus styles for action buttons in dnd list
2. Keyboard move controls for collectables

### DIFF
--- a/api/app/controllers/api/v1/reading_groups/relationships/reading_group_categories_controller.rb
+++ b/api/app/controllers/api/v1/reading_groups/relationships/reading_group_categories_controller.rb
@@ -54,7 +54,7 @@ module API
           end
 
           def reading_group_category_params
-            attributes = [:title, :description, :position]
+            attributes = [:title, :description, :position, :markdown_only]
             relationships = [:projects, :resources, :resource_collections, :texts]
 
             param_config = structure_params(attributes: attributes, relationships: relationships)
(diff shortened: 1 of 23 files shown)


dananjohnson (changes requested):
This is great, Lauren! Thanks for fixing those bugs too. I have two small suggested changes:

- Can the title be optional on the markdown block? Or include a flag to hide it? It seems like it shouldn't be required.
- The markdown block should be visually distinct from the categories. How about just making the background transparent and removing the block padding?
Screenshot 2025-01-30 at 3 04 26 PM

1aurend:
Oh yeah, that looks good.

Re: optional title— it's currently a required property in the api. I can take a look at what side effects there might be of making it optional, and whether a flag would be easier. Perhaps another option could be to assign a title to the markdown blocks on creation, something like `markdown_{uid}`, always hide the title on markdown blocks, and then update our rte styles, so those block title styles could be replicated by putting an `h2` into the markdown body?

Current styles:
Screenshot 2025-01-30 at 3 44 09 PM

dananjohnson:
Yeah I like the generated title idea! Since it's a markdown block, they can just add headings if they need them. (And yes to add the RTE styles 👍🏻 )

1aurend:
@dananjohnson this should be ready to go. Do you want to give it another quick look and unblock merging (or just merge)?

One thing for later— h1 is allowed in markdown in Manifold. Those should probably be converted to h2 by the system. If we wanted to do that, it would either be an api change in the html rendering, or we could add a React markdown renderer and stop using dangerouslySetInnerHTML here.

dananjohnson (approved):

dananjohnson:
Go ahead and merge it. Let's set the h1 question aside for now. I'm not sure it's worth implementing the constraints on our end since it's largely up to content editors to use proper heading structure.

PR #3809, opened 2025-01-30
[E] Repurpose project_resource_editor to manage texts as well

scryptmouse:
* Additionally, correct discrepancy with marketeer management of Makers

Resolves MNFLD-949


scryptmouse:
@1aurend , I did a search and replace for `project_resource_editor` in the client, and also fixed up the i18n, but it could still use a frontend pass.

1aurend:
@scryptmouse is this ready to merge? Your change to entitlements looks good to me.

scryptmouse:
Should be!

PR #3824, opened 2025-03-10
[E] Implement consistent open graph content for entities

1aurend:


1aurend:
@dananjohnson this should be ready for some testing. Because of the differences between entity field names, there's still not a one-size-fits-all approach to generating og tag content. However, I think there are a few good improvements in here— the og fields appear consistently in admin with labeling for which fields are fallbacks; the entity head content logic calls out that by-entity fallback arrangement; and the new hooks make it easier to access the meta title for sharing in various places.

client/src/frontend/components/entity/useEntityHeadContent/index.js
```diff
@@ -0,0 +1,143 @@
...
+    descriptionPlaintext: parentDescriptionPlaintext
+  } = parent?.attributes ?? {};
+  return parentSocialDescription ?? parentDescriptionPlaintext;
+};
+
+const maybeCoverImage = (entity, fallback) => {
+  if (entity.attributes.coverStyles)
+    return entity.attributes.coverStyles.mediumSquare;
+  return fallback;
+};
```
dananjohnson, line 81: I'd expect this to often be a book or journal cover, in which case we probably don't want to crop it. Let's switch to the `medium` cover style instead.

client/src/frontend/components/entity/useEntityHeadContent/index.js
```diff
@@ -0,0 +1,143 @@
...
+      return maybeCoverImage(
+        entity,
+        maybeAvatarImage(entity, fallbackImageFor(parent))
+      );
+    case "journalVolumes":
+      return fallbackImageFor(parent);
+    default:
+      return null;
+  }
+};
```
dananjohnson, line 122: Very nice way of handling the different conditions here!

client/src/frontend/components/entity/useEntityHeadContent/index.js
```diff
@@ -0,0 +1,143 @@
...
+  if (entity.attributes.coverStyles)
+    return entity.attributes.coverStyles.mediumSquare;
+  return fallback;
+};
+
+const maybeAvatarImage = (entity, fallback) => {
+  if (entity.attributes.avatarStyles)
+    return entity.attributes.avatarStyles.mediumSquare;
+  return fallback;
+};
```
dananjohnson, line 87: Similarly here, let's not crop this.

dananjohnson (approved):
Great work! I noted a couple tiny things. Approving so that you can go ahead and merge after they're addressed.

PR #3826, opened 2025-03-18
[F] Implement PG search filtering

timbot1789:
This implements a separate Filtering applicator that supports using the `pg_search` gem rather than `searchkick`. It also modifies several models to support `pg_search` based filtering. Currently `pg_search` can be activated in an environment by setting an environment variable 'USE_PG_SEARCH' to true.

See the [apply_spec.rb](https://github.com/ManifoldScholar/manifold/pull/3826/files#diff-b30e9927ac51020b121370585eedf262f974a082a2467e8e348beeff4b58b298) for a list of currently supported models

#### Models that do not need to support `pg_search` because they already implement custom `by_keyword`:
- entitlement
- annotation
- reading group
- comment

#### Some Notes / Important differences from Searchkick:
- pg_search [can search through associations](https://github.com/Casecommons/pg_search?tab=readme-ov-file#searching-through-associations), but it's somewhat limited. It can't search more than one association deep, and it doesn't index the association. So we'll want to keep an eye on performance for some association-heavy searches.
- Since pg_search calls into Postgres, not Ruby, all field names passed to it need to be the postgres name for the corresponding column, not the ruby method name. So if we have a column `name` that we alias in rails with `alias name title`, we need to pass pg_search `name`, not `title`. This also means that fields like `title_plaintext` are unavailable. They could be implemented as a PG function, however.

#### Possible Improvements:
- I haven't tried to implement highlighting here, mostly because I haven't seen it in the front end in areas that seemed to be calling into keyword search. pg_search supports it though, so we can add it in another PR.
- I haven't intensely tested around typos, incomplete words, etc

### DIFF
--- a/api/app/models/maker.rb
+++ b/api/app/models/maker.rb
@@ -15,6 +15,7 @@ class Maker < ApplicationRecord
   include SerializedAbilitiesFor
   include WithParsedName
   include SearchIndexable
+  include HasKeywordSearch
 
   # Associations
   has_many :collaborators, dependent: :destroy
@@ -35,7 +36,7 @@ class Maker < ApplicationRecord
   validate :name_is_present!
 
   # Search
-  pg_search_scope :keyword_search, against: KEYWORD_SEARCH_ATTRIBUTES
+  has_keyword_search! against: KEYWORD_SEARCH_ATTRIBUTES
   searchkick(word_start: TYPEAHEAD_ATTRIBUTES,
              callbacks: :async,
              batch_size: 500)
--- a/api/app/models/maker.rb
+++ b/api/app/models/maker.rb
@@ -2,6 +2,7 @@
 class Maker < ApplicationRecord
   # Constants
   TYPEAHEAD_ATTRIBUTES = [:first_name, :last_name].freeze
+  KEYWORD_SEARCH_ATTRIBUTES = %i[name first_name middle_name last_name display_name].freeze
 
   PACKAGING_ATTRIBUTES = %i[id name first_name middle_name last_name display_name suffix prefix].freeze
 
@@ -34,6 +35,7 @@ class Maker < ApplicationRecord
   validate :name_is_present!
 
   # Search
+  pg_search_scope :keyword_search, against: KEYWORD_SEARCH_ATTRIBUTES
   searchkick(word_start: TYPEAHEAD_ATTRIBUTES,
              callbacks: :async,
              batch_size: 500)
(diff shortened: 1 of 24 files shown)


api/app/models/maker.rb
```diff
@@ -34,6 +35,7 @@ class Maker < ApplicationRecord
   validate :name_is_present!
 
   # Search
+  pg_search_scope :keyword_search, against: KEYWORD_SEARCH_ATTRIBUTES
```
scryptmouse, line 38: One thing I would suggest for this is building a concern to put a thin wrapper around this call rather than having it all be in every model. Given that this is something that will be applied on near every model, it's useful to be able to introspect. We may also want to provide some sort of normalizations in the future, in which case it will be nice to have it all centralized:

```ruby
module HasKeywordSearch
  extend ActiveSupport::Concern

  included do
    extend Dry::Core::ClassAttributes
    
    defines :keyword_search_options, type: ::Types::Hash.map(::Types::Symbol, ::Types::Any)
    
    keyword_search_options({}.freeze)
  end

  module ClassMethods
    # @return [void]
    def has_keyword_search!(**options)
      keyword_search_options options.freeze

      pg_search_scope :keyword_search, **options
    end

    def has_keyword_search?
      keyword_search_options.present?
    end
  end
end
```

Include that in ApplicationRecord and then it becomes easy to be like `Project.has_keyword_search?` and know if we're going to use that or some other strategy.

scryptmouse (changes requested):
Looking good! I think putting one abstraction in would be beneficial to build off of.

api/app/models/maker.rb
timbot1789: Oh ok. I was thinking of doing something like this, but was having trouble coming up with a clean way to do it. So to implement this on the model, it would look something like:

```
class Text < ApplicationRecord
  include HasKeywordSearch

  ...

  keyword_search_options {associated_against: {titles: [:value]}, against: [:description]}
end
```

That's pretty clean. I'll give that a go. I think I'll move the `include PgSearch::Model` into that concern as well, as the `pg_search_scope` line depends on that anyway.

scryptmouse: Ope, sorry, it would be `has_keyword_search! associated_against: {titles: [:value]}, against: [:description]`—the idea being we wrap the `pg_search_scope` call with that `has_keyword_search!` method. It's a much, much simpler version of the abstraction I'm using on SCV, etc.

scryptmouse: In the future, if we want to do some normalization of the fields, that'll give us one spot to do it instead of having to update the call to `pg_search_scope` in 20+ models.

timbot1789: Updated according to description

scryptmouse (approved):
this is continuing to look good!

PR #3832, opened 2025-04-10
PG Multisearch

timbot1789:
This PR converts the general search for manifold from Elasticsearch to Postgres.

To test, you will need to set the ENV variable `USE_PG_SEARCH=true`, otherwise the server will continue to use elasticsearch. You can do this by adding that line to the `docker/manifold.env` file.


scryptmouse:
Spent some time looking through this this evening, looking good! I want to spend a bit more time looking with fresh eyes but I think you're taking the right approach here with the serialization and abstractions.

scryptmouse:
Closing, changes now part of #3834

PR #3834, opened 2025-04-21
[F] Tie up PG Search Feature with Text Section Nodes

scryptmouse:
* Add ability to search text section nodes for excerpts to jump to
  when searching text sections
* Add some indexing and application-relevant associations to the
  pg_search multisearch index table to better reflect the shape of
  data that we have in Manifold
* Repurpose `search:reindex` task to rebuild pg_search multisearch
  data
* Disable certain elasticsearch tests that will soon to be removed
  entirely
* Update a number of tests that failed because associations are
  being read for search indexing earlier than they were previously


scryptmouse:
Now part of #3863

PR #3837, opened 2025-04-30
[B] Remove oauth from sign in forms

1aurend:


1aurend:
@dananjohnson this is the minimum needed to remove oauth buttons from the sign in screens. I left the components in case we need them again at some point, but let me know if you want me to remove the oauth button component, etc. altogether.

dananjohnson (approved):
Go ahead and leave them for now. 👍🏻

PR #3839, opened 2025-05-05
[B][E] Admin annotations fixes

1aurend:


dananjohnson:
These all look like API changes. Is a BE review needed?

dananjohnson:
Oh sorry! I was viewing a single commit, not all file changes. Let me take another look.

dananjohnson (changes requested):
Looks good; however, there's still an issue with the flag count. If I flag and then unflag my own annotation, then flag it again, I get `1 resolved flag` in the admin.

Screenshot 2025-05-05 at 2 19 53 PM

1aurend:
> Looks good; however, there's still an issue with the flag count. If I flag and then unflag my own annotation, then flag it again, I get `1 resolved flag` in the admin.
> Screenshot 2025-05-05 at 2 19 53 PM

I believe this is working as designed— a user flagging, unflagging, then reflagging an annotation doesn't trigger an unresolved flag to prevent users from spamming flag functionality. @scryptmouse can confirm. Given the flag data, there's no way to differentiate on the FE between a flag in this state and one resolved by an admin, hence you see that `1 resolved`:
Screenshot 2025-05-05 at 2 40 38 PM Were you thinking this should appear as an active flag?

dananjohnson:
Gotcha. What tripped me up was that when I resolved my own flag, I saw nothing under the "Flags" in the admin, but when I flagged it again, I saw "1 resolved flag". I was thinking there would continue to be nothing under "Flags".

Maybe we're overcomplicating this edge case situation. What if we just don't show the "Report" button on one's own annotations? Is there any reason that we would want it there?

1aurend:
Right— well, I hid the 1 resolved notification when it's resolved by the flagger, per our discussion in slack. Original behavior was to always show 1 resolved. I kind of wonder if we should revert that for consistency.

I'm onboard with not showing report on one's own annotations. That makes sense, but I don't see how it would address the issue above. Perhaps its not clear, but the creator for `resolvedByCreator` is the flag creator, not the annotation creator. The same thing happens in the flag display if you flag-unflag-flag someone else's annotation.

dananjohnson:
Oh ha thanks for clarifying. Yeah let's maybe revert it then. It's not perfect but at least it's consistent. What do you think?

1aurend:
@dananjohnson ready for a last look/approval.

@scryptmouse could I get a quick 👀 on this, since we're so close to release. in all cases when deleting an annotation, the FE hits `DELETE /api/v1/annotations/${annotationId}`, which wasn't previously a route. The api was falling back to the destroy action in `text_sections/relationships/annotations_controller`, which has some scopes applied, so I was seeing 404s when trying to delete annotations in an archived RG.

scryptmouse (approved):
The API changes make sense to me!

dananjohnson (approved):

PR #3857, opened 2025-05-21
[F] Support text tracks + transcripts for media resources

1aurend:


scryptmouse (approved):

1aurend:
@scryptmouse a question for you on this— what's the right way to handle removing the original captions_track from the first pass at caption support? I removed the related attributes from the resource serializer and params, since we're no longer using them, but I didn't know if it was safe to just remove the column from the table. The captions work has only been on Edge, so there's probably not much chance of data loss.

scryptmouse:
```bash
docker compose exec web bin/rails g migration RemoveOriginalCaptionsTrack
```

Open the file that was generated at `db/migrations/wholelottanumbers_remove_original_captions_track.rb`

Remove the `change` method and instead write two methods like:

```ruby
def up
  execute <<~SQL
  ALTER TABLE resources DROP COLUMN IF EXISTS captions_track_data;
  SQL
end

def down
  # Intentionally left blank
end
```

PR #3860, opened 2025-05-22
[E] Track share event on url copy

1aurend:


dananjohnson (commented):
If it's not too much trouble, what do you think about exposing a callback param on the copy hook and calling the track function in response to a successful write to clipboard? That way if copying fails for some reason, it's not tracked.

1aurend:
That's smart. Shouldn't be too hard to do.

1aurend:
Just an observation while updating this— it looks like we only display the share count in analytics for texts. Might make sense to add a cumulative count to the instance stats.

dananjohnson:
> Just an observation while updating this— it looks like we only display the share count in analytics for texts. Might make sense to add a cumulative count to the instance stats.

Maybe so. @tsmyre what you do think?

tsmyre:
I would be on board with that; makes all the sense.

1aurend:
It balances the list nicely too...

Screenshot 2025-05-22 at 2 31 18 PM

dananjohnson (approved):

PR #3869, opened 2025-05-28
[F] Add ingestion message and endpoints

timbot1789:


api/app/controllers/api/v1/ingestions_controller.rb
```diff
@@ -44,6 +54,32 @@ def update
...
+      def reingest
+        @ingestion = load_ingestion
+        Ingestions::ReingestIngestionJob.perform_later(@ingestion.id, current_user)
+      end
+
+      def show_messages
+        @ingestion = load_ingestion
+        permitted = params.permit(:starting_at)
+        time = DateTime.parse(permitted[:starting_at])
+        render_multiple_resources IngestionMessage.where(
```
scryptmouse, line 77: In addition to moving this to an `ingestions/relationships/ingestion_messages_controller.rb`, I think it would be best to just have FE retrieve all the messages at once (via `skip_pagination`) and let them decide if they need to show anything extra, rather than paginating by time.

api/app/jobs/ingestions/process_ingestion_job.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Ingestions
+  class ProcessIngestionJob < ApplicationJob
+    def perform(ingestion_id, user_id)
```
scryptmouse, line 5: These jobs should be using AJ serialization: pass in model instances, not ids.

api/app/controllers/api/v1/ingestions_controller.rb
```diff
@@ -44,6 +54,32 @@ def update
         )
       end
 
+      def reset
+        @ingestion = load_ingestion
+        @ingestion.reset
+        @ingestion.save
+      end
+
+      def process_ingestion
```
scryptmouse, line 63: As discussed in manifold slack, rename to just `process` for consistency with the other actions.

Each of these should return a serialized ingestion

scryptmouse: Additionally, in the ingestion serializer, it should provide links for `process`, `reingest`, and `reset`. See the examples in in the reading group / reading group membership serializers, but something like:

```ruby
    link :reset do |ingestion, _|
      ManifoldApi::Container["system.routes"].reset_api_v1_ingestion_path(ingestion)
    end
```

api/app/models/ingestion.rb
```diff
@@ -20,6 +20,7 @@ class Ingestion < ApplicationRecord
   belongs_to :text, optional: true
   belongs_to :text_section, optional: true
   belongs_to :project
+  has_many :ingestion_messages, -> { reorder(created_at: :asc) }, inverse_of: :ingestion, dependent: :destroy
```
scryptmouse, line 23: Small thing, but `dependent: :delete_all`—ingestions might have lots of messages and when we start pruning ingestions and stuff, it'll be good to clear these out quickly.

api/spec/requests/api/v1/ingestions_spec.rb
```diff
@@ -38,4 +39,64 @@
...
+    let!(:project_id) { ingestion.project_id }
+
+    before do
+      stub_request(:get, "http://example.com/index.md").
+        to_return(status: 200, body: "", headers: {})
+    end
+
+    it "can start ingestion processing" do
+      patch "/api/v1/ingestions/#{ingestion.id}/process_ingestion"
+      perform_enqueued_jobs
```
scryptmouse, line 54: As Zach suggested, make separate specs for testing if the jobs are doing what they're supposed to. Whenever testing side effects, especially with requests, use `expect do ...` syntax rather than relying on line by line things. Also, use Rails route helpers rather than constructing paths manually.

This should end up looking like:

```ruby
expect do
  post process_api_v1_ingestion_path(ingestion)
end.to have_enqueued_job(Ingestions::ProcessJob).once

expect(response).to be_successful
```

scryptmouse: In the job specs, you'd want to do something like:

```ruby
expect do
  described_class.perform_now ingestion, user
end.to change { ingestion.reload.finished? }.from(false).to(true)
```

api/spec/models/ingestion_message_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe IngestionMessage, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
scryptmouse, line 4: Go ahead and delete this file, we'll never need to test messages as a model

api/spec/factories/ingestion_messages.rb
```diff
@@ -0,0 +1,5 @@
+FactoryBot.define do
+  factory :ingestion_message do
+
```
scryptmouse, line 3: For testing the ingestion messages controller, we'll wanna create a few messages with some traits associated with the different types that can come back.

api/app/serializers/v1/ingestion_message_serializer.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module V1
+  class IngestionMessageSerializer < ManifoldSerializer
+    include ::V1::Concerns::ManifoldSerializer
+
+    typed_attribute :kind, Types::String.meta(read_only: true)
+    attribute :payload
```
scryptmouse, line 8: I'm pretty sure we can do `typed_attribute :payload, Types::Any.meta(read_only: true)` here. FE can repurpose the code they used for processing responses from ActionCable to know how to handle the `payload` value based on the `kind`.

Additionally, add `typed_attribute :id, Types::Serializer::ID` in the serialized message so FE has something to track which message(s) they've seen / processed to display to the user.

api/app/models/ingestion.rb
```diff
@@ -95,7 +96,7 @@ def add(severity, message = nil, _progname = nil)
     log_buffer << line
     return if severity == "DEBUG"
 
-    IngestionChannel.broadcast_to self, type: "log", payload: line
+    IngestionMessage.create(kind: "log", payload: line, ingestion: self)
```
scryptmouse, line 99: `ingestion_messages.create!(kind: "log", payload: line)`—use the association, and create with a bang. It should never reasonably fail, and if it ever does, it should be loud and let us know.

api/app/models/ingestion_message.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+# A message stored during resource ingestion to be later returned to the client
+class IngestionMessage < ApplicationRecord
+  belongs_to :ingestion
```
scryptmouse, line 5: This should have `validates :kind, :payload, presence: true` to reflect the DB constraints.

api/config/routes.rb
```diff
@@ -101,7 +101,14 @@
       resources :subjects
       resources :categories, except: [:create, :index]
       resources :makers
-      resources :ingestions, only: [:show, :update]
+      resources :ingestions, only: [:show, :update] do
+        member do
+          get "show_messages"
+          patch "reset"
```
scryptmouse, line 107: Use post, not patch to match the other similar actions we have that use these sorts of methods.

scryptmouse (changes requested):

api/app/controllers/api/v1/ingestions_controller.rb
timbot1789: Followup from slack: 
We can't name the function `process` because Rails reserves it (without telling anybody). We'll name it `do_process` instead, and map the `process` endpoint to it.
https://castiron.slack.com/archives/C040BGBE926/p1748627817765919

api/spec/factories/ingestion_messages.rb
timbot1789: Added some traits and a test for them

scryptmouse (changes requested):
This branch has the good_job database stuff in its api/db/structure.sql—can this get rebased so it doesn't include them and can go cleanly into master?

timbot1789:
@scryptmouse It took a little bit to get my DB properly reset, but `structure.sql` should be fixed now

PR #3878, opened 2025-06-07
[B][C] Few bug fixes

dananjohnson:
@1aurend take a look at the third commit in particular. This resolves that weird `Could not find "store"` error we were seeing. I traced the issue to a `confirm` prop that you were trying to pass through the `<CategoryNewToggle>` component you created to the existing `<NewCategory>` component. `<NewCategory>` is wrapped in the `withConfirmation` HOC, so the prop doesn't need to be passed down. And in `<CollectionEditor>`, where you first tried to pass it down, the value `confirm` didn't exist, meaning that `confirm` was interpreted as the [built-in browser function](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm). My guess is that the server-side error and blank screen are the result of calling a browser function in node. It's odd that the error surfaced as a Redux error, but that's my best guess.

If you want to test it, the reliable way I've found to reproduce the issue is on a fresh `yarn watch`. Once the error has been triggered once, the SSR server shuts down, so you won't see it again until you restart (and the client-side JS is fine with using `confirm`). So on a fresh `yarn watch`, you'll know it's working if you reload the RG edit route and don't see the error or a browser console warning that SSR is missing.

### DIFF
--- a/client/src/backend/containers/reading-group/Annotations.js
+++ b/client/src/backend/containers/reading-group/Annotations.js
@@ -204,7 +204,7 @@ function ReadingGroupAnnotationsContainer({
 ReadingGroupAnnotationsContainer.propTypes = {
   readingGroup: PropTypes.object,
   refresh: PropTypes.func,
-  route: PropTypes.string,
+  route: PropTypes.object,
   confirm: PropTypes.func,
   entitiesListSearchProps: PropTypes.func,
   entitiesListSearchParams: PropTypes.object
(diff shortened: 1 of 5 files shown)


1aurend (approved):
Thanks @dananjohnson for tracking that down! I pulled this down and replicated as you suggested. This definitely seems like the right fix. I've made a similar mistake before, and I thought in the past that the linter would catch it. Am I imagining that?

I also pushed a commit updating the imports and removing the old create component that we're no longer using.

dananjohnson:
I only noticed it because my IDE colored the prop value differently than usual. There may be a way to catch it with a linter, but I don't know what that rule would be.

Thanks for the review and that commit! I'll merge this.

PR #3883, opened 2025-06-12
[B][E] Improve maker name display in lists

1aurend:


scryptmouse (approved):

api/app/models/project_summary.rb
```diff
@@ -10,6 +10,7 @@ class ProjectSummary < ApplicationRecord
 
   delegate :collected_by?, to: :project
   delegate :entitlement_subject_url, to: :project
+  delegate :creator_names, to: :project
```
1aurend, line 13: @scryptmouse just want to confirm it's okay to do this, since I'm assuming it adds another db query? The collaborator names don't show up in lists until after the whole project/text has been fetched and added to redux store without this.

PR #3888, opened 2025-06-24
[F] ULP Development

1aurend:


1aurend:
Thanks! A couple things I want to clarify:
>    Not a new issue, but the interaction with the preview card is weird. It took me a while to realize that the preview text is expandable. Can we remove that and just make the whole thing a link to the in-context annotation?

Are you thinking we would do this for this list of resource annotations only, or were you thinking for preview cards in general? I think the latter could be complicated. You wouldn't want to have to leave the My Notes page to view the full subject of annotation. But it does seem unlikely that a resource would be attached to a large block of text, or even if it was, that it would be useful to see it in this context.

>    I don't love the Delete UI/UX. (Never been a fan of it in the reader either.) Could you use maybe a secondary button style instead and have it open a confirmation modal? For some reason "remove" also sounds like a better word to use than "delete". Maybe because I see it as more of a relation than a record.

This is on the comments, yes? I think Terence mentioned this recently as well. Comments have a `deleted` property that annotations don't, hence the two-step delete. I assume there's history there, but would it make sense to make the delete behavior consistent between the two?

dananjohnson:
> Are you thinking we would do this for this list of resource annotations only, or were you thinking for preview cards in general?

In general, I don't think the expanding capacity of the cards is dialed in. It's not really clear you can expand them just by looking at them. I'd be OK scoping an improvement just to this list of resource annotations for now (i.e. removing the expand/collapse altogether).

> would it make sense to make the delete behavior consistent between the two?

Again, let's solve this for the resource annotation list only for now. It'd take some more thinking to have a solution that worked both there and for comments, where the design matches the other action buttons.

1aurend:
@dananjohnson This should be ready for another look. I think I've addressed all of your feedback, except the target style for reasons we discussed. I'll also bring up the notation vs. annotation language in the channel today. For the resource collection detail, I applied the two-column layout from the resource detail to get annotations in there. I think it's better than always putting them below the fold, but the styles might need to be dialed in a bit further. I haven' been satisfied with anything I've tried for the column header, for instance. Let me know what you think!

dananjohnson (changes requested):
Almost there, @1aurend!

- For the resource detail, does it make sense that we're hiding comments altogether for unauthenticated users? Are comments on resources never public?
- Interaction with the preview card is much better! However, while we do want to the whole card to be clickable, the HTML will have to change a bit to be spec- and WCAG-compliant. Specifically, the Remove button can't be nested inside the `<a>` wrapping the card. You'll need to move the `<a>` lower down—wrap it around the actual annotated text—then use a pseudo-element to expand the click area to fill the card. The items in [this CENC news list](https://staging.foodbankcenc.org/news) are one place you could reference for code.
  - Can you left-align the Remove button?
- The two-column layout on the Resource Collection detail doesn't look quite right to me. I'm not sure what the alternative could be, but I can play around with it and see if I can come up with anything.
- Seeing this in my console: 
```
WARNING in ./src/global/components/Annotation/Annotation/ResourceAnnotation/styles.js 25:3-19
export 'transientOptions' (imported as 'transientOptions') was not found in '@emotion/styled' (possible exports: default)
```

1aurend:
Splitting into My Notes work that can merge now (#3892) and resource detail work to build on.

PR #3892, opened 2025-07-18
[F] ULP My Notes

1aurend:


1aurend:
@dananjohnson this includes everything from the ULP work that should be able to merge now. We may still want to do more, but I spent some time this morning improving the behavior of the Collapse on the note summaries. I removed the button from ones that aren't actually expandable, so it doesn't seem like you can click, and fixed some of the transitions.

PR #3894, opened 2025-07-21
[F] Feature: Expand Resources

1aurend:


1aurend:
@dananjohnson sorry for the delay. This branch will need to be rebased after #3892 goes in, but should be ready to build on. I'll get the initial api changes we talked about in here shortly, so rendering annotation blocks in the reader is unblocked.

PR #3900, opened 2025-07-28
[B] Various admin fixes

1aurend:
Also resolves:
- MNFLD-1056
- MNFLD-1064

### DIFF
--- a/client/src/api/notifications.js
+++ b/client/src/api/notifications.js
@@ -8,7 +8,7 @@ export default {
     return {
       level: 0,
       heading: "The project has been deleted.",
-      body: `Alas, it has passed into the endless night.`,
+      body: `Alas, the project has passed into the endless night. However, it may take a minute or two for all traces to disappear from the frontend.`,
       expiration: defaultExpiration
     };
   },
--- a/client/src/api/notifications.js
+++ b/client/src/api/notifications.js
@@ -8,7 +8,7 @@ export default {
     return {
       level: 0,
       heading: "The project has been deleted.",
-      body: `Alas, the project has passed into the endless night. However, it may take a minute or two for all traces to disappear from the frontend.`,
+      body: `Alas, the project has passed into the endless night. However, it may take a minute or two for all traces to disappear from your Manifold library.`,
       expiration: defaultExpiration
     };
   },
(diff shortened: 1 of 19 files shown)


1aurend:
@dananjohnson happy to discuss handling of entities marked for purge more, if you'd like to see more changes here. Since the deletion time was fairly short, and there are so many places where entities appear— main lists, detail pages, RG homepages, etc., I settled on a sort of in-between approach: (1) added a note to the project and text delete notifications, saying it may take a few minutes for the entity to be completely scrubbed from the frontend; (2) applied the `removes` redux store option consistently to text deletions— I believe this only affects the deleting user's view, but at least pagination is recalculated in the admin list for that user— and (3) exposed `markedForPurgeAt` on relevant entities and hid them from key, unpaginated places— the homepage and text by category list. We can certainly also hide marked entities elsewhere and/or make detail pages for marked entities 404. Just let me know how you think we should proceed.

client/src/api/notifications.js
```diff
@@ -8,7 +8,7 @@ export default {
     return {
       level: 0,
       heading: "The project has been deleted.",
-      body: `Alas, it has passed into the endless night.`,
+      body: `Alas, the project has passed into the endless night. However, it may take a minute or two for all traces to disappear from the frontend.`,
       expiration: defaultExpiration
     };
   },
```
1aurend, line 14: At some point, I think it might be nice to replace this list of notifications that are handled via the redux store. They're easy to miss, and now that we're doing most of this stuff with hooks, it's nice to be able to customize the data passed into the messages in the relevant component.

dananjohnson: I agree. Then we can also localize these messages.

client/src/config/app/locale/en-US/json/shared/notifications.json
```diff
@@ -20,6 +20,7 @@
     "resource_collection_delete": "The resource collection has been destroyed.",
     "resource_delete": "The resource has been destroyed.",
     "text_delete": "The text has been destroyed.",
+    "text_delete_body": "{{title}} has passed into the endless night. However, it may take a minute or two for all traces to disappear from the frontend.",
```
dananjohnson, line 23: I like how you've continued the poetic resonance of the first sentence 🙁 I do wonder if "frontend" is the right term though. Maybe "…from your Manifold library" or "Manifold instance" is more clear?

dananjohnson (approved):
Nice job! I think the way you've handled that delay in removing all remnants of the text is a good solution for this situation. Waiting a couple minutes is totally fine with a little heads-up to the user.

client/src/config/app/locale/en-US/json/shared/notifications.json
1aurend: Good call— I like "Manifold library"

PR #3911, opened 2025-08-21
[F] Add browse image assets to text editor

1aurend:


1aurend:
@dananjohnson this should be ready for a polish pass now. Let me know if you'd like to do it or if you'd like to give me feedback— I'm happy either way. At a minimum, I think we should add a a list view (rather than the grid) for mobile.

dananjohnson:
Awesome! I'll take a look Monday at the latest.

dananjohnson (commented):
Nice! Why don't you take the polish pass. I'd maybe try something along these lines:
Image

We don't have exact existing design for this view, but I think the grid of resource types in the new Add a Resource drawer could be a sort of precedent.

And agreed about a list view on mobile, probably a simplified version of the ResourceRow entity list component. Alternatively, maybe there's a switch that toggles between list and thumbnail view rather than tying it to viewport size?

A couple other little things:
- note the weird issue with the pagination ellipses in the screenshot
- Is it possible to have only one modal open at a time to avoid nesting? At the very least, can we avoid doubling up the backdrop?

dananjohnson:
Also, the modal height seems not quite right. It should have a max-height that's in `dvh` units, but otherwise it should be based on the content.

1aurend:
This should be ready for another look @dananjohnson. I decided to tie list view to breakpoint because we just don't have any info to show to fill in something like a resource row for these assets. There's really only name, thumbnail, and file extension, so it looks like a lot of blank space at wider breakpoints.

dananjohnson:
@1aurend I rebased this and made a couple little adjustments, but otherwise I thought it looked good. I'm gonna go ahead and merge it into the feature branch.

PR #3914, opened 2025-09-03
[POC] Ensure valid resource annotation DOM nesting

1aurend:


1aurend:
@dananjohnson I think I've got everything working as we discussed, but this could use another set of 👀 for regressions. We also need to finalize the messaging/button labels/headings for the new flow.

dananjohnson (changes requested):
This appears to be working well! No DOM nesting warnings. I didn't test the block annotation restriction extensively, but it did work as expected with the tests I did, and the blockRegex value seems fine. People may `display: block;` other elements in their CSS, but we can't do much about that.

With the order of annotation steps changed, I think we could adjust the modal headings a bit. Let's add "1. " and "2. " to make it more clear there are steps. And maybe "Add a resource annotation" could become something like "2. Choose your resource" to better indicate the action of that step. I think the warning and disabled state in the style modal could also use a bit of polish, but may need to experiment a bit to say exactly how.

I also see two small a11y issues to address:

- [ ] the inline resource button doesn't have any text to give it an accessible name, just an SVG
- [ ] the Sidebar component needs `aria-hidden="true"` to make its contents hidden from the accessibility tree

PR #3918, opened 2025-09-15
[B] Adjust admin project collection drawers

1aurend:


dananjohnson:
@1aurend could you take a look on Monday at the commits I added here? https://github.com/ManifoldScholar/manifold/pull/3918/commits/f304885110f6463cbcd9c9eadaaceae952f22b72 in particular. That one makes a global change to drawers by switching the order of the drawer element and the focus trap. I'm not really sure why we've always had the trap inside the drawer, but flipping it fixes the issues we've seen where clicking in the padding of the drawer would close it. It also aligns more with how `focus-trap` and `focus-trap-react` demonstrate how to use the libraries. My testing suggests it's a safe change, but it'd be good to get some extra 👀 on it.

This is your PR, so just let me know if things look OK and I'll approve and merge it. Thanks!

1aurend:
@dananjohnson this change makes sense to me. I also pulled it down and clicked around a bit and didn't see any drawer regressions. So, I'd say go ahead and merge.

PR #3923, opened 2025-09-26
[F] Add env config

timbot1789:


api/spec/rails_helper.rb
```diff
@@ -122,8 +122,8 @@
 # Ensure imagemagick is loaded and accessible. If it's not, there is no sense
 # in running the test suite because dozens of tests are gonna fail.
 begin
-  MiniMagick.cli
-rescue MiniMagick::Error => e
+  ImageProcessing::MiniMagick.source('data/assets/images/1x1.png')
```
timbot1789, line 125: MiniMagick Removed the `cli` method in version 5. As MiniMagick isn't a project dependency (ImageProcessing is), we shouldn't have to care about upgrades to it, and shouldn't be calling into it. This test still checks that imagemagick is loaded, but does not call out to MiniMagick

timfrazee:
Preempted by #3998

PR #3925, opened 2025-09-30
[E] Improve authorization UX

1aurend:
@dananjohnson do you mind giving this WIP a review? I already see a few things I want to test/update. I'll add notes after I create this PR. At this point, the new UX should be working and should generally create a consistent experience across the various ways of encountering an auth error. Try pasting in the url of a draft project and reader section and also clicking links to texts, resources, etc. from a restricted project detail page. However, the implementation is kind of piecemeal— in SSR, I'm still handling these errors outside of the store/router context, for example, and reader routes have some extra nuance because, when we recently started fetching the section as well as the text in SSR, it made it possible to get multiple auth errors on a single route. Anyhow, I've tried to make as few code changes as possible, since this touches so much global stuff. Do you think that's the wrong approach? Should we push the implementation further? Use a url param everywhere rather than checking both params and router state, for example?


client/src/frontend/containers/Login/index.js
```diff
@@ -1,12 +1,38 @@
...
+
+  useEffect(() => {
+    const searchParams = new URLSearchParams(location.search);
+    const redirectUri = searchParams.get("redirect_uri");
+    const shouldNotify = !!redirectUri || !location.state?.postLoginRedirect;
+
+    if (shouldNotify && !notifiedRef.current) {
+      notifyUnauthorized();
+      notifiedRef.current = true;
+    }
```
1aurend, line 34: I want to test to see if this ref check is still necessary. I was getting maximum depth errors from notifications here at one point, but this got simplified a lot.

client/src/global/components/FatalError/index.js
```diff
@@ -1,84 +1,72 @@
...
-      error: PropTypes.object,
-      type: PropTypes.string
-    }).isRequired,
-    headerLineOne: PropTypes.string,
-    headerLineTwo: PropTypes.string,
-    dismiss: PropTypes.func,
-    dispatch: PropTypes.func,
-    i18n: PropTypes.object
-  };
+export default function FatalError(props) {
```
1aurend, line 17: I'm leaning toward pulling the presentational part of this component out into a separate component so we have a component which can render outside redux/router context and wrapper which needs both and handles redirects when called from within the app.

client/src/global/components/FatalError/index.js
```diff
@@ -87,84 +75,108 @@ class FatalError extends PureComponent {
       <Redirect
         to={{
           pathname: url,
-          state: { projectAuthorizationError: this.error }
+          state: {
+            showLogin: true,
+            postLoginRedirect: redirectPath,
+            setIsRedirecting: true
```
1aurend, line 81: Need to remove `setIsRedirecting` here. I tried using state here to update state in the Manifold Container to prevent the double redirect, and I'm still not sure why it didn't work. Only the error check worked.

client/src/global/containers/Manifold/index.js
```diff
@@ -197,11 +214,13 @@ class ManifoldContainer extends PureComponent {
...
           />
-          {fatalError.error ? (
+          {/* Add isRedirecting check here to ensure that FatalError doesn't trigger another redirect before we've finished navigating home on logout */}
+          {fatalError.error && !this.state.isRedirecting ? (
             <div className="global-container">
               <FatalError
                 dispatch={this.props.dispatch}
                 fatalError={fatalError}
+                redirectPath={this.props.location.pathname}
               />
```
1aurend, line 224: This is a weird one, but if you try on Edge to logout while viewing a project detail or one of its subpages, you don't get redirected home on logout but rather back to the project detail. I don't think this is intentional, and it's odd to only have the behavior on project pages. As far as I can tell, it's a side-effect of this component handling both the redirect and displaying errors. It seems like there's a render where the error still exists in the redux store after redirect has been called. That triggers this component to trigger a duplicate redirect.

client/src/theme/styles/components/global/notifications/notification.js
```diff
@@ -35,6 +35,10 @@ export default `
       @starting-style {
         opacity: 0;
       }
+
+      &.notification--error {
+        background-color: var(--color-notification-error-light);
+      }
```
1aurend, line 41: I went with the existing notification styles with an error style for now. Open to ideas for other ways to style/display this.

dananjohnson (commented):
This is looking good! I got the login page with the access denied message with every unauthorized request i attempted. The notification could be polished a bit—the Dismiss button, at least, needs more contrast. (We could also probably just remove that button in this context.)

One thing I'm wondering about is if/how we could have more granular control over the messaging. The original request came up in the context of users trying to view a reading group they weren't a part of. Ideally, we'd say something more like, "You need to join this group to view it." or something. At the very least, I think we need to provide accurate messaging. Logging in won't always resolve the authorization issue. What do you think?

1aurend:
We have varying amounts of context for messaging — much more for a client-side navigation than for SSR — but we always have at least the request path and the message coming back from the api, which usually has the record title in it, something like "You are not authorized to read [title]". Can we come up with a set of messages that we can select from based on pathname?

dananjohnson:
Got it. So yeah just thinking more about this, there are basically two different scenarios we need to account for:
1. not logged in (e.g. trying to view public reading groups page)
2. insufficient access (a reader trying to access the admin, a project editor accessing a project they don't work on, the RG example i mentioned above)

For the first one, it makes sense that we show the Access Denied message on the login page with a redirect_uri param. The current implementation of that is probably sufficient (screenshot from Edge).

Screenshot 2025-10-01 at 3 07 08 PM

For the second one, it doesn't actually make sense to redirect them to /login. We want to just provide a message, "you don't have access to this project/resource/reading group, etc", and not display the authorized content. Is it possible to do this within the context of the application rather than with the global error screen? Like if I'm an editor who can access the backend but not a particular project, I'm still in the backend when I see the access denied message. That seems ideal to me, but I don't know how complicated that is. At the very least, the error screen needs better messaging.

Screenshot 2025-10-01 at 3 11 14 PM

1aurend:
That's a helpful distinction. I'll have to think some more about how to implement 2. The hard part is still SSR. Depending on how much time we want to spend, we can do pretty much whatever we want once we're in the client. However, currently when SSR encounters a fatal error, it just sends backs some simple html with the error message. It omits the high-level contexts like store and routing, so we can't just drop a link back home into the server html. From what I understand, we can solve this either by rendering more in the error SSR or by redirecting to another route, and then rendering the error message there. What do you think about either redirecting to `/` with the unauthorized message in a modal or adding a `/unauthorized` route with a simple message to handle these cases? I need to double check, but we should be able to check the request for a token to determine where to redirect in SSR.

1aurend:
OK— @dananjohnson this is ready for another review. Messages can still use some refining, but this branch should now:
1. When no user is logged in:
  - Redirect to either `/login` or to the `failureRedirect` specified in the `Authorize` hoc with the login overlay open
  - Show an Access Denied message
  - Redirect to the original path after login
2. When a user is logged in:
  - Redirect to `failureRedirect` specified, or if not, to the list route or to `/` or `/backend` 
  - Show an Access Denied message as a global notification

These should happen for both client-side navigations and SSR, and regardless of whether there's an actual api error or just a failed auth check, though the access message might not be exactly the same. We can fix this, of course by making them all generic, but I left all the existing, unique `failureNotification` messages that are passed to the hoc. These only show up when there's no api error, however, because api errors are caught higher up the component tree.

A few things I don't like:
1. There's a fair bit of duplicated code in my new `AppWrapper` for fatal error and in the refactored `Authorize` hoc. The former handles api errors (and the unique pre-existing redirect for access restricted projects where users can view the detail page but not texts or resources), while the latter handles client-side auth errors.
2. I'm using a combination of query params and router state to handle notifications and redirects. It would be great to do it all with params, but I didn't refactor that far. Additionally, it seems like router state isn't passed to the client when the redirect happens in SSR, so it's possible for the same redirect to contain different information in the different contexts.
3. Notifications... sigh... it may have been a mistake to handle these notifications through the existing mechanism. Notifications are handled through the redux store, and they seem like another thing that isn't passed to the client from SSR. There are a number of workarounds for that in here that are clunky. Also, because they're set in the store and rendered in the Manifold container, timing the notifications and their animation with the redirects is tricky. The notification can show up too soon.

Anyhow, looking forward to your thoughts on where we should go from here.

dananjohnson (changes requested):
Thanks for all your work on this! I think it's looking pretty good. I do wonder whether doing a redirect and notification is the right pattern for permissions errors. My sense is that generally this is best handled with a 403 response and an error message shown in place of the content of the page. Like this basically (but obviously nicer looking; maybe riffing off the global error screen):
Image

Is that architecturally even a feasible idea? Does it simplify things? Make them more complicated? We've put a fair amount of time into this already, so I don't want to pursue it if it's going to be a pain. I do think the redirect/notification approach is working pretty well.
 
A few other things I noticed:
* The public reading groups route (/groups) and group detail pages don't appear to require log-in anymore. This seems like a regression. Edge doesn't behave this way.
* It seems like it takes a long time and a client render to get the login page to show up. It hangs like this for awhile. Do you know why?
Image

* Trying to enter the backend as a user with a reader role redirected me to the login page, which doesn't seem right. Should that go to the homepage instead?
* Redirecting to the reading groups list page causes the filters on that page to show the "Reset search and filters" button, presumably because of the failure param in the URL.

1aurend:
Notes on the small things. Will reply to the larger question separately...
 
>     * The public reading groups route (/groups) and group detail pages don't appear to require log-in anymore. This seems like a regression. Edge doesn't behave this way.

Whether this page requires login is determined by the public reading groups disabled setting. I get the redirect locally, if I turn that setting on. Let me know if you see something different.

>     * It seems like it takes a long time and a client render to get the login page to show up. It hangs like this for awhile. Do you know why?

Not entirely. It looks like nothing inside `Container.Form` can render on the server, but I haven't spent time figuring out why. The same thing happens if you load an admin page that's a form and not a list.
 
>     * Trying to enter the backend as a user with a reader role redirected me to the login page, which doesn't seem right. Should that go to the homepage instead?

Yes, agreed.

>     * Redirecting to the reading groups list page causes the filters on that page to show the "Reset search and filters" button, presumably because of the failure param in the URL.

Thanks for the catch!

dananjohnson:
> Whether this page requires login is determined by the public reading groups disabled setting. I get the redirect locally, if I turn that setting on. Let me know if you see something different.

Thanks for clarifying! I think I got it mixed up in my head that one of the spam mitigation things we did was to prevent unauthenticated users from viewing those pages anymore. But you're right, it's a setting in the admin.

1aurend:
@dananjohnson I just pushed the quickest of POCs of showing an error message instead of redirecting, and I do think it simplifies things. If you think we should move forward with this, I can clean up the code and actually build an error message component. 

Two things to note:
1. We can show the correct status code for frontend pages, but it's 200 for at least some admin pages with our current architecture. If the user can view the entity, the api request returns successfully, and the auth check only happens client-side several containers down the route tree. Maybe I'm just missing it, but I don't know how to update the response status at that point.
2. Because of how the containers are nested, what the user sees for header/footer/background depends a little on what level of access they have. A reader user with no backend access with get thrown out at the first backend container and only ever see the frontend style header. A user with some admin access will see the admin layout components. We could adjust the auth checks, so the full path is checked higher up the tree, but I haven't looked into how much work that would be.

1aurend:
> > Whether this page requires login is determined by the public reading groups disabled setting. I get the redirect locally, if I turn that setting on. Let me know if you see something different.
> 
> Thanks for clarifying! I think I got it mixed up in my head that one of the spam mitigation things we did was to prevent unauthenticated users from viewing those pages anymore. But you're right, it's a setting in the admin.

I don't love how it works, fwiw. It's another case where the auth check is client-side, so there's a flash of the content in SSR before the login redirect.

dananjohnson:
Thanks @1aurend! I do like this better than making a decision about where to redirect the user, which is likely to be unexpected and unintuitive to them. I also like it better than the current global error screen, since it's presented in the context of the application and by showing the header lets users navigate away as desired. Understood about the status code; it's not ideal, but I don't think that's a deal-breaker. Can we still set `<title>` correctly? So let's proceed with an actual component and then we can refine the messaging. If it were me, I would start with the style of the global error but theme the colors to the current part of the application.

1aurend:
OK @dananjohnson— I think we did it! I could use your 👀 on the error message presentation. I think the global error works very well in frontend/reader contexts and a little less well in admin. I decided to just render the global error component itself, since I already split it's presentation off from other logic, and it handles setting `<title>`. But we could also fork it into a new component for this purpose.

dananjohnson:
Nice! Yeah I agree that it doesn't look great in dark mode. I think we just need to adjust the colors a bit; I can take a stab at that.

The last thing I think is to make the messaging a bit friendlier. This isn't great for a reading group you can't access:

Screenshot 2025-10-14 at 11 42 18 AM

I probably wouldn't show `403 Error: Read Failed` at all. And have the next line say something like, `You are not authorized to view this reading group.` The other specific scenario I want to make sure we solve for is this one raised by the CUNY team: 
> Messaging when someone tries to access a reading group without verifying their account - currently goes to an error page without saying why or what to do

Can you look through and come up with a simple, DRY-ish solution here?

1aurend:
Messaging updated, and I added auth checks to a number of places where they were missing in admin. When you update the styles for the admin error, note that it can also appear nested in the records section when a user has access to some but not all of these pages. Again, if you think this is a problem, we can find a way to move the auth check higher up, but I don't want to make more work if not needed.
Screenshot 2025-10-15 at 12 08 01 PM

I will continue to dig into the unverified messaging piece...

dananjohnson:
@1aurend I pushed a couple small UI changes to the fatal error component.

I think this is very close now! One small UX thing I noticed this time as a user with editor permission: I tried to view a reading group on the frontend I didn't have access to. Got a 403 as expected. Then I clicked Enter Admin Mode and was taken to the corresponding group in the backend, which of course I didn't have access to either. Should that link redirect to the backend dashboard if someone uses it to get to a route they're not authorized to visit? I know this is kinda edge case-y, so I don't think it's worth worrying about too much.

1aurend:
> @1aurend I pushed a couple small UI changes to the fatal error component.

Thanks!

> I think this is very close now! One small UX thing I noticed this time as a user with editor permission: I tried to view a reading group on the frontend I didn't have access to. Got a 403 as expected. Then I clicked Enter Admin Mode and was taken to the corresponding group in the backend, which of course I didn't have access to either. Should that link redirect to the backend dashboard if someone uses it to get to a route they're not authorized to visit? I know this is kinda edge case-y, so I don't think it's worth worrying about too much.

I thought about this over the weekend, and I have an idea for how to handle those admin redirects generally when the user is coming from a 403 on the frontend. It should be a quick fix. I'll give it a shot, and if it turns out to be more complex, we can leave for now.

1aurend:
OK— seems like that worked. There may still be edge cases, but it's working for reading groups and projects.

dananjohnson (approved):
I think this is finally done! Shall we merge it into `next`?

PR #3926, opened 2025-09-30
[B] Add guard on new content block defaults

1aurend:


dananjohnson:
Out of curiosity, do you know how this issue was introduced? Or has it been around for a long time and we just hadn't noticed it?

1aurend:
> Out of curiosity, do you know how this issue was introduced? Or has it been around for a long time and we just hadn't noticed it?

No, I'm pretty sure I introduced it in v9 when I implemented a fix for [default attributes](https://github.com/ManifoldScholar/manifold/commit/697483f29a9aff52a68d31d67ae3b75510075cbd) not being applied.

PR #3928, opened 2025-10-02
[F] Improve theming

1aurend:


1aurend:
@dananjohnson this should be ready for an initial review. Is there anything else we wanted to accomplish here?

Note: I haven't yet rebased `next` and this branch to include the fix for saving updates to features, so that won't work here.

dananjohnson (changes requested):
This is looking good! A few little things:

- Let's adjust the language below the accent color field now, since it's not a text input. Let's just simplify: `Enter a color in an accepted CSS color format. Leave blank to restore default color.` Maybe also adjust the others a bit, to `Override the ____ color. Leave blank to restore default color.` (Interestingly and semi-related, I noticed that if you type a named color like `rebeccapurple` in the picker it will convert it for you!)
- Can you standardize the focus outline color in the picker to always be the Manifold accent? Sometimes it's blue, sometimes it's green.
- Is it possible for hitting Enter to just close the picker rather than also submitting the form?

Let's leave it there for now. Not everything we wanted, but it's definitely an improvement! Thanks Lauren!

1aurend:
>Interestingly and semi-related, I noticed that if you type a named color like `rebeccapurple` in the picker it will convert it for you!

Oh neat! I hadn't tried that.

PR #3930, opened 2025-10-06
[E][B] Improve annotation markup

1aurend:


client/src/reader/components/section/body-nodes/nodes/Math/annotationHelpers.js
```diff
@@ -103,9 +103,10 @@ export const getAnnotationStyles = (
   const interactiveAttributes =
```
1aurend, line 103: Most of this file is just duplicated from the logic in `Text` node. It would be lovely to extract it to a single place. Should we do that now?

client/src/theme/styles/components/reader/annotations/marker.js
```diff
@@ -100,6 +100,8 @@ export default `
   }
 
   .annotation-highlight {
+    ${buttonUnstyled}
+
```
1aurend, line 104: One thing I can't figure out is how to handle styles for highlights that span multiple lines. Is there a way with a button to have it flow with the text?

client/src/reader/components/section/body-nodes/nodes/Math/annotationHelpers.js
dananjohnson: Sure seems like it!

client/src/theme/styles/components/reader/annotations/marker.js
dananjohnson: This is ringing a bell, and has something to do with what MDN is saying [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button):
> By default, user agents style buttons as display: flow-root, which establishes a new [block formatting context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_display/Block_formatting_context) and centers the button's children both horizontally and vertically as long as they do not overflow. If the button is defined as a flex or grid container, the children will behave as flex or grid items. A button set to display: inline will be styled as if the value were set to display: inline-block.

I think the UA does this block formatting when a button's content extends to multiple lines, and there's really nothing we can do about it. So I think we'll just have to use `<span role="button" tabindex="0">`. (And make sure the key events match `<button>`: https://adrianroselli.com/2022/04/brief-note-on-buttons-enter-and-space.html).

dananjohnson (commented):
This looks close!

1aurend:
@dananjohnson is this ready to merge?

PR #3933, opened 2025-10-09
[B] Treat <i> as element rather than mark

1aurend:


1aurend:
@dananjohnson can you 👍 this when you have a minute? It's the simple fix I mentioned in slack.

PR #3941, opened 2025-10-20
[B] Fix login a11y issues from TPGi report

jen-castiron:
Changes
Login
[B] Add aria-modal=true to login dialog

[B] Add aria-required=true to login inputs
-- Add ariaRequired prop to BaseInput

[B] Add outline to login inputs, button
-- Add --outline-width: 2px to root appearance vars for future use

[B] Fix login live region a11y issues
-- Change span live region to output
-- Always render the output tag, only render inner html dynamically
-- Empty output tag on form submission

[B] Fix login route a11y issues
-- Change dialog bg wrapper class, but keep dialog's bg color. The background class applies appropriate theme vars for dark backgrounds
-- Add border to login inputs using theme vars

Screenshot 2025-10-21 at 11 33 55 AM

### DIFF
--- a/client/src/global/components/sign-in-up/Overlay/index.js
+++ b/client/src/global/components/sign-in-up/Overlay/index.js
@@ -18,6 +18,7 @@ export default function Overlay({ hideOverlay, active = false }) {
       }}
     >
       <Styled.Dialog
+        className="bg-neutral90"
         role="dialog"
         aria-modal
         aria-labelledby={uid}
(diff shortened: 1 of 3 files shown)


jen-castiron:
> This is looking good. I noticed one odd thing, which is that in Voiceover with Safari the output is announced twice. Doesn't happen with Chrome. Can you see if there's anything we're doing to cause that? If not, let's not worry too much about it.

I removed the extra aria attributes and checked to see if there was duplicate screen reader text anywhere, but I don't see anything. Even after removing the aria attributes, Safari is announcing output twice. 

Some StackOverflow answers suggested that it was only happening with nested tags, but when I removed the nested span, VoiceOver still announced it twice. 

From what I understand just searching for a fix, this is a widespread VoiceOver bug.

PR #3942, opened 2025-10-20
[B] Fix a11y issues in Contents, Collect Toggle component

jen-castiron:
# Changes
These are all the changes that were listed under the "Contents" component in the TPGi a11y spreadsheet: https://docs.google.com/spreadsheets/d/1D1DzN7D5s9HeiH_r8sqI5PcNJuRUhFiiHw89G5Iov94/edit?gid=1125208726#gid=1125208726

A large portion of these changes revolve around the Collect Toggle, as there were several rows on the a11y sheet for this component. I've kept each change as separate commits so you can see my progress - we may want to squish them into one later. 

[B] Update collect/uncollect label for a11y
-- Change "Collect/Uncollect {title}" to "Add/Remove {title} to My Starred or Reading Group"

[B] Fix collect toggle button a11y
-- These changes replace the change above. Dana determined that the label was unnecessarily long.
-- Remove aria-hidden on button
-- Always show a label, add or remove
-- Remove lengthy label in favor of Add or Remove
-- Remove separate screen reader status

[B] Set darker hover color to "About this text" in TOC drawer (Row 36)

[B] Only hide title bar inner on mobile, do not hide h1 tag
-- The h1 tag was removed on mobile devices with `display: none`

[B] Add guard to collecting helpers collectedIdsForCollection
-- This bug came up  as I was navigating the app

[B] Fix spacing around Collecting Toggle, remove are you sure (Row 12)
-- Add width to TOC on tablet devices
-- Add margin around Collecting Toggle (if it's inline)
-- Add margin around Collecting Toggle in Reader Header

### DIFF
--- a/client/src/frontend/components/collecting/Toggle/Text.js
+++ b/client/src/frontend/components/collecting/Toggle/Text.js
@@ -10,12 +10,6 @@ function determineText(view, t) {
         key: "remove",
         text: t("actions.remove")
       };
-    case "remove-confirm":
-    case "remove-confirm-active":
-      return {
-        key: "remove-confirm",
-        text: t("messages.confirm")
-      };
     case "add":
     case "add-active":
       return {
--- a/client/src/frontend/components/collecting/Toggle/index.js
+++ b/client/src/frontend/components/collecting/Toggle/index.js
@@ -15,10 +15,9 @@ const COLLECTABLE_TYPE_RESTRICTED_LIST = ["journals"];
 
 const { request } = entityStoreActions;
 
-function determineView(collected, hovered, confirmed, isCollecting) {
+function determineView(collected, hovered, isCollecting) {
   if (collected) {
-    if (hovered && !confirmed) return "remove-active";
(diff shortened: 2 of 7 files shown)


jen-castiron:
@dananjohnson These changes are ready for your review. I'll consolidate the Collect toggle into one commit if it's good to go!

dananjohnson (changes requested):
I can't pinpoint what's causing it, but your branch introduces a regression in the reader TOC drawer where you can't tab through the links and toggles with a keyboard anymore. Tabbing only seems to cycle between the first link, toggle, and the `<body>` element. I tested on `master` to confirm it was a new regression. Can you see anything in your code that would cause that?

Otherwise I thought everything else you did here looked good!

jen-castiron:
@dananjohnson that was weird! The old css for the button had `visibility: hidden` and was preventing tabbing, but I don't know why you couldn't tab to other links. This was force pushed - pull the fresh branch and give it another shot

PR #3943, opened 2025-10-20
[B] Catch api error + add messaging re: creating a maker

1aurend:


client/src/config/app/locale/en-US/json/backend/projects.json
```diff
@@ -11,6 +11,8 @@
     "manage_resource_collections": "Manage Resource Collections",
     "contributors_header": "Manage Contributors",
     "add_contributor_label": "Add Contributor",
+    "contributor_maker_error": "Maker is required. If adding a new maker, you must first create the maker in the records tab before adding as a collaborator.",
```
dananjohnson: `records tab` is a bit confusing. If HTML is allowed here, we should just do something like `...you must first <a href="">create the maker record</a> in the …`. Otherwise, lets say something like `on the Maker Records page` to be more clear.

dananjohnson (changes requested):
The fix is working well. I made one comment about the error message.

PR #3946, opened 2025-10-21
[B] Fix a11y issues with Reader Header

jen-castiron:
## Changes
The Reader Header used mostly CSS to show & hide menus. These changes update the header menus to use a disclosure pattern similar to the nested header navigation menus we use elsewhere. The Notes dialog remains a dialog.

Rows 15 & 22 

[B] Fix Reader Header Options menu a11y
-- Adds a11y props to options button
-- Adds separate mobile version of options menu with inert state
-- Adds focus trap to separate mobile options menu

[B] Update to a disclosure pattern similar to the header nested navigation menus
-- Updates search, visibility, and appearance menus to a nested disclosure pattern

### DIFF
--- a/client/src/global/components/atomic/DisclosureNavigationMenu/hooks/useMenuState.js
+++ b/client/src/global/components/atomic/DisclosureNavigationMenu/hooks/useMenuState.js
@@ -12,9 +12,10 @@ export default function useMenuState(initialVisible = false, onBlur) {
 
   function handleBlur(event) {
     if (!contentRef.current || !disclosureRef.current) return;
+    const target = event.relatedTarget || event.target;
     if (
-      contentRef.current.contains(event.relatedTarget) ||
-      disclosureRef.current.contains(event.relatedTarget)
+      contentRef.current.contains(target) ||
+      disclosureRef.current.contains(target)
     )
       return;
     // state can optionally be controlled externally


jen-castiron:
@dananjohnson , this PR is ready to review as well.

dananjohnson:
@jen-castiron after reviewing the reader options menu, I felt like the approach wasn't quite right. It's not really a disclosure, because focus was being trapped, but it's also not really a dialog. I had to play around with the idea to see if it worked, which is why I made the change myself, but basically my thinking is that we don't need to hide the menu at all. I adjusted the approach so that focusing into the menu shifts it into view, and focusing out shifts it back. The Options toggle is just there for mouse/touch users then, so I removed its button semantics. Sorry for sending you in the wrong direction. This feels more accessible and simpler to me now, but let me know what you think!

Otherwise I thought what you did with the other menus looked great! I noticed one small contrast issue when a menu toggle is both active and focused.

Screenshot 2025-10-23 at 1 15 57 PM

jen-castiron:
@dananjohnson , sweet, I'll take a look!

jen-castiron:
@dananjohnson The issue with the button focus state color is fixed. I also found a bug with the Content button and fixed that too 👍 
Screenshot 2025-10-28 at 3 28 06 PM

jen-castiron:
@dananjohnson I'll take a look. I think some of those issues might be resolved by the change from radio options to button switchers, but it's still a weird bug.

jen-castiron:
@dananjohnson Bug fixed! useMenuState was closing the dialog if it couldn't find a `relatedTarget`. I've updated it to check for either `relatedTarget` or `target`

dananjohnson (approved):
Nice! Everything's working now.

PR #3948, opened 2025-10-22
[E] Improve resource alt text management

1aurend:


1aurend:
@dananjohnson here's my proposal for how to migrate away from the metadata field. I setup this separate PR in case you'd like a BE review.

dananjohnson (changes requested):
A couple little issues on the client side:
- [ ] If you add alt text on the Create Resource page in the backend, it doesn't seem to stick. The field is empty once the resource is created.
- [ ] The text color in the alt text field isn't white like the other inputs in the form.
Image

Last thing is it might be good to add some comments to what you're doing in the Properties component. It's not really clear just looking at the code why there's so much data munging.

I also think it'd be good to get a BE review on this.

1aurend:
@scryptmouse do you mind giving this a quick 👀 ? For context, if it's useful, the current resource alt text UX is a bit confusing— the field is buried in the metadata tab and not clearly tied to a particular upload. Because of how the metadata form fields are generated programmatically in the client, it seemed better to just remove alt from metadata and switch to the alt stored on the attachment itself.

1aurend:
@dananjohnson feel free to make adjustments/merge while I'm out if you're working on resources.

PR #3950, opened 2025-10-24
[B] Fixes for Audit 2

jen-castiron:
## Changes
The accessibility issues in Audit 2 were fairly strait forward and simple, so it made sense to group them together. See commits for list of changes.


jen-castiron:
@dananjohnson This PR is ready for review! It includes all changes needed for Audit #2. Details in commit titles/messages

client/src/reader/components/control-menu/AppearanceMenuBody.js
```diff
@@ -247,29 +267,48 @@ class AppearanceMenuBody extends Component {
...
-            <Utility.IconComposer
-              icon="CheckUnique"
-              size={30}
-              className="appearance-menu__color-scheme-icon"
-            />
-            <span className="screen-reader-text">{option.label}</span>
-          </label>
-        ))}
-      </fieldset>
+        </div>
```
dananjohnson: If you switch this from a fieldset/legend to divs, then you need to restore the association of the label with the group using `aria-labelledby`.

dananjohnson (changes requested):
Audit # 2 fixes are looking pretty good! I've noted a few things below and in comments. 

Now that I've used it myself, I'm not 100% happy with how the Appearance menu looks with the new contrast setting. I might just experiment a bit and see if anything sticks. Have you tested this setting with Windows high contrast mode yet? If not, please do that next time you're in the office.

A few little issues:
- [x] One straggler `<li>` here in the visibility menu:

Image

- [x] Also, I think we should remove the list markup in the Appearance menu as well. They didn't call it out, but we should be consistent with the Visibility menu.
- [ ] The resource tiles look good now, but resource collection tiles still need a focus indicator on the project homepage.

Image

PR #3951, opened 2025-10-27
[F] Editor style utilities— alignment

1aurend:


1aurend:
@dananjohnson this should be ready for an initial review. This version handles lists and paragraphs nested in multiple divs better than the version I showed in confab but still with some limitations. I have in mind an approach that would be even more precise, but I'd like to touch base after you've had time to review this to determine whether to spend the time.

Happy to say more about any choices I made here, but I don't want to bias your initial impressions of whether this feels like it's working as expected.

dananjohnson (commented):
@1aurend I think this is looking great! I tested with various elements and selections and got mostly the results I'd expect. A couple things I noted:

- In lists, alignment gets set on individual list items. I think we discussed this. Did we feel like that's preferable to setting the alignment on the list itself?
- I selected two list items and applied alignment. The utility classes were applied to both items and also oddly to a nested `<i>` element.

Image

1aurend:
>  In lists, alignment gets set on individual list items. I think we discussed this. Did we feel like that's preferable to setting the alignment on the list itself?

I went back and forth on this, and decided to follow Google docs in applying alignment to individual list items. I thought it might be the expected behavior for this level of user. But we can certainly switch to treating lists as a unit. I can't really image a case where you'd want multiple alignments in the same list?

>  I selected two list items and applied alignment. The utility classes were applied to both items and also oddly to a nested `<i>` element.

Oh right— this is a consequence of the logic that looks for whichever element is the nearest parent to text content. Depending on what we decide re: treating lists as units, I can make an exception and only apply the classes to `li` tags if we're inside a list.

1aurend:
@dananjohnson Ready for one more look

dananjohnson (approved):

PR #3952, opened 2025-10-28
[E] Always show select input labels (a11y)

jen-castiron:
## Changes

Adds labels back in for filter selects. Adjusts styles to match other areas where select labels are visible.

### DIFF
--- a/client/src/frontend/components/entity/Collection/EntityCollection.js
+++ b/client/src/frontend/components/entity/Collection/EntityCollection.js
@@ -32,7 +32,8 @@ function EntityCollection({
   nested,
   boxed,
   className,
-  filterProps
+  filterProps,
+  containerWrapPoint
 }) {
   return (
     <Styled.Wrapper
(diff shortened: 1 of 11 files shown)


jen-castiron:
@dananjohnson, I've split this one out since it could introduce a bit of design/style changes. For example, in the screenshot below you can see "All Projects" doesn't line up anymore. I've made a small adjustment to make sure the header aligns at "end", but that's not enough to fix it. And since this component is used in several places, I'm not sure what exactly I can do to make sure it lines up in this specific case. What do you think?

Screenshot 2025-10-28 at 4 34 42 PM

dananjohnson:
@jen-castiron I think the easiest solution is to absolutely position the labels. I tested it out and can push that change. However, now that the labels are visible, I'm realizing they could be improved! I'm working on that as well, and will update your branch when it's done.

dananjohnson:
OK, I made a bunch of changes to `<select>` and default `<option>` labels. I think this should be good to go, but let me know if you see any issues.

jen-castiron:
@dananjohnson Sweet! I'll take a look. I had initially made the labels absolute, but then the padding was all wrong and styling needed to be changed anyway. Thanks for working on it!

jen-castiron:
@dananjohnson the labels and arrows are askew on the My Notes page
Screenshot 2025-10-31 at 10 27 11 AM

dananjohnson:
> @dananjohnson the labels and arrows are askew on the My Notes page

Good catch! I'll fix it.

PR #3957, opened 2025-11-04
[B] Adjust ingestion message polling timestamps + end logic

1aurend:


1aurend:
> This seems good to me. Is there anything you need to change based on Alexa's Slack comment this morning?

I don't think so. This should be reliable. 

The only thing I can think to add might be to also poll the ingestion endpoint (maybe every minute or two?) in addition to the message endpoint. Right now, we're waiting for a particular message to come through before refetching the ingestion. If I understand correctly, because the process is now backgrounded, the only 100% way to know that the ingestion finished is to refetch the ingestion record and see that its state has transitioned. With the ingestion polling, even if something went wrong with the messages, the ingestion would show as complete if it finished. But we can probably save that for if there are further issues?

dananjohnson:
Sure. Let's release this for now, and see if we hear of any further issues.

PR #3958, opened 2025-11-04
[F] Improve Journal + Project Collection Discoverability

1aurend:
@dananjohnson @jen-castiron here's the branch with the updated library route structure. For project collections, I added redirects at the old paths to keep existing bookmarks working and updated the redirects in and out of admin.


dananjohnson:
Thanks, Lauren!

client/src/frontend/containers/Journals/index.js
```diff
@@ -0,0 +1,71 @@
...
+export default function JournalsListContainer() {
+  const filterReset = useMemo(() => ({ standaloneModeEnforced: "false" }), []);
+
+  const subjects = useFromStore({
+    requestKey: requests.feJournalSubjects,
+    action: "select"
+  });
+
+  const { pagination, filters, setFilters } = useListQueryParams({
+    initSize: 8,
```
dananjohnson: This seems to low. Can we make it match the per-page for projects?

client/src/frontend/components/entity/Collection/patterns/Journals.js
```diff
@@ -0,0 +1,72 @@
...
+  filterProps,
+  paginationProps,
+  ...passThroughProps
+}) {
+  const { t } = useTranslation();
+
+  const showFilters = !isEmpty(meta) && !isEmpty(filterProps);
+  return (
+    <EntityCollection
+      title={t("titles.journals_all")}
```
dananjohnson: @1aurend @jen-castiron I wonder if it makes sense to get rid of the "All" in the titles for these pages. Wouldn't it be fine to just say "Projects", "Project Collections", "Journals"?

client/src/global/components/atomic/EntityThumbnail/index.js
```diff
@@ -16,20 +17,30 @@ export default function EntityThumbnail({
   parentView = false,
   isListItem = false
 }) {
+  let routeName = "frontendProjectDetail";
+
+  if (entity.type === "journals") {
+    routeName = "frontendJournalDetail";
+  } else if (entity.type === "projectCollections") {
+    routeName = "frontendProjectCollection";
+  }
```
dananjohnson: Since this is a global generic component, I wonder if we should just make the route name a prop.

client/src/global/components/atomic/EntityThumbnail/index.js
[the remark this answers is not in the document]
dananjohnson: The more I look at `EntityMetadata` the more I wonder if it needs a bit of a refactor. There's a lot of shared stuff between entity types, especially UI-wise, but we might want to rethink `getMetadataProps` before adding even more logic to it.

client/src/frontend/containers/ProjectCollectionsRedirect/index.js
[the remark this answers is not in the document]
dananjohnson: Sure, thanks!

client/src/frontend/components/entity/Collection/patterns/Journals.js
1aurend: Also sounds good to me. We'll just have to sort out the mobile labels and whether we want consistency in admin:
Screenshot 2025-11-11 at 3 45 20 PM
Screenshot 2025-11-11 at 3 45 36 PM
Screenshot 2025-11-11 at 3 47 19 PM

dananjohnson: Ah, thanks for raising this. The "All" makes sense in those contexts. Still, what if we keep the mobile nav title but just use a new page title? That's how the backend routes have it.

client/src/global/components/atomic/EntityThumbnail/index.js
jen-castiron: I understand your concern, but this component isn't truly generic. It is generic to entities, which have a type and a specific route name. (The component originally went to "frontendProjectDetail", no matter what entity you gave it.) Since this component gets the full entity object with type, it seemed unnecessary to made a developer pass in the route name. 

I could see passing in the route name if there's a chance we want the preview to go somewhere other than the entity type's detail page.

jen-castiron:
@dananjohnson @1aurend 
General question on the Manifold linter... I've run into `no-nested-ternary` issues quite a bit, and noticed they get turned off in some places. We don't have that rule on our other projects - is it worth keeping around?

dananjohnson:
> @dananjohnson @1aurend General question on the Manifold linter... I've run into `no-nested-ternary` issues quite a bit, and noticed they get turned off in some places. We don't have that rule on our other projects - is it worth keeping around?

I don't love nested ternaries but am not an absolutist about it. I think my rule would be: nesting once is fine, but if you're doing more than that, find another solution like a switch. Otherwise it just gets too hard to read.

client/src/frontend/containers/Journals/index.js
dananjohnson: It looks like if you remove it, it uses the default set in `useListQueryParams`, which is `20`. That sounds like a good number to me 👍🏻

1aurend:
I noticed one odd thing on the project collections page. When I first open it, the "reset filters" button is showing:
Screenshot 2025-11-13 at 3 15 35 PM

But then if I click reset, all my collections show up, even the ones with `visible=false`:
Screenshot 2025-11-13 at 3 15 54 PM

1aurend:
> > @dananjohnson @1aurend General question on the Manifold linter... I've run into `no-nested-ternary` issues quite a bit, and noticed they get turned off in some places. We don't have that rule on our other projects - is it worth keeping around?
> 
> I don't love nested ternaries but am not an absolutist about it. I think my rule would be: nesting once is fine, but if you're doing more than that, find another solution like a switch. Otherwise it just gets too hard to read.

While I agree there's a limit, I end up disabling this lint rule a lot. I'm all for removing it.

1aurend:
These counts also really don't make sense as a user who can't see drafts:
Screenshot 2025-11-13 at 3 33 50 PM

Anyhow... just pushed a couple of commits adding the keyword filter to project collections and exposing some more counts. I went with `available_[type]_count` that reflects how many the current user can access rather than a straight published count. Let me know if something different would be better.

PR #3962, opened 2025-11-07
[F] Enable filtering by project subject on collection detail

1aurend:


1aurend:
@dananjohnson one thing I noticed while setting this up is that while we refetch subjects after someone creates/updates a subject in the backend records list, we don't refetch the list of subjects when one is added to a project. This leads to an edge case where if I have an existing subject that hadn't previously been used and I add it to a project, it doesn't appear in the filters on the library projects list until I refresh. Should we just fetch subjects each time in the projects list/project collection detail containers?

dananjohnson:
Hmm. My sense is that it's not worth fetching subjects every single time to avoid that edge case. Having to refresh the page seems like a tolerable solution.

PR #3964, opened 2025-11-13
[F] Upgrade to react-router v7

1aurend:


client/src/frontend/containers/Frontend/index.js
```diff
@@ -1,151 +1,89 @@
...
-      this.props.settings,
-      "attributes.pressLogoStyles.small"
-    );
-    return classNames({
-      "main-content": true,
-      "flex-viewport": true,
-      "extra-top": hasPressLogo
-    });
-  }
+  const commonActionsMemo = commonActions(dispatch);
```
1aurend: Clean up

client/src/frontend/containers/Frontend/index.js
```diff
@@ -1,151 +1,89 @@
...
-    return (
-      <BodyClass className={"browse"}>
-        <BreadcrumbsProvider>
-          <Utility.ScrollToTop />
+  const hasPressLogo = get(settings, "attributes.pressLogoStyles.small");
+  const mainClassName = classNames({
+    "main-content": true,
+    "flex-viewport": true,
+    "extra-top": hasPressLogo
+  });
```
1aurend: Move to markup

client/src/frontend/containers/Page/index.js
```diff
@@ -1,55 +1,38 @@
...
+  });
 
-  static mapStateToProps = state => {
-    return {
-      loading: state.ui.transitory.loading.active,
-      settings: select(requests.settings, state.entityStore),
-      page: select(requests.gPage, state.entityStore)
+  useEffect(() => {
+    return () => {
+      dispatch(flush(requests.gPage));
```
1aurend: Can clean this up and use random request key

client/src/frontend/containers/ReadingGroup/Homepage/Fetch.js
```diff
@@ -59,24 +53,17 @@ function ReadingGroupHomepageFetchContainer({
...
   return (
     <Styled.Body>
-      {childRoutes(route, {
-        childProps: {
+      <Outlet
+        context={{
           readingGroup,
           categories,
           responses,
           navigate,
```
1aurend: Shouldn't be passing navigate here

client/src/frontend/containers/ReadingGroup/index.js
```diff
@@ -63,7 +61,6 @@ export default function ReadingGroup({ route }) {
     refresh: () => setFetchVersion(prev => prev + 1),
     fetchVersion,
     navigate,
-    route,
     readingGroup,
     dispatch
```
1aurend: Unnecessary props here

1aurend:
Things to fix next week:
- The replacement for Prompt is too aggressive on drawer routes and triggers on save.
- Content blocks list doesn't refresh on add; are there others?
- Redirect after creating project permission 404s

dananjohnson (changes requested):
@1aurend this all looks and functions quite well! I looked at portions of the code, then pulled your branch down and tested it out. My testing was far from exhaustive. I did see a handful of things in the routes and user actions I tried, but in general things are working well. I think you should do a bit more testing, address any issues you see, and then we should go ahead and get this into `next` and deployed.

Great work! 👏🏻 
 
- When trying to create a new reading group, the "unsaved changes" confirm modal pops up when you click Save. The group is created, but the drawer should just close.
- Opening the Edit Reading Group drawer triggers a console warning:
```
You are trying to use a blocker on a POP navigation to a location that was not created by @remix-run/router. This will fail silently in production. This can happen if you are navigating outside the router via `window.history.pushState`/`window.location.hash` instead of using router navigation APIs.  This can also happen if you are using createHashRouter and the user manually changes the URL.
```
- Creating a new entitlement on a project triggers a console warning from `src/backend/containers/entitlements/Form.js` (though creating it is successful):
```
Form.js:21 A router only supports one blocker at a time
```
- Opening the Edit drawer on a project contributor triggers a number of proptype warnings.
- If you try to navigate to the reader with JS disabled, you'll see a server-side warning in your console. Not sure if this is worth much trouble debugging, but wanted to note it: 
```
<Navigate> must not be used on the initial render in a <StaticRouter>. This is a no-op, but you should modify your code so the <Navigate> is only ever rendered in response to some user interaction or state change.
```

PR #3968, opened 2025-11-21
[F] Add external identifiers

timbot1789:
Resolves PROJ-3275

This PR adds an `external_identifier` attribute to the `Projects`, `ProjectCollections`, `Journals`, and `UserGroups` APIs. This attribute is backed by a database table and ActiveRecord model, but is not exposed as an API endpoint itself.


timfrazee:
I've made the requested changes above so I can merge this into the saml-entitlements branch and continue work there.

PR #3975, opened 2025-12-04
[E] Add new image sizes, project collection filtering & global search fixes, 

scryptmouse:
Resolves MNFLD-1086
Resolves MNFLD-1087
Resolves MNFLD-1113

This also includes a number of recent API improvements from the `master` branch rebased to be included in next.


scryptmouse:
@dananjohnson, can you pull this branch into your local and make sure the large image size will work for your purposes? Just loading the branch won't automatically backport the sizes in existing records—that happens in an upgrade task (which runs on deploys, but not necessarily locally). However, uploading a new hero, cover image, etc should generate the styles for local testing.

dananjohnson:
Well I ended up just pushing the change to your branch since it was so quick 🙂 I'll merge it as soon as tests pass.

PR #3979, opened 2025-12-11
SAML support

timfrazee:
Adds support for arbitrary SAML SSO providers for Manifold login.

Also provides the ability for sysadmins to set a default identity provider and optionally hide local login.

Configuration of these features is done in the environment to avoid mistakes that could render users unable to log into Manifold.

### SAML

Adds support for SAML SSO providers to serve as identity providers for Manifold login.

SAML providers are defined in the environment. For each defined provider, additional settings / credentials are provided as namespaced ENV-vars.

By default, Manifold requires that it signs its metadata and Authn requests. It also requires that identity providers sign their assertions. To that end, it's necessary to generate a unique SHA256 X509 public/private key pair per-instance, which go in the `certificate` and `private_key` settings, respectively. With OpenSSL, this can be done with:

```
openssl req -new -x509 -days 365 -nodes -sha256 -out saml.crt -keyout saml.pem
```

The below example defines a SAML provider named "test1". Multiple providers can be defined as a comma-delimited list, then each provider's settings/credentials in appropriately-namespaced variables. The example shows common settings, but Manifold supports most of the options provided by [Ruby SAML](https://github.com/SAML-Toolkits/ruby-saml).

```
SAML_PROVIDER_NAMES=test1

SAML_TEST1_ENABLED=TRUE
SAML_TEST1_DISPLAY_NAME=Test SAML 1
SAML_TEST1_SP_ENTITY_ID=manifold_test_1
SAML_TEST1_IDP_ENTITY_ID=https://saml.example.com/entityid
SAML_TEST1_IDP_SSO_SERVICE_URL=https://mocksaml.com/api/saml/sso
SAML_TEST1_CERTIFICATE="-----BEGIN CERTIFICATE-----{CERT GOES HERE}-----END CERTIFICATE-----"
SAML_TEST1_PRIVATE_KEY="-----BEGIN PRIVATE KEY-----{PRIVATE KEY GOES HERE}-----END PRIVATE KEY-----"
SAML_TEST1_IDP_CERT="-----BEGIN CERTIFICATE-----{IDP CERT GOES HERE}-----END CERTIFICATE-----"
```

Additional settings:

```
SAML_TEST1_HIDDEN=TRUE # Hides the login button, but the provider remains active. Useful for testing.
SAML_TEST1_DEFAULT=TRUE # If any provider is set as default, the user will be automatically redirected to it

SAML_DISABLE_PASSWORD_AUTH=TRUE # Removes email/password authentication, leaving only SSO options
```

### User Groups

User Groups are a new model that serve to connect users to entitleable resources. Members of a user group are granted entitlements to the associated resources.

Entitlements granted through User Group memberships are granted when a user joins, and destroyed if/when a membership is destroyed. Since the membership is the entitler, this does not affect entitlements granted through other means.

This feature also adds admin UI to manage user groups, their members, and their entitleables.

### External Identifiers

External identifiers are simple records designed to hold identifier strings that connect an entitleable resource or user group in Manifold to a resource known and managed by an external system. Identifiers are unique, guaranteeing that a maximum of one resource will be returned for a given identifier.

External identifiers are defined in admin under the appropriate resources.

### External auth entitlement management

SAML IDPs can define claims named `user_groups` and `entitlements`, providing a semicolon-or-comma-delimited list of external identifiers of user groups or entitleable entities, respectively.

When a user logs in, the IDP's list of desired user groups and entitlements is compared against the user's existing identity-managed entitlements, and they are updated to match (adding or removing as needed). The user's identity is referred as the `source` on user groups and the `entitler` on entitlements, so this

### DIFF
--- a/api/app/authorizers/user_group_authorizer.rb
+++ b/api/app/authorizers/user_group_authorizer.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 class UserGroupAuthorizer < ApplicationAuthorizer
-
   # Admins only
   def self.default(_ability, user, _options = {})
     admin_permissions?(user)
--- a/api/app/controllers/api/v1/tokens_controller.rb
+++ b/api/app/controllers/api/v1/tokens_controller.rb
@@ -39,7 +39,6 @@ module API
       def token_params
         params.permit(:email, :password)
       end
-
     end
   end
 end
--- a/api/app/controllers/concerns/manages_oauth_cookie.rb
+++ b/api/app/controllers/concerns/manages_oauth_cookie.rb
@@ -42,5 +42,4 @@ module ManagesOauthCookie
       ".#{domain}"
     end
   end
-
 end
--- a/api/app/models/concerns/externally_identifiable.rb
+++ b/api/app/models/concerns/externally_identifiable.rb
@@ -12,15 +12,15 @@ module ExternallyIdentifiable
   def external_identifier=(record_or_identifier)
     case record_or_identifier.presence
     when ExternalIdentifier
-      super(record_or_identifier)
+      super
     when String
-      if self.external_identifier.present?
-        self.external_identifier.update(identifier: record_or_identifier)
+      if external_identifier.present?
+        external_identifier.update(identifier: record_or_identifier)
       else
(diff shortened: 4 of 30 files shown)


timfrazee:
Closing in favor of #3998

PR #4040, opened 2026-05-05
[B] Ensure csv resource upload attrs are set correctly

1aurend:


1aurend:
@dananjohnson this basically back-ports the fix I made to this form during the router migration. I did a quick test here, and it seems to work, but my more extensive testing was done in the migration branch rather than one based on `main`.

dananjohnson (approved):

PR #4046, opened 2026-05-07
[B][E] Improve notes drawer/list contents + UX

1aurend:


dananjohnson (commented):
The changes generally look good and worked when I tested. One thing I was thrown off by was that even after disabling reading groups on my local instance, I could still view them via the "Show" dropdown in the drawer. I could also assign existing annotations to a reading group view the annotation popup. Do we need to hide those?

Image

Image

1aurend:
> The changes generally look good and worked when I tested. One thing I was thrown off by was that even after disabling reading groups on my local instance, I could still view them via the "Show" dropdown in the drawer. I could also assign existing annotations to a reading group view the annotation popup. Do we need to hide those?

Maybe dumb question, but did you refresh after changing the setting? I noticed that I needed to refresh after disabling reading groups to get the links in the user nav to adjust accordingly. I wonder if this is the same. If it is, I don't think it should hold back this PR?

dananjohnson:
Hah, you're totally right! Thanks for pointing that out.

Let's get this change merged!

dananjohnson (approved):

PR #4050, opened 2026-05-11
Fix various Epub export bugs

timfrazee:
Fixes issues where Epub generation fails if:

* A text section node has a landmark
* The project cover is not a png/jpg
* A remote resource has no defined external source


scryptmouse (approved):
Those dang kwargs. Looks good to me!

PR #4053, opened 2026-05-19
[F] LTI Deep Linking UI

1aurend:


1aurend:
@dananjohnson I think this is at a good stopping place. Feel free to review at whatever level of detail makes sense. A couple of notes:
1. While you're in there, I'd love if you have any insights into how we can get styled-components and vite to play nicer together. SC does its job in a build; I no longer see any of the FOUC that were present with Emotion. But, it definitely feels more sluggish in dev. There are also some style flashes on nav and some HMR limitations (see the architecture doc for more on the latter). The style flashes are intermittent and seem to improve after you've visited a page a couple times, so it reads to me as either sc's parser being slower or its interaction with vite's dep optimization.
2. I believe I have search working in all the places, and I implemented the default facets we discussed, but I haven't yet removed the "everything" checkbox or update styles in the main app. I figured I'd do that in a second pass after you've had a little time to interact with it and compare the behavior to the LTI search.

PR #4063, opened 2026-06-09
fix: ensure project restricted access flow handles multiple failed fetches

1aurend:


client/src/global/components/FatalError/index.js
```diff
@@ -167,4 +213,4 @@ class FatalError extends PureComponent {
   }
 }
 
-export default withTranslation()(FatalError);
+export default withRouter(withTranslation()(FatalError));
```
dananjohnson: Does `<FatalError>` always render inside a Router context? Let's be certain it does.

1aurend: Good catch— it doesn't. I'll add a wrapper component like I did in v10.

dananjohnson (approved):

PR #4077, opened 2026-06-17
feat: complete pragmatic dnd migration

1aurend:


1aurend:
@jen-castiron @dananjohnson give this a whirl and see what you think. We should probably either implement these drop indicator styles globally or make this list consistent with the others before finalizing. 

Here's Claude's summary of how each of the findings is addressed:

| # | Finding | Status | Evidence |
|---|---|---|---|
| 0 | Deprecated `@atlaskit/tree` | ✅ Fixed | Migrated to native pragmatic-dnd; dep removed from `package.json` |
| 1 | Not marked up as a nested unordered list / no semantic context for which controls belong to which item / no parent-child semantics | ✅ Fixed | Real nested `<ul>`/`<li>`: each `<li>` wraps that row's controls + a nested `<ul>` of children. Conflicting `role="tree"` removed so list semantics apply |
| 2 | Dragging required; no single-pointer alternative (WCAG 2.5.7) | ✅ Fixed | `PopoverMenu` move actions (Move up / down / Nest / Move out) are fully click-operable — no drag needed |
| 3a | Drag control has `role="group"`, no accessible name | ✅ Fixed | Grabber is now a non-focusable `<div tabIndex={-1}>` (no role); the move control is a real `<button>` named "Reorder" |
| 3b | Uninformative "item X combined with Y" (uuid) announcements | ✅ Fixed | That was `@atlaskit/tree`'s live region. Announcements now use titles: "Moved {title}", "Entry nested under {title}" |
| 4 | Expand/collapse disclosure named "Edit", no `aria-expanded` | ✅ Fixed | Now `aria-expanded={entry.isExpanded}` + label "Collapse"/"Expand" |
| 5 | Focus goes to `<body>` after deleting an entry | ✅ Fixed | `onDelete` now moves focus to the adjacent entry's move button (or the list container when empty) |

PR #4088, opened 2026-06-24
chore: add enforce canonical domain mechanism to client

1aurend:


timfrazee (approved):
Looks good, but can't test locally. Merging, will deploy to `preview` and test there before a wider release.

PR #4093, opened 2026-07-01
Fix image attachment detection; preserve ingestion source filenames

timfrazee:
- Attempts to preserve ingested attachment filenames
  - This is not only good for posterity, but helps derive file type by extension
- DRYs up attachment type derivation logic


zdavis (approved):
LGTM but did not test

PR #4113, opened 2026-07-15
fix: clean up ingestion tempfiles proactively

timfrazee:
Fixes an issue where tempfiles weren't being explicitly unlinked, and the teardown method wasn't being invoked.

This PR creates a new base interaction class for ingestions that better manages the Ingestions::Context that's passed in, using an `ensure` block to, well, ensure that teardown happens after success, failure, or exception.


scryptmouse (approved):
this looks great

PR #4146, opened 2026-08-27
fix: missing title + form errors regression

1aurend:


1aurend:
@dananjohnson this gets all the errors rendering again. I think there's probably more work to do at some point on errors. The `Errorable` component is used quite a few places without ids, so there's probably context missing from error announcements in those cases. I removed the `nameForError` prop completely, since it looks like it became obsolete some time ago.

dananjohnson (approved):
Looks good!

PR #4161, opened 2026-09-17
fix: fix resource sort order UX

dananjohnson:
The Sort filter in resources lists on the frontend suggests that the default order is A–Z, but it's not actually applying any value, it's just the first option in the select. The API is returning the resources in the default order, which is featured resources followed by newest.

This commit creates a new default option in the select labelled "Featured, then Newest" to remove confusion and make clear to users what the actual default order is.


dananjohnson:
@timfrazee this doesn't really need BE eyes, just a 👍🏻 to be able to merge it

timfrazee (approved):

